PR #4, opened 2022-09-01
C4 dataset

knighton:


streaming/text/c4.py
```diff
@@ -0,0 +1,81 @@
+from typing import Any, Dict, Optional
+
+from transformers.models.auto.tokenization_auto import AutoTokenizer
+
+from ..base import Dataset
```
karan6181, line 5: Not a blocker but we should decide on `relative` vs `absolute` import consistency. PEP8 recommends absolute imports and the same we have in `composer`.

streaming/text/convert/c4.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def parse_args() -> Namespace:
+    """Parse commandline arguments.
+
+    Args:
+        Namespace: Commandline arguments.
+    """
+    args = ArgumentParser()
+    args.add_argument('--out', type=str, default='/datasets/mds/c4-zstd/')
```
karan6181, line 20: Can we have the default datasets start from current directory rather than root `/`. For example, `./datasets/mds/c4-zstd/`. Sometimes, there could be a permission issue where the executor doesn't have a permission to create a directory in root.

streaming/text/convert/c4.py
```diff
@@ -0,0 +1,109 @@
...
+    args.add_argument('--compression', type=str, default='zstd:7')
```
karan6181, line 21: Same as above.

streaming/text/convert/c4.py
```diff
@@ -0,0 +1,109 @@
...
+        def num_shards(self):
+            return len(self.dataset._ex_iterable.kwargs['filepaths'])  # pyright: ignore
+
+        def __iter__(self):
+            worker_info = get_worker_info()
+            if worker_info:
+                num_workers = worker_info.num_workers
+                worker_id = worker_info.id
+                shards = self.dataset._ex_iterable.kwargs['filepaths']  # pyright: ignore
+                assert len(shards) % num_workers == 0
```
karan6181, line 55: Recommendation: Can we please raise exception rather than `assert` as exception error code and error message is more descriptive and meaningful. Thanks!

streaming/text/convert/c4.py
```diff
@@ -0,0 +1,109 @@
...
+    """Main: create streaming CIFAR10 dataset.
+    Args:
+        args (Namespace): Commandline arguments.
+    """
+    splits = [
+        ('train', 'train', 364868892, 64),
+        ('validation', 'val', 364608, 8),
+    ]
+    fields = {'text': 'str', 'timestamp': 'str', 'url': 'str'}
+    hashes = args.hashes.split(',') if args.hashes else []
```
karan6181, line 96:
```suggestion
    from streaming.vision.convert.base import get_list_arg
    hashes = get_list_arg(args.hashes)
```

streaming/base/dataset.py
```diff
@@ -97,7 +97,8 @@ def __init__(self,
                  timeout: float = 60,
                  hash: Optional[str] = None,
                  batch_size: Optional[int] = None) -> None:
-        keep_zip = (remote == local) if keep_zip is None else keep_zip
+        if keep_zip is None:
```
karan6181, line 100: I see that `keep_zip: Optional[bool] = None,`, should we have `False` as a default? E.g.: `keep_zip: Optional[bool] = False,` ?

knighton: Explainer:

Let's assume zipping provides net benefit (for example, text data), so zipping is used for a given dataset. Shards will therefore exist in two forms: the zipped form which they are stored as, and the unzipped form which they are used as. Remote directories store in only the zipped form, and local directories need to store only the unzipped form. So normally, you'd want keep_zip = False, to conserve storage as locals don't need the zipped form which they are downloaded as. But there's a special case: what if there is no remote/the local is the remote? In that case, you want to not blow away the zipped shards because they are how the dataset is persistently stored. The default value of keep_zip = None does that ("False, except that special case").

streaming/text/convert/c4.py
knighton: how about we move `get_list_arg` to a `streaming.base.util`? unwise to have `text` depend on `vision`

streaming/text/convert/c4.py
knighton: Done

streaming/text/convert/c4.py
knighton: Eh?

streaming/text/c4.py
knighton: absolute imports it is

streaming/text/convert/c4.py
karan6181: Sounds good to me. Also do you have a preference of using `--in` vs `--in_root` and `--out` vs `--out_root`. I prefer latter as it is more verbose.

streaming/text/convert/c4.py
karan6181: My bad :D . My comment was for `args.add_argument('--out', type=str, default='/datasets/mds/c4-zstd/')` to have a default path from current dir rather than root. Based on your [this](https://github.com/mosaicml/streaming/pull/4/#discussion_r967383343) comment, we are anyhow leaning towards making a required argument so ignore this comment.

streaming/text/convert/c4.py
karan6181: Sounds good to me.

streaming/base/dataset.py
karan6181: Make sense. Thanks for the explanation!

karan6181 (approved):
LGTM. Thanks! Approving the PR. You can make the changes that you suggested in the comments in this PR or next PR. I am OK with anything!

streaming/text/convert/c4.py
knighton: Sure, switching everyone to in_root/out_root

streaming/text/convert/c4.py
knighton: Done

PR #6, opened 2022-09-09
Added pre-commit config

karan6181:
## Description
- Added a pre-commit config to run on every commit
- Added bare minimum `yaph`, `isort`, and `pyright` support.
- Took reference from [mosaicml/composer](https://github.com/mosaicml/composer/blob/dev/.pre-commit-config.yaml)


knighton (approved):
I assume the makefile is still good

karan6181:
> I assume the makefile is still good

Yes it is still good. Maybe we don't need a section `style` and `run` in Makefile anymore since the pre-commit will gonna anyhow run it for us.

PR #11, opened 2022-09-14
Initial Docs site content

bandish-shah:
* Docs site seed content for:
   *  Welcome page
   *  Quick Start Guide

To Do:
- [x] [Karan] Quick Start Guide
- [x] [Karan] User Guide w/ pages for porting user dataset and creating streaming wrapper for user datasets
- [x] [Karan] Add a CIFAR10 end-to-end tutorial without composer (With Native PyTorch)
- [x] [Karan] Add a FaceSynthetics end-to-end tutorial with Composer
- [x] [James] API docstrings


docs/source/getting_started/quick_start.md
```diff
@@ -0,0 +1,64 @@
...
+        'clf': np.random.randint(10),
+    }
+    for _ in range(1000)
+]
+
+# Call `MDSWriter` to iterate through the input data and write into a shard `mds` file
+with MDSWriter(dirname, columns, compression, hashes) as out:
+    for sample in samples:
+        out.write(sample)
+```
```
bandish-shah, line 43: @karan6181 would it make sense to also import `boto3` in this example and just do the upload to an S3 bucket from `dirname` as a part of this script?

karan6181: I think it is a good idea. But few usecase we need to consider here
1. Boto3 s3 requires `aws_access_key_id`, `aws_secret_access_key`, and `region_name` to instantiate an s3 client.
2. Boto3 s3 doesn't have an API to upload folder. It has `upload_file()` method which upload only single file. So we need to provide a method which iterate overate a directory and upload all the files. All of this will be a big code.

On the flip side, we can refer the user to use [AWS CLI](https://aws.amazon.com/cli/) to upload the data files. It also support to upload a folder. For example,

```
$ aws s3 cp dirname s3://mybucket/myfolder --recursive
```

I have added [AWS CLI](https://aws.amazon.com/cli/)  step in new `user_guide.md` file.

Let me know your thoughts. We can do the same in `quick_start.md` guide.

karan6181: As per offline discussion, updated the `quick_start.md` guide with using AWS CLI.

knighton (approved):

PR #13, opened 2022-09-15
Add a COCO streaming dataset class

karan6181:
## Description
- Add a COCO streaming dataset class and modified the encoding and decoding type based on v1.

### DIFF
--- a/streaming/vision/coco.py
+++ b/streaming/vision/coco.py
@@ -39,7 +39,6 @@ class COCO(Dataset):
                  timeout: float = 60,
                  hash: Optional[str] = None,
                  batch_size: Optional[int] = None):
-
         super().__init__(local, remote, split, shuffle, prefetch, keep_zip, retry, timeout, hash,
                          batch_size)
 
@@ -53,8 +52,8 @@ class COCO(Dataset):
         wtot = x['wtot']
         bbox_sizes = x['bbox_sizes']
         bbox_labels = x['bbox_labels']
+        coco_args = (img, (htot, wtot), bbox_sizes, bbox_labels)
         if self.transform:
-            img, (htot,
-                  wtot), bbox_sizes, bbox_labels = self.transform(img, (htot, wtot), bbox_sizes,
-                                                                  bbox_labels)
+            coco_args = self.transform(*coco_args)
+            img, (htot, wtot), bbox_sizes, bbox_labels = coco_args
         return img, img_id, (htot, wtot), bbox_sizes, bbox_labels
--- a/streaming/vision/coco.py
+++ b/streaming/vision/coco.py
@@ -52,8 +52,8 @@ class COCO(Dataset):
         wtot = x['wtot']
         bbox_sizes = x['bbox_sizes']
         bbox_labels = x['bbox_labels']
-        coco_args = (img, (htot, wtot), bbox_sizes, bbox_labels)
         if self.transform:
-            coco_args = self.transform(*coco_args)
-            img, (htot, wtot), bbox_sizes, bbox_labels = coco_args
+            img, (htot,
+                  wtot), bbox_sizes, bbox_labels = self.transform(img, (htot, wtot), bbox_sizes,
+                                                                  bbox_labels)
         return img, img_id, (htot, wtot), bbox_sizes, bbox_labels
--- a/streaming/vision/coco.py
+++ b/streaming/vision/coco.py
@@ -1,0 +1,60 @@
+# Copyright 2022 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any, Callable, Optional
+
+from streaming.base import Dataset
+
+
+class COCO(Dataset):
+    """
+    Implementation of the COCO dataset using streaming Dataset.
+
+    Args:
+        local (str): Local filesystem directory where dataset is cached during operation.
+        remote (str, optional): Remote directory (S3 or local filesystem) where dataset is stored. Default: ``None``.
+        split (str, optional): The dataset split to use, either 'train' or 'val'. Default: ``None``.
+        shuffle (bool, optional): Whether to shuffle the train samples in this dataset. Default: ``True``.
+        transform (callable, optional): A function/transform that takes in an image and bboxes and returns a transformed
+            version. Default: ``None``.
+        prefetch (int, optional): Target number of samples remaining to prefetch while iterating. Default: ``100_000``.
+        keep_zip (bool, optional): Whether to keep or delete the compressed file when decompressing downloaded shards.
+            If set to None, keep iff remote == local. Default: ``None``.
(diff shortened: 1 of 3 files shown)


streaming/vision/convert/coco.py
```diff
@@ -181,15 +181,15 @@ def each(dataset: _COCODetection, shuffle: bool) -> Iterable[Dict[str, bytes]]:
         img_data = dataset.images[img_id]
         img_basename = img_data[0]
         img_filename = os.path.join(dataset.img_folder, img_basename)
-        img_bytes = open(img_filename, 'rb').read()
+        img = Image.open(img_filename)
```
knighton, line 184: let's keep this as bytes so as to not decode and then reencode images (passing bytes should work here)

streaming/vision/convert/coco.py
```diff
@@ -181,15 +181,15 @@ def each(dataset: _COCODetection, shuffle: bool) -> Iterable[Dict[str, bytes]]:
...
-            'htot': np.int64(htot).tobytes(),
-            'wtot': np.int64(wtot).tobytes(),
-            'bbox_sizes': bbox_sizes.numpy().tobytes(),  # (_, 4) float32.
-            'bbox_labels': bbox_labels.numpy().tobytes(),  # int64.
+            'img': img,
+            'img_id': img_id,
+            'htot': htot,
+            'wtot': wtot,
+            'bbox_sizes': bbox_sizes,  # (_, 4) float32.
+            'bbox_labels': bbox_labels,  # int64.
```
knighton, line 192: maybe fix comment to "# list of int64" or such

streaming/vision/coco.py
```diff
@@ -0,0 +1,60 @@
...
+                 split: Optional[str] = None,
+                 shuffle: bool = True,
+                 transform: Optional[Callable] = None,
+                 prefetch: Optional[int] = 100_000,
+                 keep_zip: Optional[bool] = None,
+                 retry: int = 2,
+                 timeout: float = 60,
+                 hash: Optional[str] = None,
+                 batch_size: Optional[int] = None):
+
```
knighton, line 42: spurious newline?

streaming/vision/coco.py
```diff
@@ -0,0 +1,60 @@
...
+    def __getitem__(self, idx: int) -> Any:
+        x = super().__getitem__(idx)
+        img = x['img'].convert('RGB')
+        img_id = x['img_id']
+        htot = x['htot']
+        wtot = x['wtot']
+        bbox_sizes = x['bbox_sizes']
+        bbox_labels = x['bbox_labels']
+        if self.transform:
+            img, (htot,
```
knighton, line 57: would it be more elegant to say "args = self.transform(*args)" or would lint not like star args

streaming/vision/convert/coco.py
```diff
@@ -200,12 +200,12 @@ def main(args: Namespace) -> None:
...
-        'img_id': 'bytes',
-        'htot': 'bytes',
-        'wtot': 'bytes',
-        'bbox_sizes': 'bytes',
-        'bbox_labels': 'bytes'
+        'img': 'jpeg',
+        'img_id': 'int',
+        'htot': 'int',
+        'wtot': 'int',
+        'bbox_sizes': 'pkl',
```
knighton, line 207: sounds like we may want to add "ndarray:(shape with 0 or 1 wildcards):dtype" in the future?

streaming/vision/convert/coco.py
karan6181: It has to decode anyways [here](https://github.com/mosaicml/streaming/blob/main/streaming/base/format/mds/encodings.py#L111) right if encoding format is `jpeg`!!  Then why passing it as `bytes` ? If we pass it as `bytes` then I see that the encoding is getting skipped [here](https://github.com/mosaicml/streaming/blob/main/streaming/base/format/mds/encodings.py#L193). Wondering that's hidden since I added encoding format as `jpeg`.  So I kept encoding as `jpeg` and I put image as of type `PIL.Image` so that proper encoding and decode method gets called.

knighton: during conversion, these images are already encoded as jpeg, so we take the bytes from the individual image files and directly write them into our mds shards

we avoid decoding to pil image and reencoding back to bytes in the conversion script because it's (a) slower and (b) jpeg is lossy, so the data will in all likelihood be changed by such a process

skipping reencoding jpeg is why i added that magic where it accepts raw bytes in the image encoding method

if i am understanding you correctly

karan6181: Yes Make Sense. Thanks for the explanation. I moved back to bytes in https://github.com/mosaicml/streaming/pull/13/commits/20fab46995190a3fbd24ad7c4f4e08249de05091.

streaming/vision/convert/coco.py
karan6181: It is correct. For example, below is the execution output from the same script

```
(Pdb) bbox_size
(0.0, 0.105375, 0.911296875, 0.9311875)
(Pdb) bbox_label
58
```

streaming/vision/coco.py
karan6181: :( pylint should have handle this. We need to improve pre-config commit. I removed it in https://github.com/mosaicml/streaming/pull/13/commits/20fab46995190a3fbd24ad7c4f4e08249de05091.

streaming/vision/coco.py
karan6181: Yes. Thanks! Changed it in https://github.com/mosaicml/streaming/pull/13/commits/20fab46995190a3fbd24ad7c4f4e08249de05091. I tested it and it works.

streaming/vision/convert/coco.py
karan6181: Can you please elaborate on this ?

knighton: just speculating on a potential future type of encoding

streaming/vision/coco.py
```diff
@@ -53,8 +52,8 @@ def __getitem__(self, idx: int) -> Any:
         wtot = x['wtot']
         bbox_sizes = x['bbox_sizes']
         bbox_labels = x['bbox_labels']
+        coco_args = (img, (htot, wtot), bbox_sizes, bbox_labels)
         if self.transform:
-            img, (htot,
-                  wtot), bbox_sizes, bbox_labels = self.transform(img, (htot, wtot), bbox_sizes,
-                                                                  bbox_labels)
+            coco_args = self.transform(*coco_args)
```
knighton, line 57: oof, i didn't notice we are passing img_id as well in the return statement, i think this way looks worse, let's revert it?

karan6181: Yes you are right.

karan6181: Done in https://github.com/mosaicml/streaming/pull/13/commits/63203edee1a388f9ddc7a8edc5c72f700bf297fa.

knighton (approved):

PR #18, opened 2022-09-21
Fix wait-for-unzip race

knighton:
Fix two identical race conditions in shard preload and download respectively.

The problem: some workers were thinking they were all clear to go ahead before that was actually the case, and encountering momentarily missing, empty, or truncated shards as a result in scenarios involving enough GPUs, large enough shards, and compression. Streaming V2 handles compression differently than V1 and adds hash (and ideally file size) validation, but the "wait" path was not updated to include the added steps in prepping shards for use.

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -206,6 +206,35 @@ class Dataset(IterableDataset):
         """
         self._load_shards([shard], partition)
 
+    def _decompress_shard_part(self, zip_info: FileInfo, zip_filename: str, raw_filename: str,
+                               compression: Optional[str], wait: bool):
+        """Validate and decompress shard data.
+
+        Args:
+            zip_info (FileInfo): Compressed file info.
+            zip_filename (str): Compressed filename.
+            raw_filename (str): Decompressed filename.
+            compression (str, optional): Compression algorithm.
+            wait (bool): Whether to wait for another worker to do the work.
+        """
+        # Doer or waiter?
+        if not wait:
+            # If doer: load compressed, validate, decompress, save decompressed, maybe remove
+            # compressed to save space.
+            data = open(zip_filename, 'rb').read()
+            if self.hash:
+                assert get_hash(self.hash, data) == zip_info.hashes[self.hash]
+            data = decompress(compression, data)  # pyright: ignore
+            tmp_filename = raw_filename + '.tmp'
+            with open(tmp_filename, 'wb') as out:
+                out.write(data)
+            os.rename(tmp_filename, raw_filename)
+            if not self.keep_zip:
+                os.remove(zip_filename)
+        else:
+            # If waiter: wait on decompressed to be written.
+            wait_for_download(raw_filename, self.timeout)
+
     def _preload_shard_part(self,
                             shard: int,
                             partition: Partition,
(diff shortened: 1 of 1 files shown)


streaming/base/dataset.py
```diff
@@ -205,37 +206,99 @@ def _load_shard(self, shard: int, partition: Partition) -> None:
...
+        """
+        # Doer or waiter?
+        if not wait:
+            # If doer: load compressed, validate, decompress, save decompressed, maybe remove
+            # compressed to save space.
+            data = open(zip_filename, 'rb').read()
+            if self.hash:
+                assert get_hash(self.hash, data) == zip_info.hashes[self.hash]
+            data = decompress(compression, data)  # pyright: ignore
+            tmp_filename = raw_filename + '.tmp'
```
karan6181, line 228: Wondering why to write first in `*.tmp` file and then rename it to a writer extension such as `mds` ? Why not writing directly to `*.mds` file?

streaming/base/dataset.py
```diff
@@ -305,6 +368,54 @@ def _download_file(self, basename: str, wait: bool = False) -> str:
...
+                self._download_file(zip_info.basename, wait)
+
+            # Validate and decompress (or wait on that).
+            self._decompress_shard_part(zip_info, zip_filename, raw_filename, compression, wait)
+        else:
+            # Download the raw version.
+            self._download_file(raw_info.basename, wait)
+
+            # Doer or waiter?
+            if not wait:
```
karan6181, line 413: Question: How does `wait` value getting set to `True` or `False` inside multiple dataloader workers ? is it through lock or something ?

streaming/base/dataset.py
```diff
@@ -305,6 +368,54 @@ def _download_file(self, basename: str, wait: bool = False) -> str:
...
+        raw_filename = os.path.join(self.local, self.split, raw_info.basename)
+        if os.path.isfile(raw_filename):
+            return
+
+        # Determine if we are the doer (process performing the download), or the waiter (one waiting
+        # for it to complete). Multiple processes share a shard when it is split between workers.
+        wait = shard not in partition.shards_to_download
+
+        # Is compression used?
+        if zip_info:
```
karan6181, line 400: I am assuming `zip` keyword name means a compressed information. For example, `zip_filename` means `compressed file name` which can be `gz`, `zstd`, etc right? Then can we not used `zip` keyword for `compression` since it is misleading that we are looking for `zip` extension. Maybe use ``compressed_info` and `compressed_filename` or maybe other names?

streaming/base/dataset.py
knighton: the source of truth here is get_partition() in index.py, which assigns balanced-sized slices of shards to workers (partitions of work)

streaming/base/dataset.py
knighton: because we are doing a file existence check to determine when it is okay to proceed in the "waiter" process, the file must be entirely written as soon as it exists, so we write it to a temp file on the same filesystem then rename in place to the final name

knighton: in a future PR, we should check file size as well as existence (but hash validation is too slow in that loop), as well as validating file size everywhere hashes are being compared, but let's treat that in its own PR

karan6181: Sounds Good. Make Sense.

streaming/base/dataset.py
knighton: - i originally tried that, but there are a lot of variable names that need to be prefixed accordingly to whichever naming scheme we use and it got completely unwieldly due to length/verbosity
- if we use `compressed_`, is the opposite file `uncompressed_` ('before compression' but also 'never to have been compressed') or `decompressed_` ('after undoing compression')
- the z in gz, zstd, etc. all stand for zip
- also tried variable name scheme `cmp_` and `dec_`
- after much gnashing, went with the convention `zip_` and `raw_` which seemed clearest

karan6181: Sure. sounds good to me.

karan6181 (approved):
LGTM. Thanks!

PR #20, opened 2022-09-22
Added pydocstyle and docformatter in pre-commit config

karan6181:
## Description
- Added pydocstyle and docformatter in pre-commit config
- Added docstring as part of it
- Update the existing docstrings and comments.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -371,9 +371,7 @@ def _auto_rst_for_module(module: types.ModuleType, exclude_members: List[Any]) -
 
 
 def _modules_to_rst() -> List[types.ModuleType]:
-    """Return the list of modules for which to generate API reference rst
-    files."""
-
+    """Return the list of modules for which to generate API reference rst files."""
     document_modules: List[types.Module] = [
         streaming,
     ]
@@ -389,8 +387,8 @@ def _modules_to_rst() -> List[types.ModuleType]:
 def _generate_rst_files_for_modules() -> None:
     """Generate .rst files for each module to include in the API reference.
 
-    These files contain the module docstring followed by tables listing
-    all the functions, classes, etc.
+    These files contain the module docstring followed by tables listing all the functions, classes,
+    etc.
     """
     docs_dir = os.path.abspath(os.path.dirname(__file__))
     module_rst_save_dir = os.path.join(docs_dir, 'api_reference')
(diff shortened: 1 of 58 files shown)


docs/source/conf.py
```diff
@@ -371,13 +371,11 @@ def _auto_rst_for_module(module: types.ModuleType, exclude_members: List[Any]) -
...
 def _modules_to_rst() -> List[types.ModuleType]:
-    """Return the list of modules for which to generate API reference rst
-    files."""
-
+    """Return the list of modules for which to generate API reference rst files."""
     document_modules: List[types.Module] = [
         streaming,
     ]
-    exclude_modules: List[types.Module] = []
+    exclude_modules: List[types.Module] = [streaming.base, streaming._version]
```
knighton, line 378: why? there is a lot of documented code under base which does not happen to be imported to top-level

knighton: also the docs are missing for streaming.text, streaming.vision, we need to fix this too (__all__?)

karan6181: Added text and vision to a doc in https://github.com/mosaicml/streaming/pull/20/commits/b9e4a2d410dea27e5b08949e7ddca132ee29444d. Looks like adding all the class and method as part of `base` require some thorough investigation. I will do that in a separate PR.

streaming/__init__.py
```diff
@@ -1,10 +1,15 @@
...
+import streaming.text as text
+import streaming.vision as vision
 from streaming._version import __version__
 from streaming.base import (CSVWriter, Dataset, JSONWriter, LocalDataset, MDSWriter, TSVWriter,
                             XSVWriter)
 
 __all__ = [
-    'Dataset', 'CSVWriter', 'JSONWriter', 'MDSWriter', 'TSVWriter', 'XSVWriter', 'LocalDataset'
+    'Dataset', 'CSVWriter', 'JSONWriter', 'MDSWriter', 'TSVWriter', 'XSVWriter', 'LocalDataset',
+    'vision', 'text'
```
knighton, line 14: i assume we want `base` too here

karan6181: I tried that too but the docs is not generating for sub-directory. I will take a look in a separate PR to make it to work.

knighton (approved):

PR #22, opened 2022-09-28
Improve algorithmic complexity of sample-to-shard lookup from O(log N) to O(1)

knighton:
Improve algorithmic complexity of sample-to-shard lookup from O(log N) to O(1) by means of a lookup table followed by a comparison whose entries, corresponding to sample ID ranges, overlap at most two shards. Previously this was done by bisection.


karan6181 (approved):
LGTM. Thanks!

PR #23, opened 2022-09-28
Add enwiki-20200101 streaming dataset

knighton:
Command to perform conversion to MDS:

streaming$ python3 -m streaming.text.convert.enwiki --in_root /workdisk/vlad/mlperf-bert-nvidia/workspace/bert_data/download/results4/ --out_root /workdisk/james/enwiki/

### DIFF
--- a/streaming/text/convert/enwiki.py
+++ b/streaming/text/convert/enwiki.py
@@ -1,0 +1,114 @@
+from argparse import ArgumentParser, Namespace
+import os
+from typing import Dict, List
+
+from tqdm import tqdm
+
+from streaming.base import MDSWriter
+from streaming.base.util import get_list_arg
+
+
+def parse_args() -> Namespace:
+    """Parse command-line arguments.
+
+    Args:
+        Namespace: command-line arguments.
+    """
+    args = ArgumentParser()
+    args.add_argument(
+        '--in_root',
+        type=str,
+        required=True,
+        help='Directory path to store the input dataset',
+    )
+    args.add_argument(
+        '--out_root',
+        type=str,
+        required=True,
+        help='Directory path to store the output dataset',
+    )
+    args.add_argument(
+        '--compression',
+        type=str,
+        default='zstd:7',
+        help='Compression algorithm to use. Default: zstd:7',
+    )
+    args.add_argument(
+        '--hashes',
+        type=str,
+        default='sha1,xxh64',
+        help='Hashing algorithms to apply to shard files. Default: sha1,xxh64',
+    )
+    args.add_argument(
+        '--size_limit',
+        type=int,
+        default=1 << 25,
+        help='Shard size limit, after which point to start a new shard. Default: 1 << 27',
+    )
+    args.add_argument(
+        '--progbar',
+        type=int,
+        default=1,
+        help='tqdm progress bar. Default: 1 (True)',
+    )
+    args.add_argument(
+        '--leave',
+        type=int,
+        default=0,
+        help='Keeps all traces of the progressbar upon termination of iteration. Default: 0 ' +
+        '(False)',
+    )
+    return args.parse_args()
+
+
+def process_split(in_root: str, out_root: str, compression: str, hashes: List[str], size_limit: int,
+                  progbar: int, leave: int, basenames: List[str], split: str) -> None:
+    """Process a dataset split.
+
+    Args:
+        in_root (str): Input dataset path.
+        out_root (str): Output dataset path.
+        compression (str): Which compression to use, or empty string if none.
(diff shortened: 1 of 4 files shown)


dskhudia:
Can you add the command you used to call this script in the description so that it's available for future?

streaming/text/convert/enwiki.py
```diff
@@ -0,0 +1,114 @@
...
+        default=0,
+        help='Keeps all traces of the progressbar upon termination of iteration. Default: 0 ' +
+        '(False)',
+    )
+    return args.parse_args()
+
+
+def process_split(in_root: str, out_root: str, compression: str, hashes: List[str], size_limit: int,
+                  progbar: int, leave: int, basenames: List[str], split: str) -> None:
+    """Process a dataset split.
```
dskhudia, line 66: What does this mean? Process a dataset that has already been split to different shards?

dskhudia (approved):

karan6181:
Looks good overall. Can you please address below two points? Thanks!
1. Add a description to the PR on what the PR is about
2. `pre-commit` is failing with unused import and Missing docstring.

streaming/text/convert/enwiki.py
karan6181: No. The `process_split()` reads the raw dataset (for example, which we downloaded from internet) and converts into `*.mds` file extension with multiple shard files.

karan6181: @knighton maybe it needs a better descriptive method comments ?

dskhudia: In this case basenames is already a collection of 500 txt files.

dskhudia: Got it. So it's just reading the files one-by-one but sharding etc. is handled by the writer.

streaming/text/convert/enwiki.py
```diff
@@ -0,0 +1,119 @@
...
+        '--hashes',
+        type=str,
+        default='sha1,xxh64',
+        help='Hashing algorithms to apply to shard files. Default: sha1,xxh64',
+    )
+    args.add_argument(
+        '--size_limit',
+        type=int,
+        default=1 << 25,
+        help='Shard size limit, after which point to start a new shard. Default: 1 << 27',
```
karan6181, line 51: nit: `1 << 25` in `help` message

streaming/text/convert/enwiki.py
```diff
@@ -0,0 +1,119 @@
...
+
+def main(args: Namespace) -> None:
+    """Main: create streaming enwiki dataset.
+
+    Args:
+        args (Namespace): command-line arguments.
+    """
+    hashes = get_list_arg(args.hashes)
+
+    basenames = [f'part-{i:05}-of-00500' for i in range(500)]
```
karan6181, line 108: Is the input filename always has this predefined name? can there be a case where the filename is different for any user? Wondering if we need something like `os.walk()` to get the filename during runtime?

streaming/text/enwiki.py
```diff
@@ -0,0 +1,101 @@
...
+            raise ValueError(f'Only group_method="truncate" is supported at this time.')
+
+        super().__init__(local, remote, split, shuffle, prefetch, keep_zip, retry, timeout, hash,
+                         batch_size)
+        self.tokenizer_name = tokenizer_name
+        self.max_seq_len = max_seq_len
+        self.group_method = group_method
+
+        # Build tokenizer
+        self.tokenizer = AutoTokenizer.from_pretrained(self.tokenizer_name)  # pyright: ignore
```
karan6181, line 65: is this `# pyright: ignore` still relevant here of copy/paste ?

streaming/text/convert/enwiki.py
knighton: Done

streaming/text/convert/enwiki.py
knighton: Yes, always named as such, as the output of a previous preprocessing stage

streaming/text/enwiki.py
knighton: Will try removing it

karan6181: Fixed in https://github.com/mosaicml/streaming/pull/23/commits/926d9779baed22f1e9392a6b96b43cfc3f63d68e

streaming/text/convert/enwiki.py
karan6181: Fixed in https://github.com/mosaicml/streaming/pull/23/commits/926d9779baed22f1e9392a6b96b43cfc3f63d68e.

karan6181 (approved):
LGTM. Thanks!

PR #25, opened 2022-10-03
Add unittest for compression

karan6181:
## Description
- Add unittest for compression algorithms.

### DIFF
--- a/streaming/base/compression/compression.py
+++ b/streaming/base/compression/compression.py
@@ -6,7 +6,7 @@
 import bz2
 import gzip
 from abc import ABC, abstractmethod
-from typing import Dict, Iterator, Optional, Set, Tuple, Type
+from typing import Dict, Iterator, Optional, Set, Tuple, Type, Union
 
 import brotli
 import snappy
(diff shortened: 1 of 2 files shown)


streaming/base/compression/compression.py
```diff
@@ -132,7 +132,7 @@ class Snappy(Compression):
 
     extension = 'snappy'
 
-    def compress(self, data: bytes) -> bytes:
+    def compress(self, data: Union[bytes, str]) -> bytes:
```
karan6181, line 135: Looks like snappy compression algorithms takes string as well apart from bytes. Hence, made this changes.

knighton: we are not using this with strings for our purposes, let's not enable that use case here?

karan6181: Looks like python annotations is very strict. If we keep it as `bytes` and you pass a string, then it won't throw any error and at a same time, it compressed the data and return the data in `bytes` !! 😞

knighton: would rather keep it uniform with the other compression algo interfaces and call this silent extra capability an easter egg? up to you if you feel strongly about it

knighton (approved):

streaming/base/compression/compression.py
karan6181: Cool. Let me change it back to `bytes` as this is not a major concern.

karan6181: Done in https://github.com/mosaicml/streaming/pull/25/commits/0e5d87e61ee2b027df73751826c615755ffdf499.

PR #26, opened 2022-10-04
Fix hang when compression is used but compressed files are not retained

knighton:
Fix a bug where we were waiting on an ephemeral downloaded file that another process removes as soon as it decompresses. We should wait on the decompressed form instead, which is retained.

Tracking Issue: [CO-1175](https://mosaicml.atlassian.net/browse/CO-1175)


karan6181 (approved):
LGTM. Thanks!

PR #28, opened 2022-10-06
Fix preprocessing for English Wikipedia dataset

knighton:
This PR adds two directories: (a) the official tfrecord preprocessing, and (b) this code ported to use MDS, kept as otherwise identical as possible.

    streaming/text/convert/enwiki/mds/
    streaming/text/convert/enwiki/tfrecord/

Procedure to generate MDS shards:

### Training split:

- `python3 make_train_parallel.py`  # Create dataset split as 128 subsets
  - calls `create_pretraining_data.py`
- `python3 merge_shard_groups.py`  # Merge subsets into one dataset split

### Evaluation split:

- `./make_eval.sh`
  - calls `create_pretraining_data.py`  #  Create dataset splilt from eval file
  - calls `pick_eval_samples.py`  # Pick the 10k samples


dskhudia:
How many samples in the eval for this?

dskhudia (approved):

PR #29, opened 2022-10-06
Add long_description for packaging

bandish-shah:
* Updated `description` in setup.py
* Added `long_description` to include README during packaging


karan6181 (approved):
LGTM! Thanks for fixing it!

PR #31, opened 2022-10-06
Fix enwiki dataset

dskhudia:
bert models expect data to contain `input_ids`, `token_type_ids`, `attention_mask` and `labels`. Fix loaded data to use these names.

### DIFF
--- a/streaming/text/enwiki.py
+++ b/streaming/text/enwiki.py
@@ -51,11 +51,14 @@ class EnWiki(Dataset):
         self.field_dtypes = {
             'input_ids': np.int32,
             'input_mask': np.int32,
+            'attention_mask': np.int32,
             'segment_ids': np.int32,
+            'token_type_ids': np.int32,
             'masked_lm_positions': np.int32,
             'masked_lm_ids': np.int32,
             'masked_lm_weights': np.float32,
             'next_sentence_labels': np.int32,
+            'labels': np.int32,
         }
 
     def __getitem__(self, idx: int) -> Any:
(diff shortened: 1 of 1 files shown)


streaming/text/enwiki.py
```diff
@@ -68,7 +71,18 @@ def __getitem__(self, idx: int) -> Any:
...
+        if 'input_mask' in obj.keys():
+            obj['attention_mask'] = obj.pop('input_mask')
+        obj.pop('next_sentence_labels')
+        obj.pop('masked_lm_weights')
         for key, value in obj.items():
             dtype = self.field_dtypes[key]
             obj[key] = np.frombuffer(value, dtype)
+        input_len = len(obj['input_ids'])
+        obj['labels'] = np.full((input_len,), -100)
+        obj['labels'][obj['masked_lm_positions']] =  obj['masked_lm_ids']
```
knighton, line 85: double space

streaming/text/enwiki.py
```diff
@@ -68,7 +71,18 @@ def __getitem__(self, idx: int) -> Any:
...
+        obj.pop('masked_lm_positions')
+        obj.pop('masked_lm_ids')
```
knighton, line 87: Would it be simpler to return a new dict over changing in-place?

knighton (approved):

streaming/text/enwiki.py
dskhudia: What do you prefer? I can do either.

dskhudia: Created a new dict.

PR #32, opened 2022-10-06
Skip pre-commit check for enwiki convert skip to have code parity

karan6181:
## Description
- PR https://github.com/mosaicml/streaming/pull/28 added a Enwiki dataset conversion scripts for streaming.
- As part of the code parity with the official tfrecord preprocessing, pre-commit fails with multiple linting problem. Hence, this PR exclude the enwiki convert directory pre-commit checks.
- As part of a long term vision, we should move all those scripts to a common sub-folders.

### DIFF
--- a/streaming/text/enwiki.py
+++ b/streaming/text/enwiki.py
@@ -3,10 +3,9 @@
 
 """English Wikipedia 2020-01-01 streaming dataset."""
 
-import numpy as np
-from typing import Any, Dict, Optional
+from typing import Any, Optional
 
-from transformers.models.auto.tokenization_auto import AutoTokenizer
+import numpy as np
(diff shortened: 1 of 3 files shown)


streaming/text/enwiki.py
```diff
@@ -3,10 +3,9 @@
 
 """English Wikipedia 2020-01-01 streaming dataset."""
 
-import numpy as np
-from typing import Any, Dict, Optional
+from typing import Any, Optional
 
-from transformers.models.auto.tokenization_auto import AutoTokenizer
```
karan6181, line 9: unused import. Hence removed.

knighton (approved):

PR #34, opened 2022-10-07
Parallel tfrecord creation, validate sample counts vs MDS

knighton:


streaming/text/convert/enwiki/tfrecord/make_train_parallel.py
```diff
@@ -0,0 +1,25 @@
+import os
+
+num_cpus = os.cpu_count()
+num_shards = 500
+input_file_pattern = '/tmp/enwiki_preproc/results4/part-%05d-of-00500'
```
karan6181, line 5: Does this path exist ? Should we get it from command line as argparse ?

knighton: Done

karan6181 (approved):

PR #37, opened 2022-10-11
Elastic world size deterministic shuffle with mid-epoch resumption

knighton:
Design-wise, the streaming dataset is standalone and must not depend on Composer, ie for tracking where we are in the epoch samples and which epoch it is. This is in conflict with the distributed, sharded, forked, and/or ephemeral nature of dataloaders, datasets, and/or workers. We solve crossing the rank<->worker and worker<->epoch chasms by putting the state in shared memory

The good: we solve determinism/MER (including while changing world size/num workers) in shuffle.py by pregenerating the shuffle in such and such fancy ordering that won't knock over S3 and then replaying (split-truncate-join) all earlier-this-epoch training sessions to get each partition of samples comprising the remaining epoch. The shuffle part takes a handful of seconds for each session to process C4's shard sample counts on 1cpu laptop python. Code for dataset `__iter__`:

```
        sample_slot = self.cursor.new_session()
        sequences = get_epoch(self.shard_sizes, self.shuffle, self.seed, self.cursor.get_epoch(),
                              self.cursor.each_session())
        todos = sequences[dist.get_worker()]

        for index in todos:
            self.cursor.step_sample(sample_slot)
            yield self[index]

        self.cursor.clear_sessions()
        self.cursor.step_epoch()
```

The bad: this PR makes the problem easier in two big ways in order to solve it:
* No streaming (shards are assumed to exist uncompressed locally) -- we'd just rip the shard prefetch loop from streaming v2 but the problem needs to be nailed down on the local filesystem only case first
* Evenly divisible (dataset split size is evenly divisible by batch_size which is evenly divisible by devices x workers) -- since we track the samples seen of each worker we should be able to exactly recreate the samples seen of imbalanced earlier-this-epoch training sessions, and we would adapt get_partition() from v1/v2

The ugly: synchronization, looking for something like pthread_mutexattr_setpshared (or implement locks using filesystem)


knighton:
Test this PR with https://github.com/knighton/composer/tree/james/mer

```
diff --git a/composer/core/state.py b/composer/core/state.py
index 73d6cf15..989ff802 100644
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -243,6 +243,8 @@ class State(Serializable):
             +-----------------------+-------------------------------------------------------------+
             | run_name              | The run name for training.                                  |
             +-----------------------+-------------------------------------------------------------+
+            | dataset_state         | The dataset iteration state for checkpointing.              |
+            +-----------------------+-------------------------------------------------------------+
 
         timestamp (Timestamp): The current training timestamp.
         train_dataloader (Iterable): The training dataloader. (May be ``None`` if not training.)
@@ -276,6 +278,7 @@ class State(Serializable):
         dataloader: Optional[Iterable] = None,
         dataloader_label: Optional[str] = None,
         dataloader_len: Union[int, Time[int]] = -1,
+        dataset_state: Optional[Dict[str, Any]] = None,
 
         # precision
         precision: Union[str, Precision] = Precision.FP32,
@@ -303,6 +306,7 @@ class State(Serializable):
         self._dataloader = None
         self._dataloader_label = None
         self.set_dataloader(dataloader, dataloader_label, dataloader_len)
+        self.dataset_state = dataset_state
         self._max_duration = None
         self.max_duration = max_duration
 
@@ -351,6 +355,7 @@ class State(Serializable):
             'train_metrics',
             'eval_metrics',
             'run_name',
+            'dataset_state',
         ]
 
         self.train_metrics: Dict[str, Metric] = {}
@@ -501,7 +506,10 @@ class State(Serializable):
 
         for attribute_name in self.serialized_attributes:
             attribute_value = getattr(self, attribute_name)
-            if attribute_name == 'model':
+            if attribute_name == 'dataset_state':
+                if self.dataloader and hasattr(self.dataloader.dataset, 'state_dict'):
+                    serialized_value = self.dataloader.dataset.state_dict()
+            elif attribute_name == 'model':
                 # Save model directly instead of by class name, since model may be wrapped by DistributedDataParallel
                 # If it is DDP wrapped, do not save the `module.` prefix, as that is an implmentation detail
                 with get_fsdp_rank0_cpu_save_context(
@@ -619,7 +627,10 @@ class State(Serializable):
                 # It's possible some attributes we removed
                 continue
 
-            if attribute_name == 'model':
+            if attribute_name == 'dataset_state':
+                if self.dataloader and hasattr(self.dataloader.dataset, 'load_state_dict'):
+                    self.dataloader.dataset.load_state_dict(serialized_value)
+            elif attribute_name == 'model':
                 self.load_model_state(state, strict=strict)
             elif attribute_name == 'optimizers':
                 self.load_optim_state(state)
```

streaming/base/shared.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self, count: int, filelock_path: str, shm_path: str) -> None:
+        self.count = count
+        self.filelock_path = filelock_path
+        self.shm_path = shm_path
+
+        dirname = os.path.dirname(filelock_path)
+        os.makedirs(dirname, exist_ok=True)
+        self.lock = FileLock(filelock_path)
+
+        size = 3 * np.int32().nbytes
```
karan6181, line 38: Can you add a comment on why there is a number `3` ?

streaming/base/shared.py
```diff
@@ -0,0 +1,135 @@
...
+        try:
+            self._shm = SharedMemory(shm_path, True, size)
+        except:
+            self._shm = SharedMemory(shm_path)
+
+        self._arr = np.ndarray(3, buffer=self._shm.buf, dtype=np.int32)
```
karan6181, line 44: Can you add a comment on why there is a number `3` ?

knighton: Done

streaming/base/shared.py
knighton: Done

streaming/base/shared.py
```diff
@@ -0,0 +1,135 @@
...
+        self.num_enter = 0
+        self.num_exit = count
+        self.flag = True
+
+    def __del__(self):
+        """Destructor clears array that references shm."""
+        del self._arr
```
karan6181, line 52: why not calling `self._shm.close() ` and `self._shm.unlink()` in the body of destructor?

knighton: because this barrier's shared memory is almost certainly attached to other processes here

numpy arrays whose buffers are backed by shm don't reference count properly

karan6181: Hmmm I see. Then this could lead to a memory leak right ?

knighton: no, shms are garbage collected when the last holder goes out of scope, also i manually delete the numpy array there

streaming/base/world.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class World:
+    """Information about the nodes, ranks and workers of this run.
+
+    Warning: be careful as to whether this object was initialized in a worker (if workers are used)
+    or in a rank (which will claim one worker per rank).
+
+    Nodes:
+      - node / num_nodes
```
karan6181, line 18: just for our own understanding, we can add a comment that `each node should have a same number of ranks (GPUs)`, meaning, we support homogeneous clusters.

streaming/base/world.py
```diff
@@ -0,0 +1,57 @@
...
+
+        info = get_worker_info()
+        if info:
+            self.worker_of_rank = info.id
+            self.workers_per_rank = info.num_workers
+        else:
+            self.worker_of_rank = 0
+            self.workers_per_rank = 1
+
+        self.worker = self.rank * self.workers_per_rank + self.worker_of_rank
```
karan6181, line 51: Should we rename this to `global_worker` for easy understanding?

streaming/base/world.py
```diff
@@ -0,0 +1,57 @@
...
+        self.num_workers = self.num_ranks * self.workers_per_rank
```
karan6181, line 52: Should we rename this to `global_num_workers` for easy understanding?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+        will become invalidated when it goes out of scope.
+
+        Args:
+            epoch (int): The current epoch.
+
+        Returns:
+            Tuple[NDArray[np.int64], SharedMemory]: Session and handle to shared memory.
+        """
+        world = World()
+        shm_name = f'{self._prefix}_session_{epoch}'
```
karan6181, line 272: We can move all the naming line to a `getter()` method ? This way it would be easy to understand and modify at later point since it would be at just one place, i.e. in a `getter()` method

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+        Note: this also returns the underlying shared memory object (if it exists), because the
+        returned array will become invalidated when it goes out of scope.
+
+        Args:
+            epoch (int): The current epoch.
+
+        Returns:
+            Tuple[Optional[NDArray[np.int64]], Optional[SharedMemory]]: Maybe session, maybe shm.
+        """
+        shm_name = f'{self._prefix}_session_{epoch}'
```
karan6181, line 297: We can move all the naming line to a `getter()` method ? This way it would be easy to understand and modify at later point since it would be at just one place, i.e. in a `getter()` method

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+    def _get_partition(self, epoch: int, sessions: List[NDArray[np.int64]],
+                       world: World) -> Optional[NDArray[np.int64]]:
+        # Local leader generates the partitions.
+        if world.is_local_leader:
+            sequences = get_epoch(self.shard_sizes, self.shuffle, self.seed, epoch, sessions)
+            base = world.node * world.ranks_per_node * world.workers_per_rank
+            for rank_of_node in range(world.ranks_per_node):
+                for worker_of_rank in range(world.workers_per_rank):
+                    worker = base + rank_of_node * world.workers_per_rank + worker_of_rank
+                    name = f'{self._prefix}_part_{worker:03}'
```
karan6181, line 315: We can move all the naming line to a `getter()` method ? This way it would be easy to understand and modify at later point since it would be at just one place, i.e. in a `getter()` method

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+                    shm = SharedMemory(name, True, size)
+                    shm.buf[:] = sequence.tobytes()
+
+        self._barrier()
+
+        # Load our partition.
+        name = f'{self._prefix}_part_{world.worker:03}'
+        try:
+            shm = SharedMemory(name)
+        except:
```
karan6181, line 329: too generic exception catch ???

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+        num_workers = shm.size // np.int64().nbytes
+        cur_session = np.ndarray(num_workers, buffer=shm.buf, dtype=np.int64)
+        return cur_session, shm
+
+    def _get_partition(self, epoch: int, sessions: List[NDArray[np.int64]],
+                       world: World) -> Optional[NDArray[np.int64]]:
+        # Local leader generates the partitions.
+        if world.is_local_leader:
+            sequences = get_epoch(self.shard_sizes, self.shuffle, self.seed, epoch, sessions)
+            base = world.node * world.ranks_per_node * world.workers_per_rank
```
karan6181, line 311: looks like a unique number generator per worker process. Can you please add a comment about `base` here ?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+                    size = len(sequence) * np.int64(0).nbytes
+                    if not size:
+                        continue
+                    shm = SharedMemory(name, True, size)
+                    shm.buf[:] = sequence.tobytes()
+
+        self._barrier()
+
+        # Load our partition.
+        name = f'{self._prefix}_part_{world.worker:03}'
```
karan6181, line 326: Same as above of moving to `getter()`  method.

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+        try:
+            shm = SharedMemory(name)
+        except:
+            return None
+        todos = np.frombuffer(shm.buf, np.int64).copy()
```
karan6181, line 331: better name than `todos` ;-) . maybe `sample_ids` ??

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+        shm.unlink()
```
karan6181, line 332: I think we also need `shm.close()` before `shm.unlink()`. For example:

```
shm.close()
shm.unlink()
```

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+        presumed_epoch = self.next_epoch
+        epoch, old_sessions = self._resume(presumed_epoch)
+        cur_session, _ = self._create_cur_session(epoch)
+
+        world = World()
+        assert world.num_workers == self._barrier.count
+        self._barrier()
+
+        # Update the pre-incremented epoch counter.
+        world = World()
```
karan6181, line 533: Do we need to instantiate the `World()` again here? Since like it is already instantiated 3 lines above!!

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+
+        # Placeholder for _resume_shm, a shared memory object where load_state_dict() saves its
+        # data to be picked up by __iter__().
+        self._resume_shm = None
+
+        # Create the barrier.
+        total_workers = world.ranks_per_node * (self.num_workers or 1)
+        self._barrier_filelock_path = os.path.join('/tmp', 'mds', self._prefix, 'barrier_filelock')
+        self._barrier_shm_path = f'{self._prefix}_barrier_shm'
+        self._barrier = SharedBarrier(total_workers, self._barrier_filelock_path,
```
karan6181, line 178: Can we rename `self._barrier` to `self._local_barrier` or `self._per_node_barrier` since it is confusing with `torch.dist.barrier`

streaming/base/shared.py
```diff
@@ -0,0 +1,139 @@
...
+    def flag(self, flag: bool) -> None:
+        """Set property flag.
+
+        Args:
+            flag (bool): The flag value.
+        """
+        self._arr[2] = bool(flag)
+
+    def __call__(self) -> None:
+        """A set number of processes enter, wait, and exit the barrier."""
```
karan6181, line 112: Can you please add bunch of logging message with `debug` level saying `lock acquire by process x`, `lock release by process x`. etc since it would very helpful to debug the issue later on.

```
log.debug('sentence')
```

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+        world = World()
+        if world.is_local_leader:
+            self.next_epoch = epoch + 1
+
+        self._barrier()
+
+        # Get the samples for this worker to process.
+        sessions = old_sessions + [cur_session]
+        sample_ids = self._get_partition(epoch, sessions, world)
+        if sample_ids is None:
```
karan6181, line 542: In what scenario will the `sample_ids` be `None`? After the last sample in last batch ??? Can you please add the comment in that line? Thanks!

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+            epoch (int): Which epoch. On noticing that a new epoch has started, we exit this thread
+                because a new one will soon be running, with different sample IDs.
+            sample_ids (NDArray[np.int64]): The samples to download the shards of.
+        """
+        # Create or attach shard_states array.
+        name = f'{self._prefix}_shard_states'
+        size = len(self.shard_sizes) * np.uint8(0).nbytes
+        try:
+            shm = SharedMemory(name, True, size)
+        except:
```
karan6181, line 484: too generic exception catch?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+            shm = SharedMemory(name)
+        shard_states = np.ndarray(len(self.shard_sizes), buffer=shm.buf, dtype=np.uint8)
+
+        filename = os.path.join('/tmp', 'mds', self._prefix, '_shard_states_filelock')
```
karan6181, line 488: Few things:
1. use `os.sep` instead of `/`.
2. why `mds` ? since we support multiple writers of different type. Can we get `mds` name during runtime ?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+            # If we're out of samples this epoch, exit this thread because we are done downloading.
+            if self._download_index == num_samples:
+                break
+
+            # If we are requested to only pre-download so many samples, if we have as many or more
+            # downloaded already, we wait and check again later.
+            if self.predownload is not None:
+                samples_ahead = self._download_index - self._iter_index
+                if self.predownload <= samples_ahead:
+                    sleep(0.07)
```
karan6181, line 508: are you a big fan of James Bond movie ? ;) wondering where does `0.07` value comes from ?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,639 @@
...
+            lock.release()
+            while shard_states[shard_id] != _ShardState.DOWNLOADED:
+                sleep(0.07)
+        elif state == _ShardState.DOWNLOADED:
+            lock.release()
+        else:
+            raise RuntimeError('Unknown shard state')
+
+    def _download_thread(self, epoch: int, sample_ids: NDArray[np.int64]) -> None:
+        """Download the relevant shards in the background while we are being iterated.
```
karan6181, line 471: Wondering how many threads will there be running at once? For example, let say, the model training is happening on 1 GPU with `num_workers` as `8` in `torch.utils.data.DataLoader`. Does each worker process has 1 thread running to download the shards? or 1 node has only 1 thread running to download the shards?

Can you please add the comment saying how many thread would be running at a same time in avg.case?

karan6181 (commented):
General comment: please avoid using `assert` statement in the src code. It is ok to use `assert` statement in the test code. raise Exception instead of using `assert`. Thanks!

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+        # Get the samples for this worker to process.
+        sessions = old_sessions + [cur_session]
+        sample_ids = self._get_partition(epoch, sessions, world)
+        if sample_ids is None:
+            return
+
+        # Iterate while downloading.
+        self._iter_index = 0
+        self._download_index = 0
+        Thread(target=self._download_thread, args=(epoch, sample_ids)).run()
```
karan6181, line 547: I am not very positive about this line w.r.t releasing the resources and killing the thread. I am thinking, can we modified it as below? 

1. assign a thread local variable and put `daemon=True` such as
```
self.thread = Thread(target=self._download_thread, args=(epoch, sample_ids))
self.thread.start()
```
2. And put everything under `try..finally` block with waiting and release the thread. For example

```
try:
  .......
  self.thread = Thread(target=self._download_thread, args=(epoch, sample_ids))
  self.thread.start()
  ......
finally:
  if self.thread is not None:
    self.thread.join()
    self.thread = None
```

streaming/base/dataset/streaming.py
karan6181: Please assign all the constant at the top of the file with ALL_CAPS.

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+        with open(tmp_filename, 'wb') as out:
+            out.write(data)
+        os.rename(tmp_filename, raw_filename)
+
+        # Maybe remove compressed to save space.
+        if not self.keep_zip:
+            os.remove(zip_filename)
+
+    def _download_shard_part(self,
+                             shard: int,
```
karan6181, line 386: looks like `shard` variable is not getting used in this function !!!!

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+        for info in obj['shards']:
+            shard = reader_from_json(local, split, info)
+            self.shards.append(shard)
+
+        # Build the Index (for partitioning and mapping samples to shards).
+        self.shard_sizes = np.array([x.samples for x in self.shards])
+        self.index = Index(self.shard_sizes)
+
+        # Setup for coordinating.
+        device = torch.device(f'cuda:{world.rank_of_node}')
```
karan6181, line 138: what if someone wants to run on CPU ? They won't be having a `cuda` device right ?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+
+        # Add a coordinated random prefix to all shm names for uniqueness.
+        if world.is_leader:
+            tensor[0] = np.random.randint(1 << 60)
+        dist.broadcast(tensor, 0)
+        self._prefix = f'{int(tensor):016x}_{self.split}'
+
+        # Set up the epoch counter.
+        #
+        # Note: we do not assume that the end of __iter__() will ever be reached, so we need to
```
karan6181, line 157: Can you please explain this sentence `Note: we do not assume that the end of __iter__() will ever be reached, so we need to` more ? I didn't quite get it on why it will never reach end of __iter__()

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+        self.shards = []
+        for info in obj['shards']:
+            shard = reader_from_json(local, split, info)
+            self.shards.append(shard)
+
+        # Build the Index (for partitioning and mapping samples to shards).
+        self.shard_sizes = np.array([x.samples for x in self.shards])
+        self.index = Index(self.shard_sizes)
+
+        # Setup for coordinating.
```
karan6181, line 137: Can we move line 137 to 153 in a separate method? `__init__()` method is too crowded.

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+        self._next_epoch_arr = np.ndarray(1, buffer=self._next_epoch_shm.buf, dtype=np.int64)
+        self._next_epoch_arr[0] = 0
+
+        # Placeholder for _resume_shm, a shared memory object where load_state_dict() saves its
+        # data to be picked up by __iter__().
+        self._resume_shm = None
+
+        # Create the barrier.
+        total_workers = world.ranks_per_node * (self.num_workers or 1)
+        self._barrier_filelock_path = os.path.join('/tmp', 'mds', self._prefix, 'barrier_filelock')
```
karan6181, line 175: Few things:
1. use `os.sep` instead of `/`.
2. why `mds` ? since we support multiple writers of different type. Can we get `mds` name during runtime ?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+            return epoch, []
+
+        # Parse the existent resume state.
+        obj = json.loads(bytes(shm.buf).decode('utf-8'))
+
+        # Check if the resume state is stale.
+        if obj['epoch'] < epoch:
+            # Clean up stale state.
+            if world.is_local_leader:
+                shm.unlink()
```
karan6181, line 248: `shm.close()` before `shm.unlink()` ??

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+        Returns:
+            Tuple[int, List[NDArray[np.int64]]]: Pair of (resumed epoch, old sessions).
+        """
+        world = World()
+
+        # Get the resume state, if it exists.
+        name = f'{self._prefix}_resume'
+        try:
+            shm = SharedMemory(name)
+        except:
```
karan6181, line 237: `except FileExistsError` ?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+        Args:
+            epoch (int): The current epoch.
+
+        Returns:
+            Tuple[Optional[NDArray[np.int64]], Optional[SharedMemory]]: Maybe session, maybe shm.
+        """
+        shm_name = f'{self._prefix}_session_{epoch}'
+        try:
+            shm = SharedMemory(shm_name)
+        except:
```
karan6181, line 299: `except FileExistsError` ?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,638 @@
...
+                cur_session[world.worker] += 1
+                sample_id = sample_ids[self._iter_index]
+                yield self[sample_id]
+                self._iter_index += 1
+                continue
+            sleep(0.07)
+
+        # Any code after the yields will never be reached by the Composer trainer.
+
+    def _all_gather_current_session(self, session: NDArray[np.int64]) -> None:
```
karan6181, line 560: move this method to distributed.py file ?

karan6181 (commented):
Points related to code structure:

1. Can we create a class for `State` and add setter method (`state_dict` or any other name) and getter method (`load_state_dict` or any other name) and move it our from `streaming.py` file and create a separate file for it ?
2. Same thing for `session`, create a separate class for it and move out from `streaming.py` .

streaming/base/dataset/streaming.py
karan6181: Setting daemon to True will let the main thread exit even though the workers are blocking. This will solves the hanging issue if any.

streaming/base/dataset/streaming.py
knighton: Done

streaming/base/dataset/streaming.py
knighton: there's one download thread per worker, which is invoked at the start of the epoch (because even if we are persistent, the shards to download this worker could conceivably change for some reason), and exits as soon as it detects a new epoch has started (because only one epoch can be valid at a time)

knighton: we use a thread here because the main thread is busy iterating which sample to yield, while the thread updates which sample to download

in a perfect world, it might be faster to have it skip samples that some other thread is already downloading, so you'd track three indices: sample to yield, last sample that is downloaded, and sample that we are downloading

knighton: we would tune this to optimize the number of concurrent shard downloads to maximize throughput, which is backend-dependent, say local filesystem is probably approaching 1

knighton: Updated docstring?

streaming/base/dataset/streaming.py
knighton: Fixed

streaming/base/dataset/streaming.py
knighton: a user might get the number of samples from len(), then step what iter() returns with next() len() number of times

streaming/base/dataset/streaming.py
knighton: Fixed

streaming/base/dataset/streaming.py
knighton: Renamed variable, which is the worker offset of this node

streaming/base/shared.py
knighton: logging on lock acquire/release would be verbose and performance prohibitive, let's test this separately

streaming/base/world.py
knighton: Added explanatory comments

streaming/base/world.py
knighton: Added explanatory comments

streaming/base/world.py
knighton: Added explanatory comments

streaming/base/dataset/streaming.py
knighton: out of scope

streaming/base/dataset/streaming.py
knighton: fixed (FileNotFoundError)

streaming/base/dataset/streaming.py
knighton: fixed

streaming/base/dataset/streaming.py
knighton: done

streaming/base/dataset/streaming.py
knighton: kNiGhToN, jAmEs KnIgHtOn

knighton: done

streaming/base/dataset/streaming.py
knighton: renamed to self._worker_barrier

streaming/base/dataset/streaming.py
knighton: fixed

streaming/base/dataset/streaming.py
knighton: done

streaming/base/dataset/streaming.py
knighton: redesigned, daemon=True

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,623 @@
...
+            if self.iter_index < self.download_index:
+                yield self.sample_ids[self.iter_index]
+                self.iter_index += 1
+                continue
+            if self.is_stopped:
+                break
+            sleep(TICK)
+
+
+class Dataset(IterableDataset):
```
mvpatel2000, line 81: Do we want to rename to `StreamingDataset` based on poll in slack?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,623 @@
...
+            Iterator[int]: Each sample, having been downloaded.
+        """
+        while self.iter_index < self.total:
+            if self.iter_index < self.download_index:
+                yield self.sample_ids[self.iter_index]
+                self.iter_index += 1
+                continue
+            if self.is_stopped:
+                break
+            sleep(TICK)
```
mvpatel2000, line 78: why is this sleep necessary?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,623 @@
...
+
+        # Build the Index (for partitioning and mapping samples to shards).
+        self.shard_sizes = np.array([x.samples for x in self.shards])
+        self.index = Index(self.shard_sizes)
+
+        # Determine and distribute shuffle seed and shm prefix.
+        if shuffle_seed is None:
+            shuffle_seed = np.random.randint(1 << 60)
+        prefix_int = np.random.randint(1 << 60)
+        if 1 < world.num_ranks:
```
mvpatel2000, line 173: nit: can we do `world.num_ranks > 1`? a little more readable / consistent code style

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,623 @@
...
+
+        This is called on each copy of the dataset when resuming.
+
+        We just save the state to shared memory for workers to pick up when __iter__ is next
+        called. We use shm because changes to this copy of the dataset wouldn't be picked up by
+        persistent workers.
+
+        Args:
+            obj (Dict[str, Any]): The state.
+        """
```
mvpatel2000, line 615: Can we say something saying see `state_dict` for structure of state obj?

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,623 @@
...
+            'sample_in_epoch': offset + sample_in_epoch,
+        }
+
+    def load_state_dict(self, obj: Dict[str, Any]) -> None:
+        """Load a dict containing training state (called from non-worker process).
+
+        This is called on each copy of the dataset when resuming.
+
+        We just save the state to shared memory for workers to pick up when __iter__ is next
+        called. We use shm because changes to this copy of the dataset wouldn't be picked up by
```
mvpatel2000, line 610: can we override this fn and have another variant that takes `epoch`, `sample_in_epoch` as args?

mvpatel2000 (commented):
I understand about 47% of this PR, but the part I understand LGTM. Defer to Karan for approvals.

Only concern I have rn is a user could resume at a particular point, and then if composer loads a ckpt, it will override the user resumption. There's no way Composer rn can check if user has already resumed easily. I _think_ that's fine, but we should confirm. I'll ping Abhi.

streaming/base/dataloader.py
```diff
@@ -0,0 +1,87 @@
...
+
+from torch import Tensor
+from torch.utils.data import DataLoader
+from transformers.tokenization_utils_base import BatchEncoding
+
+from streaming.base.dataset import Dataset
+from streaming.base.world import World
+
+
+class StreamingDataLoader(DataLoader):
```
karan6181, line 16: When would user instantiate the `StreamingDataLoader` class? is it when user is using vanilla pytorch model (without composer) and wants to do checkpointing/resumption ??

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,623 @@
...
+            decompressing downloaded shards. If set to None, keep iff remote is local. Defaults to
+            ``None``.
+        download_retry (int): Number of download re-attempts before giving up. Defaults to ``2``.
+        download_timeout (float): Number of seconds to wait for a shard to download before raising
+            an exception. Defaults to ``60``.
+        validate_hash (str, optional): Optional hash or checksum algorithm to use to validate
+            shards. Defaults to ``None``.
+        shuffle_seed (int, optional): Seed for shuffling, or ``None`` for random seed. Defaults to
+            ``None``.
+        shuffle_world_size (int, optional): Canonical world size for shuffling. Defaults to
```
karan6181, line 111: Maybe also mentioned that the `shuffle_world_size` is useful when doing `resumption` !!!

streaming/base/dataset/streaming.py
```diff
@@ -0,0 +1,623 @@
...
+        batch_size (int, optional): Batch size of its DataLoader, which affects how the dataset is
+            partitioned over the workers. Defaults to ``None``.
+    """
+
+    def __init__(self,
+                 local: str,
+                 remote: Optional[str] = None,
+                 split: Optional[str] = None,
+                 shuffle: bool = False,
+                 predownload: Optional[int] = 100_000,
```
karan6181, line 122: is this number `100_000` realistic? Because this number is for every `num_workers` process right? For example, let say, you are doing a model training on a 1 node with 8 GPUs with 8 `num_workers` each per GPU, which means total 64 num_process per node, which leads to 64 * 100_000 = 6_400_000 number of samples !!! 

Should we lower the default number of `10_000` maybe !! Whats your thought on that! Also, leaving as it is causes no harm at all. Wanted to know your thoughts on choosing `100_000` !! Thanks!

streaming/base/shared.py
```diff
@@ -0,0 +1,149 @@
...
+    @flag.setter
+    def flag(self, flag: bool) -> None:
+        """Set property flag.
+
+        Args:
+            flag (bool): The flag value.
+        """
+        self._arr[2] = bool(flag)
+
+    def __call__(self, num_procs: int) -> None:
```
karan6181, line 113: Why pass the `num_procs` here? Can we pass the `num_procs` in the `__init__()` method of `class SharedBarrier` ? The shared barrier object will be initialized only for one cluster configuration and `world.num_workers` will not going to change in one cluster configuration run.

streaming/base/shared.py
```diff
@@ -0,0 +1,151 @@
...
+        self._arr = np.ndarray(3, buffer=self._shm.buf, dtype=np.int32)
+        self.num_enter = 0
+        self.num_exit = -1
+        self.flag = True
+
+    def __del__(self):
+        """Destructor clears array that references shm."""
+        try:
+            del self._arr
+            self._shm.close()
```
karan6181, line 56: I added the `close()` and `unlink()` method since looks like there was a memory link. I was getting a below warning message while working on a unit test

```
/opt/homebrew/Cellar/python@3.10/3.10.7/Frameworks/Python.framework/Versions/3.10/lib/python3.10/multiprocessing/resource_tracker.py:224: UserWarning: resource_tracker: There appear to be 1 leaked shared_memory objects to clean up at shutdown
```

Ref: https://docs.python.org/3/library/multiprocessing.shared_memory.html#multiprocessing.shared_memory.SharedMemory.close

streaming/base/dataset/streaming.py
knighton: it is waiting on a shard download which takes seconds, no use spinning

streaming/base/dataloader.py
```diff
@@ -0,0 +1,87 @@
...
+            int: Number of samples.
+        """
+        if isinstance(batch, (dict, BatchEncoding)):
+            for value in batch.values():
+                return len(value)
+            raise ValueError('Batch is empty')
+        elif isinstance(batch, Tensor):
+            return len(batch)
+        else:
+            return len(batch[0])
```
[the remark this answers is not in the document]
knighton: the batch is a sequence (e.g. x, y)

streaming/base/dataloader.py
```diff
@@ -0,0 +1,87 @@
...
+
+        Args:
+            batch (Any): The batch.
+
+        Returns:
+            int: Number of samples.
+        """
+        if isinstance(batch, (dict, BatchEncoding)):
+            for value in batch.values():
+                return len(value)
```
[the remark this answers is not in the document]
knighton: the batch is a dict (eg, {'x': x, 'y': y}

streaming/base/dataset/streaming.py
knighton: renamed

streaming/base/shared.py
```diff
@@ -0,0 +1,149 @@
...
+import numpy as np
+from filelock import FileLock
+
+TICK = 0.07
+
+
+class SharedBarrier:
+    """A barrier that works inter-process using a file lock and shared memory.
+
+    We set the number of processes (and thereby initialilze num_exit) on the first time this object
```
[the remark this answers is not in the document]
knighton: fixed

streaming/base/dataset/streaming.py
knighton: about 60k samples/shard in c4

streaming/base/shared.py
knighton: we don't know `num_workers` when per-rank copy of streaming dataset is initialiized

streaming/base/dataloader.py
knighton: yes

streaming/base/dataset/streaming.py
knighton: 100K per worker. no, workers do not generate shard files

streaming/__init__.py
```diff
@@ -6,10 +6,10 @@
...
-from streaming.base import (CSVWriter, Dataset, JSONWriter, LocalDataset, MDSWriter, TSVWriter,
-                            XSVWriter)
+from streaming.base import (CSVWriter, Dataset, JSONWriter, LocalIterableDataset, LocalMapDataset,
+                            MDSWriter, TSVWriter, XSVWriter)
 
 __all__ = [
-    'Dataset', 'CSVWriter', 'JSONWriter', 'MDSWriter', 'TSVWriter', 'XSVWriter', 'LocalDataset',
-    'vision', 'text'
+    'Dataset', 'CSVWriter', 'JSONWriter', 'MDSWriter', 'TSVWriter', 'XSVWriter',
+    'LocalIterableDataset', 'LocalMapDataset', 'vision', 'text'
```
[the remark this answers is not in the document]
knighton: yes

streaming/base/dataset/local_iterable.py
```diff
@@ -0,0 +1,76 @@
...
+from typing import Any, Dict, Iterator, Optional
+
+import numpy as np
+from torch.utils.data import IterableDataset
+
+from streaming.base.format import reader_from_json
+from streaming.base.index import Index
+
+
+class LocalIterableDataset(IterableDataset):
```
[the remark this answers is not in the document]
knighton: local iterable dataset is removed

streaming/__init__.py
```diff
@@ -6,10 +6,10 @@
 import streaming.text as text
 import streaming.vision as vision
 from streaming._version import __version__
-from streaming.base import (CSVWriter, Dataset, JSONWriter, LocalDataset, MDSWriter, TSVWriter,
-                            XSVWriter)
+from streaming.base import (CSVWriter, Dataset, JSONWriter, LocalIterableDataset, LocalMapDataset,
```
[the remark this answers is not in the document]
knighton: local map dataset provides a top lline for convergence and is needed to quantify the impact of approximate shuffle strategies

streaming/base/dataset/streaming.py
knighton: done

streaming/base/dataset/streaming.py
knighton: streaming repo has been ordering args LT < GT, let's future PR this

streaming/base/format/mds/reader.py
```diff
@@ -120,4 +120,8 @@ def get_sample_data(self, idx: int) -> bytes:
             begin, end = np.frombuffer(pair, np.uint32)
             fp.seek(begin)
             data = fp.read(end - begin)
+        if not data:
```
karan6181, line 123: While working on a test case, I found out that if the sample index is greater than it should be in a shard file, then `data` ends up being empty byte `b''` and `decode_sample()` methods fails with `ValueError: not enough values to unpack (expected 1, got 0)`. Added a check here, so that we have a proper error message. Now, it says `IndexError: Relative sample index 3 is not present in the shard.00000.mds file.` which is helpful to debug.

knighton:
better_downloading

streaming/base/dataloader.py
```diff
@@ -0,0 +1,92 @@
...
+        Args:
+            obj (Dict[str, Any]): The state.
+        """
+        if isinstance(self.dataset, StreamingDataset):
+            return self.dataset.load_state_dict(obj)
+
+    def __del__(self) -> None:
+        """Terminate the workers during cleanup."""
+        if self._iterator is not None:
+            self._iterator._shutdown_workers()  # type: ignore [reportGeneralTypeIssues]
```
mvpatel2000, line 92: I'm still surprised we don't need atexit here... but if it works, it works. *shrug*. Something to keep eye on in future though

streaming/base/dataset.py
```diff
@@ -1,364 +1,384 @@
...
-        self.index = Index(samples_per_shard, batch_size)
+        # Build the Index (for partitioning and mapping samples to shards).
+        self.shard_sizes = np.array([x.samples for x in self.shards])
+        self.index = Index(self.shard_sizes)
+
+        # Determine and distribute shuffle seed and shm prefix.
+        if shuffle_seed is None:
+            shuffle_seed = np.random.randint(1 << 60)
+        prefix_int = np.random.randint(1 << 60)
+        if 1 < world.num_ranks:
```
mvpatel2000, line 189: nit: slightly more consistent with code style elsewhere
```suggestion
        if world.num_ranks > 1:
```

mvpatel2000 (approved):
LFG 🚀 

Minor feature request: A frequent use case we're going to hit is being able to skip batches in event of loss spike. `load_state_dict` is a bit of a pain to use... it'd be nice if we could just add `set_epoch`, `set_sample_in_epoch`. Should just only edit elements 1 and 2 of `self._resume_shm`

streaming/base/dataset.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/37/commits/82fb0ad0ae95bd0bc3d993605221a1bf928755c3.

PR #43, opened 2022-10-14
Add CodeQL security scanner and Dependabot workflow

karan6181:
## Description
- Add CodeQL security scanner and Dependabot workflow
- Update the package dependency versions to be less strict.

## Issues
- Solves: CO-1248

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,49 +24,49 @@ classifiers = [
 ]
 
 install_requires = [
-    'boto3==1.24.37',
-    'Brotli==1.0.9',
-    'datasets==2.4.0',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -24,49 +24,49 @@
 ]
 
 install_requires = [
-    'boto3==1.24.37',
```
knighton, line 27: how did you figure out all these upgrades?

knighton (approved):

setup.py
karan6181: From Composer setup.py.

PR #50, opened 2022-10-17
Add Build publish to PyPI and create GitHub release workflow

karan6181:
## Description
- Add Build and publish to PyPI workflow
- Add create GitHub release workflow
- Update existing actions to latest version

## Usage
The `pypi-release.yaml` workflow will be kick in with the below commands
```
git tag -a "v0.0.2b" -m "pypi package versioning"
git push --tags
```

With the above 2 lines, the GitHub actions will do the following
- Run pre-commit for linting issue
- Run pytest to ensure it passes
- Create a pypi wheel, check package description on PyPI using `twine`, uploads the pypi wheel to pypi website.
- Create a GitHub Release


knighton (approved):

.github/workflows/precommit.yaml
```diff
@@ -16,13 +18,19 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    strategy:
+      matrix:
+        python_version:
+          - "3.9"
```
bandish-shah, line 24: It looks like the `install` workflow runs on Python 3.7, 3.8 and 3.9.  Why does this only run on 3.9?

.github/workflows/pytest.yaml
```diff
@@ -8,18 +9,25 @@ on:
...
+  workflow_call:
   workflow_dispatch: {}
 jobs:
   pytest:
     timeout-minutes: 30
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python_version:
+          - "3.9"
```
bandish-shah, line 21: It looks like the `install` workflow runs on Python 3.7, 3.8 and 3.9.  Why does this only run on 3.9?

.github/workflows/pypi-release.yaml
```diff
@@ -0,0 +1,77 @@
+name: Publish mosaicml-streaming to PyPI / GitHub
+
+on:
+  push:
+    tags:
+      - "v*"
```
bandish-shah, line 6: This will run on any tag created on `dev` or `main`.  Is that what you want?

.github/workflows/precommit.yaml
```diff
@@ -35,7 +43,7 @@ jobs:
         run: |
           set -ex
           python -m pip install --upgrade pip wheel
-          python -m pip install -e .[dev]
+          python -m pip install -e .${{ matrix.install_version }}
       - name: Linting
         run: |
           pre-commit run --all-files
```
bandish-shah, line 49: IIRC you have to do a `pre-commit install` before you can run or does this go ahead and install `pre-commit` as well?

.github/workflows/precommit.yaml
```diff
@@ -1,4 +1,5 @@
 name: pre-commit
+
```
bandish-shah, line 2: Should probably just call this workflow "Linting" or "Code Quality" or something.  Yes we're using `pre-commit` but that's more of a runner vs. being descriptive of what checks are actually being performed here.

bandish-shah (commented):
This looks great!  Few questions inline before I approve.

.github/workflows/precommit.yaml
karan6181: We want to make sure that the streaming installation works as expected on 3.7, 3.8 and 3.9. However, we are running the pre-commit workflow on the latest python to make sure it passes since it is just used for linting. Making it to run on a multiple python version is just adding more redundancy without having much benefit.

.github/workflows/pytest.yaml
karan6181: Same as above.

.github/workflows/precommit.yaml
karan6181: pre-commit is getting installed as part of [dev](https://github.com/mosaicml/streaming/blob/main/setup.py#L47) in setup.py file and all the dependency linting will get install inside command `pre-commit run --all-files`.

.github/workflows/pypi-release.yaml
bandish-shah: Scratch this, I forgot there was only a `main` branch in this repo.

.github/workflows/precommit.yaml
karan6181: Done in https://github.com/mosaicml/streaming/pull/50/commits/052061bd11fb8c286bb726c73bf10fec330e1398.

bandish-shah: We should rename this file as well to match the workflow name.

.github/workflows/precommit.yaml
bandish-shah: I think pyright has different linting rules per Python version so technically it isn't redundant.  It might be overkill though so let's keep it minimal for now and we can add additional versions if necessary.

bandish-shah (approved):
Couple of follow ups, approving to unblock.

.github/workflows/precommit.yaml
karan6181: Done in https://github.com/mosaicml/streaming/pull/50/commits/796fbdbbbf7cad3179d8651314f4f7f171d10ac0.

PR #57, opened 2022-10-23
Add Distributed test and supported multi device unittest

karan6181:
## Description
- Added Distributed test class to simulate the distributed workflow on a single machine
- Added multi-device dataloader unittest
- Run pytest sequentially due to mutli-processing issue with `-n auto`

### DIFF
--- a/tests/common/distributed.py
+++ b/tests/common/distributed.py
@@ -16,7 +16,6 @@ from _pytest.fixtures import FixtureRequest
 from _pytest.outcomes import Skipped
 from torch.multiprocessing import Process
 
-logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 # Worker timeout after the first worker has completed.
--- a/tests/common/distributed.py
+++ b/tests/common/distributed.py
@@ -1,0 +1,228 @@
+# Copyright 2022 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
+import contextlib
+import inspect
+import logging
+import os
+import socket
+import time
+from typing import Any, Dict, List
+
+import pytest
+import torch.distributed as dist
+import torch.multiprocessing as mp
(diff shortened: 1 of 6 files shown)


tests/common/distributed.py
```diff
@@ -0,0 +1,227 @@
...
+        # Fetch custom pytest marker `world_size`
+        # and override with a class parameter
+        for mark in getattr(request.function, 'pytestmark', []):
+            if mark.name == 'world_size':
+                self.world_size = mark.args[0]
+
+    def get_free_tcp_port(self) -> int:
+        """Get a free socket port to listen on."""
+        # from https://www.programcreek.com/python/?CodeExample=get+free+port
+        tcp = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
knighton, line 103: perhaps just socket.socket() as those args are the defaults? url mention not necessary :)

tests/test_distributed.py
```diff
@@ -0,0 +1,121 @@
...
+        assert dist.get_rank() < 4
+
+
+# @pytest.mark.skip
+class TestInit(DistributedTest):
+    world_size = 2
+
+    @pytest.mark.parametrize('batch_size', [128])
+    @pytest.mark.parametrize('drop_last', [False, True])
+    @pytest.mark.parametrize('num_workers', [1, 8])
```
knighton, line 44: Should we add zero here?

...('num_workers', [0, 1, 8])?

karan6181: Added in https://github.com/mosaicml/streaming/pull/57/commits/8741970777041113ee42870ed439f33131211835.

tests/common/distributed.py
karan6181: Changed in https://github.com/mosaicml/streaming/pull/57/commits/8741970777041113ee42870ed439f33131211835.

PR #59, opened 2022-10-24
Fixed rendering of pypi image

karan6181:
## Description
- pypi website does not support dark mode and does not understand GitHub tag. Hence, pypi website renders both the images (dark and light mode).
- This PR removes the dark image for pypi rendering since it only support light mode.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,24 @@ with open(os.path.join(os.path.dirname(__file__), 'streaming', '_version.py')) a
     exec(f.read(), version_globals, version_locals)
     streaming_version = version_locals['__version__']
 
+# Use repo README for PyPi description
+with open('README.md', 'r', encoding='utf-8') as fh:
+    long_description = fh.read()
+
+# Hide the content between <!-- SETUPTOOLS_LONG_DESCRIPTION_HIDE_BEGIN --> and
+# <!-- SETUPTOOLS_LONG_DESCRIPTION_HIDE_END --> tags in the README
+while True:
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -16,6 +16,24 @@
     exec(f.read(), version_globals, version_locals)
     streaming_version = version_locals['__version__']
 
+# Use repo README for PyPi description
+with open('README.md', 'r', encoding='utf-8') as fh:
```
knighton, line 20: i am curious why we need to specify r and utf-8 here?

karan6181: - `r`: we don't want to modify readme. The change that we are making is for pypi website.
- `utf-8`: UTF-8 is one of the most commonly used encodings, and Python often defaults to using it.

setup.py
```diff
@@ -16,6 +16,24 @@
...
+    start_tag = '<!-- SETUPTOOLS_LONG_DESCRIPTION_HIDE_BEGIN -->'
+    end_tag = '<!-- SETUPTOOLS_LONG_DESCRIPTION_HIDE_END -->'
+    start = long_description.find(start_tag)
+    end = long_description.find(end_tag)
+    if start == -1:
+        assert end == -1, 'there should be a balanced number of start and ends'
+        break
+    else:
+        assert end != -1, 'there should be a balanced number of start and ends'
+        long_description = long_description[:start] + long_description[end + len(end_tag):]
```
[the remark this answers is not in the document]
karan6181: using `slicing` instead of `split` since we don't want to split all the content and then join

PR #64, opened 2022-10-29
Mid Epoch Resumption Draft

mvpatel2000:
## Merge Checklist:

### General

### Tests


streaming/base/dataloader.py
```diff
@@ -0,0 +1,21 @@
...
+
+from torch.utils.data import DataLoader
+
+from streaming.base.dataset import Dataset
+
+class StreamingDataLoader(DataLoader):
+
+    def state_dict(self):
+        if isinstance(self.dataset, Dataset):
+            return self.dataset.state_dict(self._iterator._num_yielded)
```
knighton, line 16: Maybe it’s safer for future changes to do a pass-through __iter__ in this class and track num_yields ourself?

mvpatel2000: What do you mean by pass-through iter? Like, wrap the iterator? 
`self._iterator = StreamingIterator(self._iterator)`?

knighton: Something like…?

```
for batch in super().__iter__():
    self.num_yielded += len(batch[0])
```

But then I remembered dataloaders are only managing their set of workers, so this needs dist right?

mvpatel2000: hm... I'm not sure if this would work. You need to override the actual iterator instead of just the dataloader I think? it might interplay in a messy way otherwise. Would need to try it out. 

I think it should be fine without dist. IIRC dataloader runs on main thread, not per worker, and each rank surfaces a batch. We need to multiply by # of ranks for batch size, but that should be fine?

knighton: Sorry, meant avoiding cracking open DataLoader where possible, for example

```
class DataLoader(object):
    def __init__(self, num_batches, batch_size):
        self.num_batches = num_batches
        self.batch_size = batch_size

    def __iter__(self):
        self._num_yielded = 0
        for i in range(self.num_batches):
            a = i * self.batch_size
            yield list(range(a, a + self.batch_size))
            self._num_yielded += self.batch_size


loader = DataLoader(10, 4)
for batch in loader:
    print(batch)
print(loader._num_yielded)


class StreamingDataLoader(DataLoader):
    def __iter__(self):
        self.num_yielded = 0
        for batch in super().__iter__():
            yield batch
            self.num_yielded += len(batch)


loader = StreamingDataLoader(10, 4)
for batch in loader:
    print(batch)
print(loader.num_yielded)
```

karan6181:
Closing the PR as the actual implementation is in this PR: https://github.com/mosaicml/streaming/pull/37

PR #71, opened 2022-11-11
The Pile (conversion + streaming dataset)

knighton:


streaming/base/compression/compression.py
```diff
@@ -162,6 +162,7 @@ def decompress(self, data: bytes) -> bytes:
     'bz2': Bzip2,
     'gz': Gzip,
     'snappy': Snappy,
+    'zst': Zstandard,
     'zstd': Zstandard,
```
karan6181, line 166: Looks like a duplicate ? Do we still need `zstd` ?

streaming/base/compression/compression.py
```diff
@@ -142,7 +142,7 @@ def decompress(self, data: bytes) -> bytes:
 class Zstandard(LevelledCompression):
     """Zstandard compression."""
 
-    extension = 'zstd'
+    extension = 'zst'
```
karan6181, line 145: Why changing from `zstd` to `zst` ? From official webpage https://facebook.github.io/zstd/, looks like the abbreviation is `zstd`.

streaming/text/convert/pile.py
```diff
@@ -0,0 +1,150 @@
+# Copyright 2022 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""English Wikipedia 2020-01-01 streaming dataset conversion script."""
```
karan6181, line 4: Looks like a copy/paste ? need to update for pile dataset ?

streaming/text/convert/pile.py
```diff
@@ -0,0 +1,150 @@
...
+        for in_file, counts in zip(in_files, counters):
+            obj = {
+                'file': in_file,
+                'counts': counts,
+            }
+            print(json.dumps(obj, sort_keys=True))
+
+
+if __name__ == '__main__':
+    main(parse_args())
```
karan6181, line 150: Can you put a global comment on how one should download the raw pile dataset ? and how the directory structure looks like ? Otherwise, it would be confusing for the people on how this script!!

streaming/text/convert/pile.py
```diff
@@ -0,0 +1,150 @@
...
+        '--out_root',
+        type=str,
+        required=True,
+        help='Directory path to store the output dataset',
+    )
+    args.add_argument(
+        '--compression',
+        type=str,
+        default='zst:16',
+        help='Compression algorithm to use. Default: zst:16',
```
karan6181, line 41: How to know which compression algorithm to use ?

streaming/base/compression/compression.py
knighton: "zstd" for backwards compatibility

karan6181 (commented):
The `test_compression.py` is failing. It's an easy fix with `extension` name.

streaming/base/compression/compression.py
knighton: - we currently name compression algos by their filename extensions (for example, "gz" not "gzip")
- to my surprise, it turns out "zst" is the official extension, not "zstd" (search for ".zst" in https://github.com/facebook/zstd and elsewhere)

streaming/text/convert/pile.py
knighton: ah thx

streaming/text/convert/pile.py
knighton: - the user can judge the tradeoff for themself by running our compression benchmark on their own data
- compression levels seem to scale logarithmically in slowness
- i picked an aggressive level here because this dataset is quite large and this is a one-time calculation one on hand, and the highest number i could bear on the other hand

streaming/base/compression/compression.py
karan6181: Can you update all the convert scripts and all the test suite from `zstd` to `zst` so that future user don't use `zstd` anymore and in future we can remove `zstd` from out code base!!!

streaming/text/convert/pile.py
samhavens: Can we document this reasoning somewhere?

streaming/base/compression/compression.py
knighton: reverted usage of zst

streaming/text/convert/pile.py
knighton: Fixed

karan6181 (approved):
LGTM. Thanks!

PR #89, opened 2022-12-07
Tune predownload default from 100K to 20K.

knighton:


karan6181 (approved):
What is the reason behind downgrading the default value from 100k to 20k? Wanted to understand the pros and cons of this action! Thanks!

knighton:
Abandoning this PR.

100K samples/worker is a lot of slack, but text shards can be large enough that we should go this far out to ensure the next shard is in flight, worst case. Need to do more testing.

PR #90, opened 2022-12-08
Make StreamingDataset state_dict() more flexible

knighton:


streaming/base/dataset.py
```diff
@@ -666,23 +666,29 @@ def __iter__(self) -> Iterator[Dict[str, Any]]:
...
-    def state_dict(self, sample_in_epoch: int) -> Dict[str, Any]:
+    def state_dict(self, num_samples: int, from_beginning: bool) -> Dict[str, Any]:
         """Get a dict containing training state (called from non-worker process).
 
         This is called on rank zero.
 
         Args:
-            sample_in_epoch (int): The number of samples processed so far in the current epoch.
+            num_samples (int): The number of samples processed so far in the current epoch.
+            from_beginning (int): Whether we are counting from start of this epoch, or start of
```
karan6181, line 676:
```suggestion
            from_beginning (int): Whether we are counting samples from start of this epoch, or start of
```

streaming/base/dataloader.py
```diff
@@ -84,7 +86,7 @@ def load_state_dict(self, obj: Dict[str, Any]) -> None:
             obj (Dict[str, Any]): The state.
         """
         if isinstance(self.dataset, StreamingDataset):
-            return self.dataset.load_state_dict(obj)
+            self.dataset.load_state_dict(obj)
```
karan6181, line 89: Why no `return` statement ?

streaming/base/dataloader.py
```diff
@@ -72,7 +72,9 @@ def state_dict(self) -> Optional[Dict[str, Any]]:
         """
         if isinstance(self.dataset, StreamingDataset):
             world = World()
-            return self.dataset.state_dict(self.num_samples_yielded * world.num_ranks)
+            num_samples = self.num_samples_yielded * world.num_ranks
+            from_beginning = False
+            return self.dataset.state_dict(num_samples, from_beginning)
```
karan6181, line 77: How about `return self.dataset.state_dict(num_samples, False)` ?

streaming/base/dataloader.py
knighton: it returns None; is superfluous

streaming/base/dataset.py
```diff
@@ -666,23 +666,29 @@ def __iter__(self) -> Iterator[Dict[str, Any]]:
...
+                just this (potentially resumed) training run this epoch.
 
         Returns:
             Dict[str, Any]: The state.
         """
         world = World()
         epoch = self.next_epoch - 1
         epoch, offset = self._resume(world, epoch)
+        if from_beginning:
```
karan6181, line 685: maybe add a comment here saying that `from_beginning = True` is used when using `StreamingDataloader` ??

knighton: added note

streaming/base/dataloader.py
knighton: done

streaming/base/dataset.py

karan6181 (approved):

PR #97, opened 2022-12-20
Lazily instantiate the worker barrier in __iter__ (so it all pickles).

knighton:
## Merge Checklist:
_Put an `x` without space in the boxes that apply. If you are unsure about any checklist, please don't hesitate to ask. We are here to help! This is simply a reminder of what we are going to look for before merging your pull request._

### General
- [ ] I have read the [contributor guidelines](https://github.com/mosaicml/streaming/blob/main/CONTRIBUTING.md)
- [ ] This is a documentation change or typo fix. If so, skip the rest of this checklist.
- [ ] I certify that the changes I am introducing will be backward compatible, and I have discussed concerns about this, if any, with the MosaicML team.
- [ ] I have updated any necessary documentation, including [README](https://github.com/mosaicml/streaming/blob/main/README.md) and [API docs](https://github.com/mosaicml/streaming/tree/main/docs) (if appropriate).

### Tests
- [ ] I ran `pre-commit` on my change. (check out the `pre-commit` section of [prerequisites](https://github.com/mosaicml/streaming/blob/main/CONTRIBUTING.md#prerequisites))
- [ ] I have added tests that prove my fix is effective or that my feature works (if appropriate).
- [ ] I ran the tests locally to make sure it pass. (check out [testing](https://github.com/mosaicml/streaming/blob/main/CONTRIBUTING.md#running-tests))
- [ ] I have added unit and/or integration tests as appropriate to ensure backward compatibility of the changes.

PR #101, opened 2022-12-22
Set multiprocessing method as fork for Mac OS

karan6181:
## Description of changes:
- Set multiprocessing method as fork for Mac OS instead of `spawn` which is a default method.
- Streaming uses a FileLock for sharing the resources between ranks and workers which is Unpickleable except method `fork`.

## Issue #, if available:
CO-1589

If the multiprocessing method is `spawn` or `forkserver`, then running streaming application on Mac OS results in below error

```
File "/Users/sh/.pyenv/versions/3.10.6/lib/python3.10/multiprocessing/context.py", line 288, in _Popen
    return Popen(process_obj)
  File "/Users/sh/.pyenv/versions/3.10.6/lib/python3.10/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/Users/sh/.pyenv/versions/3.10.6/lib/python3.10/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/Users/sh/.pyenv/versions/3.10.6/lib/python3.10/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/Users/sh/.pyenv/versions/3.10.6/lib/python3.10/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
TypeError: cannot pickle '_thread.lock' object
```

This PR fixed this issue by using multiprocessing method as `fork`.

## Merge Checklist:

### DIFF
--- a/streaming/__init__.py
+++ b/streaming/__init__.py
@@ -3,9 +3,6 @@
 
 """MosaicML Streaming Datasets for cloud-native model training."""
 
-import multiprocessing as mp
-import sys
-
 import streaming.text as text
 import streaming.vision as vision
 from streaming._version import __version__
@@ -16,11 +13,3 @@ __all__ = [
     'StreamingDataLoader', 'StreamingDataset', 'CSVWriter', 'JSONWriter', 'MDSWriter', 'TSVWriter',
     'XSVWriter', 'LocalDataset', 'vision', 'text'
 ]
-
-IS_MACOS = sys.platform == 'darwin'
-
-# Set the multiprocessing start method to `fork` for MAC OS since
-# streaming uses a FileLock for sharing the resources between ranks
-# and workers which is Unpickleable except method `fork`.
-if IS_MACOS:
-    mp.set_start_method('fork', force=True)
--- a/streaming/__init__.py
+++ b/streaming/__init__.py
@@ -3,6 +3,9 @@
 
 """MosaicML Streaming Datasets for cloud-native model training."""
 
+import multiprocessing as mp
+import sys
+
 import streaming.text as text
 import streaming.vision as vision
 from streaming._version import __version__
@@ -13,3 +16,11 @@ __all__ = [
     'StreamingDataLoader', 'StreamingDataset', 'CSVWriter', 'JSONWriter', 'MDSWriter', 'TSVWriter',
     'XSVWriter', 'LocalDataset', 'vision', 'text'
 ]
+
+IS_MACOS = sys.platform == 'darwin'
+
+# Set the multiprocessing start method to `fork` for MAC OS since
+# streaming uses a FileLock for sharing the resources between ranks
+# and workers which is Unpickleable except method `fork`.
+if IS_MACOS:
+    mp.set_start_method('fork', force=True)
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -5,6 +5,7 @@
 
 import json
 import os
+import sys
 from enum import IntEnum
 from multiprocessing.shared_memory import SharedMemory
 from threading import Thread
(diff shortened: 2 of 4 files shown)


streaming/__init__.py
```diff
@@ -13,3 +16,11 @@
...
     'XSVWriter', 'LocalDataset', 'vision', 'text'
 ]
+
+IS_MACOS = sys.platform == 'darwin'
+
+# Set the multiprocessing start method to `fork` for MAC OS since
+# streaming uses a FileLock for sharing the resources between ranks
+# and workers which is Unpickleable except method `fork`.
+if IS_MACOS:
+    mp.set_start_method('fork', force=True)
```
knighton, line 26: this should probably go instead dataset __init__?

streaming/__init__.py
```diff
@@ -13,3 +16,11 @@
     'StreamingDataLoader', 'StreamingDataset', 'CSVWriter', 'JSONWriter', 'MDSWriter', 'TSVWriter',
     'XSVWriter', 'LocalDataset', 'vision', 'text'
 ]
+
+IS_MACOS = sys.platform == 'darwin'
+
+# Set the multiprocessing start method to `fork` for MAC OS since
+# streaming uses a FileLock for sharing the resources between ranks
+# and workers which is Unpickleable except method `fork`.
```
knighton, line 24: need to fix comment -- this works because fork doesn't pickle

streaming/__init__.py
karan6181: What if user uses `LocalDataset` ? Would they need this ?

knighton: yeah, in both dataset `__init__`s (or a common dataset base class but I personally don't think there is enough to warrant a base class?)

karan6181: Done in https://github.com/mosaicml/streaming/pull/101/commits/b3e9ee30820f97457712ce7e529e12464add9cbe.

streaming/__init__.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/101/commits/b3e9ee30820f97457712ce7e529e12464add9cbe.

dakinggg:
I think a library setting a forced mp context is a bit odd, and may have unintended side effects. Also CUDA does not support fork: https://pytorch.org/docs/stable/notes/multiprocessing.html#cuda-in-multiprocessing

karan6181:
> I think a library setting a forced mp context is a bit odd, and may have unintended side effects. Also CUDA does not support fork: https://pytorch.org/docs/stable/notes/multiprocessing.html#cuda-in-multiprocessing

For Mac OS, we have to change the mp method to `fork` to make it unpickleable. The Linux system by default comes with `fork` as a start method and hence we are not touching that. Also, we might have to re-visit the logic of FileLock and SharedBarrier in future to see if we can avoid setting mp method manually

dakinggg:
Yeah, I understand the reasoning, just saying that setting the mp context may cause weird downstream errors for users if they had code written that relied on spawn being used. I think James' other PR _should_ work, just something weird going on with the `SharedBarrier`...so, not a blocking comment, just a comment.

karan6181:
> Yeah, I understand the reasoning, just saying that setting the mp context may cause weird downstream errors for users if they had code written that relied on spawn being used. I think James' other PR _should_ work, just something weird going on with the `SharedBarrier`...so, not a blocking comment, just a comment.

Make sense. Once we figure out the solution to James PR (https://github.com/mosaicml/streaming/pull/97), we can revert the changes made in this PR so that we don't set mp start method manually.

dakinggg:
Yeah, I'm still looking at that PR 😅

knighton:
Okay, how about this: I'll try to debug my lazily instantiating filelock PR this morning and if that fails I'll merge this one

knighton:
The alternatives to FileLock and SharedMemory are (a) using the filesystem and having race conditions and other ugly problems or (b) rewriting that part as a c++ extension (eg `pthread_mutexattr_setpshared`) which would take time

PR #103, opened 2022-12-23
Removed cuda memory allocation which was causing CUDA OOM

karan6181:
## Description of changes:
- Removed the CUDA memory allocation which was used to synchronize the `shuffle_seed` and `prefix_int` across ranks.
- Based on how randomness works, the initial value `shuffle_seed` or `prefix_int` would be same across all the ranks, hence, there is no need for `dist.broadcast` operation.
- User needs to set the seed for getting the deterministic behavior on single node and multi-node training which is the same behavior as before.

## Issue #, if available:
CO-1591

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -119,8 +119,7 @@ class StreamingDataset(IterableDataset):
             an exception. Defaults to ``60``.
         validate_hash (str, optional): Optional hash or checksum algorithm to use to validate
             shards. Defaults to ``None``.
-        shuffle_seed (int, optional): Seed for Deterministic data shuffling. Defaults to
-            ``9176``.
+        shuffle_seed (int): Seed for Deterministic data shuffling. Defaults to ``9176``.
         num_canonical_nodes (int, optional): Canonical number of nodes for shuffling with resumption.
             Defaults to ``None``, which is interpreted as the number of nodes of the initial run.
         batch_size (int, optional): Batch size of its DataLoader, which affects how the dataset is
@@ -137,7 +136,7 @@ class StreamingDataset(IterableDataset):
                  download_retry: int = 2,
                  download_timeout: float = 60,
                  validate_hash: Optional[str] = None,
-                 shuffle_seed: Optional[int] = 9176,
+                 shuffle_seed: int = 9176,
                  num_canonical_nodes: Optional[int] = None,
                  batch_size: Optional[int] = None):
         self.local = local
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -8,17 +8,14 @@ import os
 from enum import IntEnum
 from multiprocessing.shared_memory import SharedMemory
 from threading import Thread
-from time import sleep, time
+from time import sleep
 from typing import Any, Dict, Iterator, Optional, Tuple
 
 import numpy as np
-import torch
 from filelock import FileLock
 from numpy.typing import NDArray
-from torch import distributed as tdist
 from torch.utils.data import IterableDataset
 
-from streaming.base import distributed as dist
 from streaming.base.compression import decompress
 from streaming.base.format import reader_from_json
 from streaming.base.format.base.reader import FileInfo
@@ -28,6 +25,7 @@ from streaming.base.partitioning import get_partitions
 from streaming.base.shared import SharedBarrier
 from streaming.base.shuffle import get_shuffle
 from streaming.base.storage import download
+from streaming.base.util import is_file_exist
 from streaming.base.world import World
 
 # Time to wait, in seconds.
@@ -152,9 +150,10 @@ class StreamingDataset(IterableDataset):
         self.download_timeout = download_timeout
         self.validate_hash = validate_hash or None
 
-        if tdist.is_available() and not tdist.is_initialized() and torch.cuda.is_available() and \
-                'RANK' in os.environ:
-            tdist.init_process_group('nccl')
+        if self.download_retry < 0:
+            raise ValueError('parameter `download_retry` can never be negative.')
+        if self.download_timeout < 0:
+            raise ValueError('parameter `download_timeout` can never be negative.')
 
         # Seed is set below.
         world = World()
@@ -168,7 +167,12 @@ class StreamingDataset(IterableDataset):
             filename = self._download_file(basename)
         else:
             filename = os.path.join(local, self.split, basename)  # pyright: ignore
-        dist.barrier()
+
+        # Everyone waits for the file to become populated.
+        is_file_exist(filename, TICK, self.download_timeout,
+                      f'{filename} file took too long to download')
+
+        # dist.barrier()
         obj = json.load(open(filename))
         if obj['version'] != 2:
             raise ValueError('Unsupported version')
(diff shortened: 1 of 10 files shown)


dakinggg:
QQ: Is it actually guaranteed that `prefix_int` and `shuffle_seed` are the same across all ranks (including multi node)? I think this may only be true if the main training script seeds randomness.

karan6181:
> QQ: Is it actually guaranteed that `prefix_int` and `shuffle_seed` are the same across all ranks (including multi node)? I think this may only be true if the main training script seeds randomness.

For single node with 8 GPUs, I have verified that the `prefix_int` and `shuffle_seed`  is same across all ranks by printing those variables from all the ranks. Multi-node testing is in-progress.

streaming/base/dataset.py
```diff
@@ -168,7 +167,12 @@ def __init__(self,
             filename = self._download_file(basename)
         else:
             filename = os.path.join(local, self.split, basename)  # pyright: ignore
-        dist.barrier()
+
+        # Everyone waits for the file to become populated.
+        is_file_exist(filename, TICK, self.download_timeout,
+                      f'{filename} file took too long to download')
+
+        # dist.barrier()
```
dakinggg, line 175: delete commented out code

streaming/base/dataset.py
```diff
@@ -168,7 +167,12 @@ def __init__(self,
             filename = self._download_file(basename)
         else:
             filename = os.path.join(local, self.split, basename)  # pyright: ignore
-        dist.barrier()
+
+        # Everyone waits for the file to become populated.
+        is_file_exist(filename, TICK, self.download_timeout,
```
dakinggg, line 172: suggest renaming this function to `wait_for_file_to_exist`

dakinggg (commented):
As discussed offline, suggest seeding the randomness explicitly or creating an rng instance so as not to rely on randomness being the same across nodes/ranks

streaming/base/dataset.py
```diff
@@ -152,9 +150,10 @@ def __init__(self,
         self.download_timeout = download_timeout
         self.validate_hash = validate_hash or None
 
-        if tdist.is_available() and not tdist.is_initialized() and torch.cuda.is_available() and \
-                'RANK' in os.environ:
-            tdist.init_process_group('nccl')
+        if self.download_retry < 0:
+            raise ValueError('parameter `download_retry` can never be negative.')
```
knighton, line 154: Capitalize

Maybe "Number of download retries must be nonnegative"?

streaming/base/dataset.py
```diff
@@ -152,9 +150,10 @@ def __init__(self,
...
+        if self.download_timeout < 0:
```
knighton, line 155: <= 0

Maybe "Download timeout must be greater than zero seconds"

streaming/base/util.py
```diff
@@ -18,3 +20,26 @@ def get_list_arg(text: str) -> List[str]:
         List[str]: Splits, if any.
     """
     return text.split(',') if text else []
+
+
+def is_file_exist(filename: str, sleep_time: float, timeout: float, err_msg: str) -> None:
```
knighton, line 25: sleep_time -> poll_interval?

streaming/base/dataset.py
knighton: Or "Download timeout (in seconds) must be greater than zero"

streaming/base/util.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/103/commits/bb6649673cbd20eeec770ef30d57668b3ed8afab.

streaming/base/dataset.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/103/commits/bb6649673cbd20eeec770ef30d57668b3ed8afab. I kept the parameter name for the user to know what to change.

streaming/base/dataset.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/103/commits/bb6649673cbd20eeec770ef30d57668b3ed8afab. I kept the parameter name for the user to know what to change.

streaming/base/dataset.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/103/commits/bb6649673cbd20eeec770ef30d57668b3ed8afab.

streaming/base/dataset.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/103/commits/bb6649673cbd20eeec770ef30d57668b3ed8afab.

karan6181:
> QQ: Is it actually guaranteed that `prefix_int` and `shuffle_seed` are the same across all ranks (including multi node)? I think this may only be true if the main training script seeds randomness.

Yes, as of now, for deterministic behavior, user needs to set the seeds. If the user sets the seeds, then the behavior is deterministic on single and multi-node.

dakinggg:
Should we set a default shuffle seed? Help the user avoid surprising outcomes, like having a different shuffle on each node. (Fwiw, I don't think our examples in the examples repo explicitly set the shuffle seed)

knighton:
Can we just broadcast seed over gloo or something? Non-deterministic is a bug

dakinggg:
Is there a problem with setting a default shuffle seed? To avoid adding unnecessary distributed stuff. Iirc having a separate gloo process group on mcloud is non trivial.

knighton:
Good points... sorry am in Guanajuato with terrible reception... cannot really field code reviews atm... defer to what y'all think is best

abhi-mosaic:
Would also prefer to avoid calling `torch.distributed` ! Greatly reduces the number of potential bugs. A default seed seems ok with me, unless someone can think of potential bugs?

karan6181:
> Should we set a default shuffle seed? Help the user avoid surprising outcomes, like having a different shuffle on each node. (Fwiw, I don't think our examples in the examples repo explicitly set the shuffle seed)

Yes, we can. As of today, we kept the `shuffle_seed` to `None` as a default value to get the data randomness behavior and if the user needs a deterministic behavior, then they would have to set the `seed` value in their code, although doesn't necessarily have to set `shuffle_seed` explicitly. If user is using composer, then they can directly add `composer.utils.reproducibility.seed_all(seed)` in their script. 

Currently, I don't think so of any application where user needs a data randomness behavior in every run, and if they would, then they can set the `shuffle_seed` to different values in every run to get the data randomness behavior. 

The commit https://github.com/mosaicml/streaming/pull/103/commits/a8ef59cc5203b07f2011d96c4fe06e17cd6cbc9f adds the default `shuffle_seed` value.

karan6181:
> Would also prefer to avoid calling `torch.distributed` ! Greatly reduces the number of potential bugs. A default seed seems ok with me, unless someone can think of potential bugs?

Agree. Adding a `torch.distributed` calls might surfaces some unknown bugs. With having a specific or a default `shuffle_seed`, user can get a deterministic behavior and non-deterministic behavior with different seed value in every run.

streaming/base/dataset.py
```diff
@@ -139,7 +137,7 @@ def __init__(self,
                  download_retry: int = 2,
                  download_timeout: float = 60,
                  validate_hash: Optional[str] = None,
-                 shuffle_seed: Optional[int] = None,
+                 shuffle_seed: Optional[int] = 9176,
```
dakinggg, line 140: Type is no longer optional I think

dakinggg (approved):
LGTM, thanks!

streaming/base/dataset.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/103/commits/69425859864b6df1340636f4eb65245f7ac84538.

PR #118, opened 2023-01-10
Make __getitem__ block to download shard if not present.

knighton:
## Description of changes:

Make `StreamingDataset[sample_id]` block to download the given sample's shard if it is not present, so that the dataset can be used lazily.

## Merge Checklist:

### General

### Tests

### DIFF
--- a/tests/test_laziness.py
+++ b/tests/test_laziness.py
@@ -1,0 +1,54 @@
+# Copyright 2022 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
+import shutil
+
+from streaming import MDSWriter, StreamingDataset
+
+
+def test_laziness():
+    num_samples = 100_000
+    local = 'my-local'
+    remote = 'my-remote'
+    columns = {'value': 'int'}
+    compression = None
+    hashes = None
+    size_limit = 10_000
+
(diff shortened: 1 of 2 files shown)


tests/test_laziness.py
```diff
@@ -0,0 +1,54 @@
...
+    compression = None
+    hashes = None
+    size_limit = 10_000
+
+    with MDSWriter(remote, columns, compression, hashes, size_limit) as out:
+        for i in range(num_samples):
+            sample = {'value': i}
+            out.write(sample)
+
+    # Verify __getitem__() downloads/accesses.
```
dakinggg, line 23: This one doesn't verify any downloading right? because you've passed `remote` as the first arg, so it only uses the already existing files there?

dakinggg (approved):
LGTM, have you done any manual testing to test the scenario where the locks are necessary (i.e. one worker accesses while another is downloading the necessary shard)? That sounds like a fairly complex unit test to write, but would be good to somehow verify that functionality.  Your comments in this PR is also making me think that we should have some basic regression testing for streaming...

tests/test_laziness.py
knighton: updated thanks

PR #149, opened 2023-02-08
Avoid common local folder when instantiating more than one StreamingDataset

karan6181:
## Description of changes:
- There is an issue where if the split is same for both `train` and `val` dataloader, which by default means using the same local folder for both `train` and `val` dataloader, the script non-deterministically fails with `FileNotFoundError: [Errno 2] No such file or directory: ‘/tmp/the-pile/val/shard.00000.mds'`. The issue doesn’t occur if using a separate local folder for `train` and `val` dataloader.
- The changes in this PR checks if the local split directory exists and empty or not and raise Exception if it exists and not empty.

## Issue #, if available:
CO-1667

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -164,10 +164,11 @@ class StreamingDataset(IterableDataset):
             raise ValueError(
                 'Parameter ``download_timeout`` (in seconds) must be greater than zero')
         local_split_path = os.path.join(self.local, self.split)
-        # Check if the directory exist and also not empty
-        if os.path.exists(local_split_path) and os.listdir(local_split_path):
+        # Check if the directory exist and is also not empty
+        if self.remote is not None and os.path.exists(local_split_path) and os.listdir(
+                local_split_path):
             raise ValueError(''.join([
-                f'Directory {local_split_path} already exist and not empty.',
+                f'Directory {local_split_path} already exist and is also not empty.',
                 'Either delete or empty the directory or provide a different `split` parameter.'
             ]))
 
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -151,11 +151,25 @@ class StreamingDataset(IterableDataset):
         self.download_timeout = download_timeout
         self.validate_hash = validate_hash or None
 
+        # Initialize the World context.
+        #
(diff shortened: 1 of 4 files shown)


streaming/base/dataset.py
```diff
@@ -151,11 +151,25 @@ def __init__(self,
...
+        # instantiate a World inside the StreamingDataset destructor.
+        self._rank_world = world = World()
+
         if self.download_retry < 0:
             raise ValueError('Parameter ``download_retry`` must be non-negative')
         if self.download_timeout < 0:
             raise ValueError(
                 'Parameter ``download_timeout`` (in seconds) must be greater than zero')
+        local_split_path = os.path.join(self.local, self.split)
+        # Check if the directory exist and also not empty
```
knighton, line 167: exists and is also

streaming/base/dataset.py
```diff
@@ -151,11 +151,25 @@ def __init__(self,
...
+        if os.path.exists(local_split_path) and os.listdir(local_split_path):
+            raise ValueError(''.join([
+                f'Directory {local_split_path} already exist and not empty.',
```
knighton, line 170: ditto

streaming/base/dataset.py
karan6181: Fixed in https://github.com/mosaicml/streaming/pull/149/commits/9294569eaca43c9abafa10541597c735c5f96104.

streaming/base/dataset.py
karan6181: Fixed in https://github.com/mosaicml/streaming/pull/149/commits/9294569eaca43c9abafa10541597c735c5f96104.

tests/test_reader.py
```diff
@@ -145,44 +143,10 @@ def test_reader_download_fail(mds_dataset_dir: Any, missing_file: str):
     assert exc_info.match(r'.*No such file or directory*')
 
 
-@pytest.mark.parametrize('created_ago', [0.5, 1.0])
```
karan6181, line 148: I removed this test since we are not testing anything significantly here

karan6181:
This is completed as part of PR https://github.com/mosaicml/streaming/pull/218.

PR #184, opened 2023-03-03
White-box weighted mixing of streaming datasets

knighton:
## Description of changes:

Add the ability to stream from multiple datasets, mixing them.

- The serialization format, designed for unified datasets, stays the same
- The StreamingDataset interface, designed for unified datasets, continues working as is
- We add an optional `streams` parameter to StreamingDataset which takes one or more sub-datasets which each have a remote (or no remote), local (or use a temp dir), proportion/repeat/samples, and additional optional download knobs. The download knobs passed to StreamingDataset are treated as defaults for the streams
- You can mix datasets by defining each either relatively (`proportion`) or absolutely (`repeat` or `samples` or none of them to sample 1:1). If using proportions, also provide `samples_per_epoch` to StreamingDataset
- During StreamingDataset init, we load each Stream’s index, concat their shards into a single listing, map streams <-> shards, calculate how to upsample/downsample, and then continue into the existing StreamingDataset init with that, treating this as mostly a bolted-on interface change

Todos:
- [x] dataset mixing interface
- [x] load multiple datasets together
- [x] global random access
- [x] upsampling/downsampling sub-datasets on `__iter__`
- [x] update usage of `keep_zip` when False
- [ ] ~dropping cold shards: `keep_raw` `keep_raw_interval`~

Core logic:

```py
        # Sample each shard of each stream according to their proportions/repeats/samples. This
        # gives us the resampled size of each underlying shard, and a mapping from each fake "big"
        # sample ID to its underlying "small" sample ID.
        pick_per_shard, small_per_big = self._resample_streams(epoch)

        # Partition the global sample space (of resampled "big" sample IDs) into a tensor of shape
        # (num physical nodes, ranks per node, workers per rank, batches per worker, samples per
        # batch) such that we have an elastically deterministic sample order.
        big_ids = get_partitions(self.samples_per_epoch, self.num_canonical_nodes, world.num_nodes,
                                 world.ranks_per_node, world.workers_per_rank, self.batch_size,
                                 sample_in_epoch)

        # If we need to shuffle, shuffle in a node-aware and *underlying* shard-aware way.
        if self.shuffle:
            shuffle = get_shuffle(self.shuffle_algo, pick_per_shard, self.num_canonical_nodes,
                                  self.shuffle_seed, epoch)
            big_ids = np.where(big_ids != -1, shuffle[big_ids], -1)

        # Now that we have partitioning and shuffled with hallucinated "big" sample IDs, we don't
        # need them anymore, and can convert back to underlying "small" sample IDs.
        return np.where(big_ids != -1, small_per_big[big_ids], -1)
```

## Merge Checklist:

### General

### Tests


knighton:
Core logic for up/down-sampling sub-datasets, called within `__iter__`:


```py
            # Sample each shard of each stream according to their proportions/repeats/samples. This
            # gives us the resampled size of each underlying shard, and a mapping from each fake
            # "big" sample ID to its underlying "small" sample ID.
            pick_per_shard, small_per_big = self._resample_streams(epoch)

            # Partition the global sample space (of resampled "big" sample IDs) into a tensor of
            # shape (num physical nodes, ranks per node, workers per rank, batches per worker,
            # samples per batch) such that we have an elastically deterministic sample order.
            big_ids = get_partitions(self.samples_per_epoch, self.num_canonical_nodes,
                                     world.num_nodes, world.ranks_per_node, world.workers_per_rank,
                                     self.batch_size, sample_in_epoch)

            # If we need to shuffle, shuffle in a node-aware and *underlying* shard-aware way.
            if self.shuffle:
                shuffle = get_shuffle(self.shuffle_algo, pick_per_shard, self.num_canonical_nodes,
                                      self.shuffle_seed, epoch)
                big_ids = np.where(big_ids != -1, shuffle[big_ids], -1)

            # Now that we have partitioning and shuffled with hallucinated "big" sample IDs, we
            # don't need them anymore, and can convert back to underlying "small" sample IDs.
            sample_ids = np.where(big_ids != -1, small_per_big[big_ids], -1)
```

streaming/base/dataset.py
```diff
@@ -367,201 +455,155 @@ def _get_progress(self, world: World) -> Tuple[int, int]:
...
+        if self.shuffle:
+            shuffle = get_shuffle(self.shuffle_algo, pick_per_shard, self.num_canonical_nodes,
+                                  self.shuffle_seed, epoch)
+            big_ids = np.where(big_ids != -1, shuffle[big_ids], -1)
+
+        # Now that we have partitioning and shuffled with hallucinated "big" sample IDs, we
+        # don't need them anymore, and can convert back to underlying "small" sample IDs.
+        return np.where(big_ids != -1, small_per_big[big_ids], -1)
+
+    def _get_worker_samples(self, world: World, epoch: int,
```
dakinggg, line 548: any particular reason you chose to switch to shared memory for this logic?

streaming/base/stream.py
```diff
@@ -0,0 +1,312 @@
...
+
+    Args:
+        remote (str, optional): Remote path or directory to download the dataset from. If ``None``,
+            its data must exist locally. Defaults to ``None``.
+        local (str, optional): Local working directory to download shards to. This is where shards
+            are cached while they are being used. Uses a temp directory if not set. Defaults to
+            ``None``.
+        split (str, optional): Which dataset split to use, if any. If provided, we stream from/to
+            the ``split`` subdirs of  ``remote`` and ``local``. Defaults to ``None``.
+        proportion (float, optional): How much to upsample or downsample this dataset, as the
```
dakinggg, line 59: can you rewrite the documentation for this param so it is more clear how it relies on the logic in the encapsulating `StreamingDataset`? i.e. don't use "we" or "any of them"

streaming/base/stream.py
```diff
@@ -0,0 +1,312 @@
...
+            raise ValueError('Remote and/or local path must be provided')
+
+        self.remote = remote
+
+        self._local = local
+        self.local = local or mkdtemp()
+
+        self.split = split or ''
+
+        if proportion is not None and repeat is not None:
```
dakinggg, line 105: the three options are all incompatible with each other, right? not just these two?

streaming/base/stream.py
```diff
@@ -0,0 +1,312 @@
...
+        if self._download_timeout is None:
+            self.download_timeout = default.download_timeout
+        if self.validate_hash is None:
+            self.validate_hash = default.validate_hash or None
+        if self._keep_zip is None:
+            self.keep_zip = default.keep_zip
+        if self._keep_raw is None:
+            self.keep_raw = default.keep_raw
+
+    def _download_file(self, basename: str) -> str:
```
dakinggg, line 170: are all these functions simply copied from where they used to live?

dakinggg (commented):
Mostly reviewed for overall design, which I think looks good. The one thing that is a bit confusing is that the `proportion` arg of `Stream` relies on setting `samples_per_epoch` in `StreamingDataset`. I'm not thinking of a way to avoid this, so I guess it just comes down to really thoroughly documenting that interdependence. Can review for details once there is some testing/examples.

streaming/base/dataset.py
```diff
@@ -307,15 +402,6 @@ def __len__(self) -> int:
         """
         return self.index.get_samples_per_device()
 
-    def _set_canonical_num_nodes(self, world: World):
```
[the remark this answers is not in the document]
knighton: a two-line method which was called exactly twice, from another internal method

not worth splitting out

streaming/base/dataset.py
knighton: - staying in memory is much faster than writing out potentially many GBs of data to disk and reading it back in parts
- this is especially the case when world size = 1
- users with slow disk have complained about this
- I originally went with the file method in the interest of getting things done simply and quickly, but have finally had time to do it right

streaming/base/stream.py
knighton: yep

I moved the lower-level downloading/validating/decompressing out to the streams, because it depends on various stream-specific arguments. thankful to be reducing the size of the core StreamingDataset

I am not sure about where the shard_states logic (the layer above that) should ultimately live -- hopefully this will shake out well when adding dropping soon-to-be cold shards

knighton:
> The one thing that is a bit confusing is that the proportion arg of Stream relies on setting samples_per_epoch in StreamingDataset. I'm not thinking of a way to avoid this, so I guess it just comes down to really thoroughly documenting that interdependence.

how about we have samples_per_epoch default to underlying samples per epoch if using proportions?

streaming/base/dataset.py
```diff
@@ -93,87 +86,154 @@ def __iter__(self) -> Iterator[int]:
...
+                         keep_raw=keep_raw)
+
+        # Normalize to a list of Streams.
+        if streams:
+            for stream in streams:
+                stream.apply_default(default)
+        else:
+            streams = [default]
+
+        # Validate proportions or repeats. Streams must each set at most one of proportion, repeat,
```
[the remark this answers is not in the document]

streaming/base/stream.py
knighton: fixed

streaming/base/stream.py
knighton: how about now

streaming/base/stream.py
```diff
@@ -0,0 +1,315 @@
...
+                 keep_raw: Optional[bool] = None) -> None:
+        self.remote = remote
+        self._local = local
+        self.local = local or mkdtemp()
+        self.split = split or ''
+
+        has_proportion = proportion is not None
+        has_repeat = repeat is not None
+        has_samples = samples is not None
+        if not (0 <= has_proportion + has_repeat + has_samples <= 1):
```
karan6181, line 105: Can we move all the validation checks to a separate method called `def _validate()` or something like that ? The same method can also be called in `StreamingDataset` `init()` method since we do some validation check there as well (For example, `if not (0 <= has_proportion + has_repeat + has_samples <= 1):` )!!

streaming/base/dataset.py
```diff
@@ -367,201 +456,166 @@ def _get_progress(self, world: World) -> Tuple[int, int]:
...
-        if os.path.isfile(raw_filename):
-            return
+        # Do expensive work that may use all the cores just once, in the local leader.
+        if world.is_local_leader:
+            # Generate the global arrangement of samples for this epoch.
+            sample_ids = self._generate_epoch(world, epoch, sample_in_epoch)
+
+            # Save the generated epoch shape to shared memory.
+            name = f'{self._prefix}_epoch_shape'
+            size = 5 * np.int64().nbytes
```
karan6181, line 579: Good to document this magic `5` number !!

streaming/base/dataset.py
```diff
@@ -367,201 +456,166 @@ def _get_progress(self, world: World) -> Tuple[int, int]:
...
 
-        # Is compression used?
-        if zip_info:
-            # Download the compressed form if missing.
-            zip_filename = os.path.join(self.local, self.split, zip_info.basename)
-            if not os.path.isfile(zip_filename):
-                self._download_file(zip_info.basename)
+            # Load the generated epoch shape from shared memory.
+            name = f'{self._prefix}_epoch_shape'
+            size = 5 * np.int64().nbytes
```
karan6181, line 597: Same here!!

streaming/base/dataset.py
```diff
@@ -367,201 +456,166 @@ def _get_progress(self, world: World) -> Tuple[int, int]:
...
 
-        Args:
-            shard_id (int): Shard ID.
-        """
-        reader = self.shards[shard_id]
-        for raw_info, zip_info in reader.file_pairs:
-            self._download_shard_part(raw_info, zip_info, reader.compression)
+        # Now clean up after ourselves.
+        if world.is_local_leader:
+            shape_shm.unlink()
```
karan6181, line 615: I am afraid `unlink()` method gets called by `num_workers` processes from dataloader and `unlink()` should only be called once per processes, otherwise it will throw some nasty error. Should the cleanup thing move as part of `__del__()` method ??

Also, you might want to have `close()` method first and then `unlink()` for proper cleanup.

tests/test_mixing.py
```diff
@@ -0,0 +1,185 @@
...
+def walk(dataset: StreamingDataset) -> List[int]:
+    return [sample['value'] for sample in dataset]
+
+
+def float_eq(a: float, b: float) -> bool:
+    return abs(a - b) < 1e-6
+
+
+def test_mixing():
+    root = '/tmp/foo'
```
karan6181, line 20: you can use `local_remote_dir: Tuple[str, str]` which creates a tmp directory for you and also deletes automatically when test finishes. Please take a look at `test_upload.py` file for example.

streaming/base/dataset.py
knighton: refactored

streaming/base/dataset.py
knighton: refactored

tests/test_mixing.py
knighton: fixed

streaming/base/stream.py
knighton: eh, it's complex and wouldn't be able to be moved entirely to one side or the other...

knighton:
Also confirmed that curves overlap when same shuffling algorithm

karan6181 (approved):
LGTM.

PR #215, opened 2023-03-29
Register atexit handler for resource cleanup

karan6181:
## Description of changes:
- Register atexit handler for resource cleanup.
  - The `atexit` module provides a simple interface to register functions to be called when a program closes down normally. Functions registered are automatically executed upon normal interpreter termination.
  - However, whenever a program is killed by a signal not handled by Python, when `os.exit()` is called, or Python fatal internal error is detected, the functions registered via this module `atexit` are not executed.
- Limit some of the test param values which tests the same thing so don't need to run tests with extra params.
- Added `seed` value to some of the tests for randomness.

## Issue #, if available:
STR-25

## Merge Checklist:

### General

### Tests

### DIFF
--- a/tests/test_reader.py
+++ b/tests/test_reader.py
@@ -39,9 +39,10 @@ def mds_dataset_dir():
 @pytest.mark.parametrize('batch_size', [None, 1, 2])
 @pytest.mark.parametrize('remote_arg', ['none', 'same', 'different'])
 @pytest.mark.parametrize('shuffle', [False, True])
+@pytest.mark.parametrize('seed', [5151])
 @pytest.mark.usefixtures('mds_dataset_dir')
 def test_dataset_sample_order(mds_dataset_dir: Any, batch_size: int, remote_arg: str,
-                              shuffle: bool):
+                              shuffle: bool, seed: int):
     num_samples = 117
     remote_dir, local_dir = mds_dataset_dir
     if remote_arg == 'none':
@@ -58,7 +59,8 @@ def test_dataset_sample_order(mds_dataset_dir: Any, batch_size: int, remote_arg:
     dataset = StreamingDataset(local=local_dir,
                                remote=remote_dir,
                                shuffle=shuffle,
-                               batch_size=batch_size)
+                               batch_size=batch_size,
+                               shuffle_seed=seed)
 
     # Test basic sample order
     rcvd_samples = 0
@@ -89,7 +91,7 @@ def test_dataset_sample_order(mds_dataset_dir: Any, batch_size: int, remote_arg:
 
 
 @pytest.mark.parametrize('batch_size', [None, 1, 2])
-@pytest.mark.parametrize('seed', [987])
+@pytest.mark.parametrize('seed', [8988])
 @pytest.mark.parametrize('shuffle', [False, True])
 @pytest.mark.usefixtures('mds_dataset_dir')
 def test_dataset_determinism(mds_dataset_dir: Any, batch_size: int, seed: int, shuffle: bool):
@@ -129,8 +131,9 @@ def test_dataset_determinism(mds_dataset_dir: Any, batch_size: int, seed: int, s
     'missing_file',
     ['index'],
 )
+@pytest.mark.parametrize('seed', [7777])
 @pytest.mark.usefixtures('mds_dataset_dir')
-def test_reader_download_fail(mds_dataset_dir: Any, missing_file: str):
+def test_reader_download_fail(mds_dataset_dir: Any, missing_file: str, seed: int):
     remote_dir, local_dir = mds_dataset_dir
 
     if missing_file == 'index':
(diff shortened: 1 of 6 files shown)


tests/test_reader.py
```diff
@@ -181,13 +188,17 @@ def test_reader_after_crash(mds_dataset_dir: Any, created_ago: float, download_t
 )
 @pytest.mark.usefixtures('mds_dataset_dir')
 @pytest.mark.parametrize('index', [17])
-def test_reader_getitem(mds_dataset_dir: Any, share_remote_local: bool, index: int):
+@pytest.mark.parametrize('seed', [5566])
```
dakinggg, line 191: is all this seed setting because every test has to use a different seed or has conflicts?

tests/test_streaming.py
```diff
@@ -15,13 +15,14 @@
 @pytest.mark.parametrize('batch_size', [128])
 @pytest.mark.parametrize('drop_last', [False, True])
 @pytest.mark.parametrize('shuffle', [False, True])
-@pytest.mark.parametrize('num_workers', [0, 1, 8])
+@pytest.mark.parametrize('num_workers', [0, 4])
```
dakinggg, line 18: why change this one? probably still good to test 1 worker?

tests/test_streaming.py
```diff
@@ -73,10 +74,10 @@ def test_dataloader_single_device(local_remote_dir: Tuple[str, str], batch_size:
         assert len(set(sample_order)) == num_samples
 
 
-@pytest.mark.parametrize('batch_size', [None, 1, 2, 4])
-@pytest.mark.parametrize('seed', [987])
+@pytest.mark.parametrize('batch_size', [1, 4])
```
dakinggg, line 77: same, why change?

tests/test_streaming.py
```diff
@@ -73,10 +74,10 @@ def test_dataloader_single_device(local_remote_dir: Tuple[str, str], batch_size:
...
+@pytest.mark.parametrize('seed', [1111])
 @pytest.mark.parametrize('shuffle', [False, True])
-@pytest.mark.parametrize('num_workers', [0, 1, 8])
+@pytest.mark.parametrize('num_workers', [0, 8])
```
dakinggg, line 80: same

tests/test_streaming.py
```diff
@@ -118,11 +117,11 @@ def test_dataloader_determinism(mds_dataset_dir: Any, batch_size: int, seed: int
...
 
 
-@pytest.mark.parametrize('batch_size', [1, 2, 4])
-@pytest.mark.parametrize('seed', [987])
+@pytest.mark.parametrize('batch_size', [1, 4])
+@pytest.mark.parametrize('seed', [2222])
 @pytest.mark.parametrize('shuffle', [False])
 @pytest.mark.parametrize('drop_last', [False, True])
-@pytest.mark.parametrize('num_workers', [0, 1, 8])
+@pytest.mark.parametrize('num_workers', [0, 8])
```
dakinggg, line 124: same

tests/test_streaming.py
```diff
@@ -155,10 +154,10 @@ def test_dataloader_sample_order(mds_dataset_dir: Any, batch_size: int, seed: in
     assert expected_sample_order == sample_order
 
 
-@pytest.mark.parametrize('batch_size', [1, 2, 4])
-@pytest.mark.parametrize('seed', [987])
+@pytest.mark.parametrize('batch_size', [1, 4])
+@pytest.mark.parametrize('seed', [3456])
 @pytest.mark.parametrize('shuffle', [False, True])
-@pytest.mark.parametrize('num_workers', [0, 1, 4, 8])
+@pytest.mark.parametrize('num_workers', [0, 4])
```
dakinggg, line 160: same

dakinggg (commented):
This looks good to me..can you run some of the regression tests? like resnet run and gpt run?

tests/test_reader.py
karan6181: no conflicts. Added seed value to the tests for sample order randomness.

tests/test_streaming.py
karan6181: `num_workers` > `0` spins up additional child workers. So `1` means it will spin up 1 child worker and 4 mean 4 child workers. since the underlying behavior is the same (spinning up child workers) so thought of only test with two scenario `0` num_workers and any ( in this case `4`).

tests/test_streaming.py
karan6181: see my above comment. Thanks!

tests/test_streaming.py
karan6181: see my above comment. Thanks!

tests/test_streaming.py
karan6181: see my above comment. Thanks!

tests/test_streaming.py
karan6181: see my above comment. Thanks!

tests/test_streaming.py
dakinggg: yeah, i'd like to keep the test around for `1` worker in all cases. I think sometimes 1 is different from >1 in a meaningful way.

karan6181:
> This looks good to me..can you run some of the regression tests? like resnet run and gpt run?

Yes, I ran Resnet model on ImageNet (1 node and 2 nodes) and GPT LLM 125m (1 node) and it ran successfully.

dakinggg (approved):

PR #218, opened 2023-04-03
Refactor StreamingDataset shared memory prefix setup 

knighton:
## Description of changes:

- Refactor and isolate/contain the weird dist/shm setup stuff.
- Walk linearly; we don't need to use shuffle_seed.
- Check for local working directory collisions.

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -7,20 +7,17 @@ import json
 import os
 from enum import IntEnum
 from threading import Thread
-from time import sleep, time
+from time import sleep
 from typing import Any, Dict, Iterator, Optional, Sequence, Tuple
 
 import numpy as np
-import torch
-import torch.distributed as dist
 from filelock import FileLock
 from numpy.typing import NDArray
 from torch.utils.data import IterableDataset
 
-from streaming.base.distributed import barrier
 from streaming.base.index import Index
 from streaming.base.partition import get_partitions
-from streaming.base.shared import CreateSharedMemory, SharedBarrier
+from streaming.base.shared import CreateSharedMemory, SharedBarrier, get_shm_prefix
 from streaming.base.shuffle import get_shuffle
 from streaming.base.stream import Stream
 from streaming.base.util import TICK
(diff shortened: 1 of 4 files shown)


streaming/base/shared.py
```diff
@@ -268,3 +272,79 @@ def cleanup(self):
...
+                shm = CreateSharedMemory(name, False).shm
+            except:
+                break
+            size = np.frombuffer(shm.buf[:4], np.int32)[0]
+            data = bytes(shm.buf[4:size])
+            text = data.decode('utf-8')
+            their_locals = text.split('\0')
+            both = set(my_locals) & set(their_locals)
+            if both:
+                raise ValueError(f'Reused local working directory: {sorted(my_locals)} vs ' +
```
karan6181, line 306: What is the suggestion here for the user? What they need to do to not see the same error message again ?

streaming/base/shared.py
```diff
@@ -268,3 +272,79 @@ def cleanup(self):
...
+    """
+    # Local leader goes first, checking and registering.
+    if world.is_local_leader:
+        # Local leader walks the existing shm prefixes starting from zero, verifying that there is
+        # no local working directory overlap.
+        for prefix_int in range(10**6):
+            prefix = f'{prefix_int:06}'
+            name = f'{prefix}_locals'
+            try:
+                shm = CreateSharedMemory(name, False).shm
```
karan6181, line 297: Can you add a comment here saying if the shared_memory doesn't exist then it will throw FileNotFoundError which means we found the unique shared_memory prefix ?

streaming/base/shared.py
```diff
@@ -268,3 +272,79 @@ def cleanup(self):
             pass
         finally:
             resource_tracker.unregister = original_rtracker_unreg
+
+
+def get_shm_prefix(my_locals: List[str], world: World) -> Tuple[str, SharedMemory]:
+    """Register or lookup our shared memory prefix.
+
+    Args:
+        my_locals (List[str]): Local working dir of each stream, relative to /. We need to verify
```
karan6181, line 281: What happens when `my_locals` is `[]` ? Meaning if someone is not using `Streams()`!

tests/test_streaming.py
```diff
@@ -98,6 +98,9 @@ def test_dataloader_determinism(mds_dataset_dir: Any, batch_size: int, seed: int
     for batch in dataloader:
         sample_order.extend(batch['id'][:])
 
+    del dataloader
```
karan6181, line 101: Why deleting dataset and dataloader ?

streaming/base/dataset.py
```diff
@@ -306,46 +303,14 @@ def __init__(self,
...
-        if self._rank_world.num_ranks > 1:
-            if dist.is_available() and not dist.is_initialized():
-                is_dist_pg_initialized = True
-                dist.init_process_group(backend='nccl' if torch.cuda.is_available() and
-                                        dist.is_nccl_available() else 'gloo',
-                                        rank=world.rank,
-                                        world_size=world.num_ranks)
-            dist.barrier()
+        # Register/lookup our shared memory prefix and filelock root directory.
+        my_locals = [os.path.abspath(stream.local) for stream in streams]
```
karan6181, line 307: what happens if `streams` is empty ?

streaming/base/dataset.py
```diff
@@ -356,33 +321,35 @@ def __init__(self,
...
         # picked up by __iter__().
         self._resume_shm = None
 
         # Placeholder for a _PartitionState which tracks state during __iter__().
         self._partition_state = None
 
+    def __del__(self) -> None:
+        """Destructor, which releases its local working directories."""
+        try:
+            self._locals_shm.buf[:4] = np.int32(0).tobytes()
```
karan6181, line 349: you need to check if `_local_shm` is present or not, e.g. if hasattr(self, '_locals_shm') because if the streaming dataset crash before allocating `self._locals_shm` then `__del__` will fail with noattributeerror.

streaming/base/shared.py
knighton: They need to not reuse a local directory

streaming/base/shared.py
knighton: how about this?

streaming/base/shared.py
knighton: we have a check in streaming dataset init for that (also using just remote/local is rolled up into a stream):

```
        if bool(streams) == (bool(remote) or bool(local)):
            raise ValueError(
                'You must provide either "streams" or "remote"/"local", but not both -- ' +
                'that would be confusing')
```

streaming/base/dataset.py
knighton: streaming dataset init requires there to be at least one stream (or remote/local which is converted into a stream)

streaming/base/dataset.py

streaming/base/shared.py
```diff
@@ -268,3 +272,107 @@ def cleanup(self):
...
+    return set(text.split('\0'))
+
+
+def get_shm_prefix(my_locals: List[str], world: World) -> Tuple[str, SharedMemory]:
+    """Register or lookup our shared memory prefix.
+
+    Args:
+        my_locals (List[str]): Local working dir of each stream, relative to /. We need to verify
+            that there is no overlap with any other currently running StreamingDataset.
+        world (World): Our place in the world.
```
karan6181, line 312: maybe a better comment than "Our place in the world." ?

karan6181 (approved):

streaming/base/shared.py
```diff
@@ -268,3 +272,107 @@ def cleanup(self):
...
+
+    Returns:
+        Tuple[str, CreateSharedMemory]: Shared memory prefix and object. The name is required to be
+            very short for Mac OSX.
+    """
+    # Check my locals for overlap.
+    my_locals_set = set()
+    for dirname in my_locals:
+        if dirname in my_locals_set:
+            raise ValueError(f'Reused local working directory: {dirname}')
```
karan6181, line 322: Can we also extend the error message to `f'Reused local working directory: {dirname}. Provide a different one.'` ?

streaming/base/shared.py
```diff
@@ -268,3 +272,107 @@ def cleanup(self):
...
+            prefix = f'{prefix_int:06}'
+            name = f'{prefix}_locals'
+            try:
+                shm = CreateSharedMemory(name, False).shm
+            except:
+                break
+            their_locals_set = _unpack_locals(bytes(shm.buf))
+            both = my_locals_set & their_locals_set
+            if both:
+                raise ValueError(f'Reused local working directory: {sorted(my_locals_set)} vs ' +
```
karan6181, line 340: Can we also extend the error message to `f'Reused local working directory: {dirname}. Provide a different one.'` ?

streaming/base/shared.py
knighton: done

streaming/base/shared.py
knighton: done

streaming/base/shared.py
knighton: done

PR #219, opened 2023-04-08
Cold shard eviction

knighton:
## Description of changes:

Cold shard eviction. Keep filesystem usage under some limit. Do this by dropping least recently accessed shards as needed when downloading new shards. This is a general solution that works with random access.

- [x] Redesign shard states: `MISSING`, `DOWNLOADING`, `PRESENT` (evicting happens under the lock)
- [x] Initialize `shard_states` by walking and normalizing stream local dirs
- [x] Implement shard eviction
- [x] Redesign `StreamingDataset` keep-related args to add `cache_limit`, drop `keep_raw`
- [x] Bookkeeping for shard access times and tracking disk usage
- [x] Automatically evict shards based on disk usage limit

## Merge Checklist:

### General

### Tests


streaming/base/dataset.py
```diff
@@ -277,86 +276,80 @@ def __init__(self,
...
+
+            # Init each stream's local dir, discovering which shards are present.
+            are_shards_present = []
+            for stream in self.streams:
+                stream_shards = stream.get_shards(world)
+                are_shards_present += stream.init_local_dir(stream_shards)
+
+            # Use that to init shard states, shard access times, and cache usage.
+            self.cache_usage = 0
+            for shard_id, is_shard_present in enumerate(are_shards_present):
```
dakinggg, line 331: is this shard id consistent?

streaming/base/dataset.py
```diff
@@ -366,21 +359,41 @@ def __del__(self) -> None:
...
+            int: Cache usage.
+        """
+        return int(self._cache_usage.get())
+
+    @cache_usage.setter
+    def cache_usage(self, cache_usage: int) -> None:
+        """Set the cache usage.
+
+        Args:
+            cache_usage (int): Cache usage.
```
dakinggg, line 394: what is cache usage? units?

streaming/base/dataset.py
```diff
@@ -709,27 +834,22 @@ def __getitem__(self, sample_id: int) -> Any:
...
+            self._shard_access_times[shard_id] = time()
         except:
-            # Get handles to the shared shard states array and its protective file lock.
-            lock, shard_states = self._get_shard_states()
-
-            # Download the shard if not already being downloaded. Block if download in progress.
-            self._download_or_skip_shard(lock, shard_states, shard_id, True)
-
-            # Finally, access the sample.
+            # Proper path: shard wasn't present. Download the shard or wait for its download to
```
dakinggg, line 846: add to comment that download shard sets the access time, i was confused for a sec why the try sets the time and the except doesn't

streaming/base/format/base/reader.py
```diff
@@ -63,6 +64,145 @@ def __len__(self) -> int:
         """
         return self.samples
 
+    def init_local_dir(self, ls: Set[str], keep_zip: bool) -> bool:
+        """Bring what shard files are present to a consistent state, returning whether prsent.
```
dakinggg, line 68:
```suggestion
        """Bring what shard files are present to a consistent state, returning whether present.
```

streaming/base/format/base/reader.py
```diff
@@ -63,6 +64,145 @@ def __len__(self) -> int:
...
+
+        Args:
+            ls (Set[str]): The listing of all files under dirname/[split/]. This is listed once
```
dakinggg, line 71: more descriptive var name please

streaming/base/format/base/reader.py
```diff
@@ -63,6 +64,145 @@ def __len__(self) -> int:
...
+
+    def get_zip_size(self) -> Optional[int]:
+        """Get the zip (compressed) size of this shard, if compression was used.
+
+        Returns:
+            Optional[int]: Size in bytes, or ``None`` if does not exist.
+        """
+        size = 0
+        for _, info in self.file_pairs:
+            if info is None:
```
dakinggg, line 164: i guess it is guaranteed that if one `info` is `None`, all `info` are `None`?

streaming/base/stream.py
```diff
@@ -173,8 +167,103 @@ def apply_default(self, default: Self) -> None:
...
+            if is_relative != has_proportion:
+                raise ValueError(f'Relative ("proportion") and absolute ("repeat", "samples", ' +
+                                 f'none) sub-dataset weights are incompatible with each other ' +
+                                 f'(error in stream {stream_id})')
+        return is_relative
+
+    @classmethod
+    def apply_weights(cls, streams: Sequence[Self], samples_per_stream: NDArray[np.int64],
+                      samples_per_epoch: Optional[int], seed: int) -> int:
+        """Given samples per stream, derive each stream's proportion/repeat/samples.
```
dakinggg, line 200: note in docstring that this function modifies the streams objects to set the derived values

streaming/base/stream.py
```diff
@@ -319,3 +407,27 @@ def get_shards(self, world: World) -> List[Reader]:
...
+        Args:
+            shards (List[Reader]): List of this stream's shards.
+
+        Returns:
+            List[bool]: List of whether each stream shard is present.
+        """
+        # List the cache directory (so that we hit the filesystem once).
+        local_dirname = os.path.join(self.local, self.split)
+        ls = set()
+        for dirname, _, subfiles in os.walk(local_dirname):
```
dakinggg, line 423: are subdirs explicitly disallowed somewhere? otherwise this would just silently skip them, yeah?

streaming/base/shared/__init__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2023 MosaicML Streaming authors
```
dakinggg, line 1: didn't check the shared mem refactor closely, assuming its mostly copy pasted

dakinggg (commented):
For testing, would be good to have
1. unit test for simple shard eviction (low cache limit)
2. mosaicgpt run with shard eviction to make sure it still runs, has good throughput, loss looks fine.

streaming/base/dataset.py
knighton: yep -- (list of streams in provided order) -> (list of shard entries in provided order in the index.json for each stream)

streaming/base/dataset.py
knighton: nice catch

streaming/base/stream.py
knighton: yeah we don't care about dirs, just shards

streaming/base/format/base/reader.py
knighton: yep, although not sure where or how to note this to the reader

streaming/base/format/base/reader.py
knighton: done

streaming/base/stream.py
knighton: done

streaming/base/dataset.py
knighton: done

streaming/base/shared.py
```diff
@@ -1,378 +0,0 @@
...
-        # Everybody waits until the flag is set to "go".
-        while not self.flag:
-            sleep(TICK)
-
-        # Note that we exited.
-        with self.lock:
-            self.num_exit += 1
-
-
-class CreateSharedMemory:
```
karan6181, line 169: Why we removed `CreateSharedMemory` ? This class ensures proper cleanup and hides unwanted warning messages.

streaming/base/dataset.py
```diff
@@ -659,79 +741,183 @@ def _get_work(self, world: World, epoch: int, sample_in_epoch: int) -> NDArray[n
...
+        This method is called internally by ``download_shard`` to clear space for more downloads.
+        """
+        # Get the shard with the oldest last access time.
+        shard_id = int(self._shard_access_times.numpy().argmin())
+        last_accessed = self._shard_access_times[shard_id]
+
+        # Verify the access time is valid (the shard has not been evicted). This will only pose a
+        # problem if there are no shards left to evict but you are trying to anyway.
+        if np.allclose(last_accessed, NEVER):
+            raise ValueError(f'Internal error: need to evict a shard to free up space, but no ' +
```
karan6181, line 806: Can we just throw a warning rather than an exception? What would be consequences if we throw a warning ?

streaming/base/dataset.py
```diff
@@ -659,79 +741,183 @@ def _get_work(self, world: World, epoch: int, sample_in_epoch: int) -> NDArray[n
...
-        Returns:
-            Tuple[FileLock, NDArray[np.uint8]]: Lock, and array.
-        """
-        # Get the filelock that protects shard_states shared memory array.
-        lock = FileLock(self.shard_states_filename)
-
-        shard_states = np.ndarray(self.num_shards, buffer=self._shard_states.buf, dtype=np.uint8)
+            # Unknown state.
+            lock.release()
+            raise RuntimeError(f'Unknown shard state: {state}')
```
karan6181, line 917: More descriptive error message if possible ?

streaming/base/dataset.py
```diff
@@ -747,138 +933,172 @@ def get_item(self, sample_id: int) -> Any:
...
         Returns:
-            Iterator[int]: Each sample ID, having been downloaded.
+            Iterator[int]: Each sample, having been downloaded.
         """
-        self._partition_state = _PartitionState(sample_ids)
-        Thread(target=self._download_thread, args=(self._partition_state,), daemon=True).start()
-        Thread(target=self._ready_thread, args=(self._partition_state,), daemon=True).start()
-        yield from self._partition_state
+        # Yield loop.
+        while True:
```
karan6181, line 1062: I get scared when I see infinite loop. Can we have some time to track and kill if it expires ?

streaming/base/format/base/reader.py
```diff
@@ -74,6 +75,152 @@ def __len__(self) -> int:
         """
         return self.samples
 
+    def _evict_raw(self) -> None:
+        """Remove all raw files belonging to this shard."""
+        for raw_info, _ in self.file_pairs:
+            filename = os.path.join(self.dirname, self.split, raw_info.basename)
+            if os.path.exists(filename):
+                os.remove(filename)
```
karan6181, line 83: Will there be a case where more than one thread/process is trying to delete a file? could be a edge case of race condition here where they pass line 82 !!! Thinking to either wrap `os.remove` in a `try..except` block where we ignore exception in `except` part or use `shutil.rmtree(filename, ignore_errors=True)` !

streaming/base/format/base/reader.py
```diff
@@ -74,6 +75,152 @@ def __len__(self) -> int:
...
+
+    def _evict_zip(self) -> None:
+        """Remove all zip files belonging to this shard."""
+        for _, zip_info in self.file_pairs:
+            if zip_info:
+                filename = os.path.join(self.dirname, self.split, zip_info.basename)
+                if os.path.exists(filename):
+                    os.remove(filename)
```
karan6181, line 91: same as above!

karan6181 (commented):
What happens when `cache_limit` value is less than the individual shard file ? What is the behavior ?

streaming/base/dataset.py
knighton: If we hit this problem, we can't evict a shard, which means we can't fulfill the purpose of this method

`download_shard()` needs this method to work to be able to have space to download a shard in all cases

I suppose we could return bool here instead of throwing an exception, and then throw an exception in the caller instead e.g. in download_shard, but that doesn't seem fundamentally better

streaming/base/format/base/reader.py
knighton: This is only called behind the lock `StreamingDataset._cache_filelock`

`StreamingDataset.evict_shard` -> takes lock -> `StreamingDataset._evict_shard` -> `Reader.evict` -> `Reader._evict_raw`

`StreamingDataset.download_shard` -> takes lock -> `StreamingDataset._evict_coldest_shard` -> `StreamingDataset._evict_shard` -> `Reader.evict` -> `Reader._evict_raw`

streaming/base/dataset.py
knighton: This loop is the epoch, no way to limit this

streaming/base/dataset.py
```diff
@@ -19,70 +19,101 @@
...
-                if sample_id != -1:  # If -1, we skip.
-                    yield sample_id
-                self.yield_index += 1
-                continue
-            if self.is_stopped:
-                break
+        # Signal threads to exit.
+        with self._lock:
+            if self._is_exiting:
+                raise ValueError('Called exit() on an _IterState that is already exiting.')
```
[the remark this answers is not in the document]
knighton: Fixed

knighton:
> What happens when `cache_limit` value is less than the individual shard file size? What is the behavior ?

throws exception

streaming/base/shared/prefix.py
```diff
@@ -54,48 +62,87 @@ def _unpack_locals(data: bytes) -> Set[str]:
...
+    local working directory overlap. When attaching to an existing shm fails, we have reached the
+    end of the existing shms. We will register the next one.
+    """
+    prefix_int = 0
+    for prefix_int in _each_prefix_int():
+        prefix = f'{prefix_int:06}'
+        name = f'{prefix}_locals'
+        try:
+            shm = SharedMemory(name, False)
+        except:
```
karan6181, line 95: any specific exception you want to catch such as `FileNotFoundError` or `FileExistError` ?

streaming/base/shared/prefix.py
```diff
@@ -54,48 +62,87 @@ def _unpack_locals(data: bytes) -> Set[str]:
...
+    raise errs[-1]
+
+
+def get_shm_prefix(my_locals: List[str], world: World, retry: int = 7) -> Tuple[str, SharedMemory]:
+    """Register or lookup our shared memory prefix.
+
+    Args:
+        my_locals (List[str]): Local working dir of each stream, relative to /. We need to verify
+            that there is no overlap with any other currently running StreamingDataset.
+        world (World): Information about nodes, ranks, and workers.
```
karan6181, line 130: nit: missing doc for `retry`.

karan6181:
> > What happens when `cache_limit` value is less than the individual shard file size? What is the behavior ?
> 
> throws exception

Can you please add a test for the same ? Also worth to add a test when `cache_limit` is zero or negative value!

streaming/base/shared/prefix.py
knighton: done

streaming/base/shared/prefix.py
knighton: done

streaming/base/shared/prefix.py
```diff
@@ -1,28 +1,36 @@
...
+"""Register or look up the prefix to use for all shared resources.
 
-For when using `threading` or `multiprocessing` from the python standard library won't do, because
-we are coordinating separately instantiated pytorch worker processes.
+The prefix is used by all workers using this StreamingDataset of this training job. This is used to
+prevent shared resources like shared memory and filelocks from colliding.
 """
 
 from multiprocessing import resource_tracker  # pyright: ignore
 from multiprocessing.shared_memory import SharedMemory
```
karan6181, line 11: Duplicate import of `from multiprocessing.shared_memory import SharedMemory` since we already have `from streaming.base.shared import SharedMemory`.

tests/test_eviction.py
```diff
@@ -0,0 +1,179 @@
...
+            assert set(os.listdir(local)) == set(
+                map(lambda f: f.replace('.zstd', ''), os.listdir(remote)))
+        else:
+            # Local has raw, remote has raw.
+            assert set(os.listdir(local)) == set(os.listdir(remote))
+
+    rmtree(local)
+
+
+def two(remote: str, local: str, keep_zip: bool):
```
dakinggg, line 44: should this test check that all shards are present?

tests/test_eviction.py
```diff
@@ -0,0 +1,179 @@
...
+                               local=local,
+                               keep_zip=keep_zip,
+                               cache_limit=1_000_000)
+    for _ in range(3):
+        for sample in dataset:  # pyright: ignore
+            pass
+    rmtree(local)
+
+
+def three(remote: str, local: str, keep_zip: bool):
```
dakinggg, line 58: should this test check that not all shards are present?

tests/test_eviction.py
```diff
@@ -0,0 +1,179 @@
...
+
+    assert os.listdir(local) == ['index.json']
+
+    for sample_id in range(dataset.num_samples):
+        dataset[sample_id]
+
+    assert set(os.listdir(local)) == full
+
+
+def five(remote: str, local: str, keep_zip: bool):
```
dakinggg, line 91: this test looks the same as 3? is it supposed to do something different?

streaming/base/shared/memory.py
```diff
@@ -1,43 +1,35 @@
...
-    """Create a new Shared Memory block or attach to an existing shared memory block.
+class SharedMemory:
+    """Improved quiet implementation of shared memory.
 
     Args:
-        name (Optional[str], optional): A unique shared memory block name. Defaults to ``None``.
-        create (Optional[bool], optional): Creates a new shared memory block or attaches to an
-            existing shared memory block. Defaults to ``None``.
+        name (str, optional): A unique shared memory block name. Defaults to ``None``.
+        create (bool, optional): Creates a new shared memory block or attaches to an existin
```
dakinggg, line 21:
```suggestion
        create (bool, optional): Creates a new shared memory block or attaches to an existing
```

streaming/base/shared/prefix.py
```diff
@@ -54,48 +62,101 @@ def _unpack_locals(data: bytes) -> Set[str]:
...
+    Returns:
+        Tuple[str, SharedMemory]: Shared memory prefix and object. The name is required to be very
+            short due to limitations of Python on Mac OSX.
+    """
+    # Check my locals for overlap.
+    my_locals_set = _check_self(my_locals)
+
+    # First, the local leader registers the first available shm prefix, recording its locals.
+    if world.is_local_leader:
+        prefix_int = _check_and_find_retrying(my_locals_set, retry)
```
dakinggg, line 155: If `_check_and_find_retrying` errors, this return value will be the error message. Is that intended? Should there be some error handling here?

streaming/base/shared/prefix.py
```diff
@@ -1,28 +1,36 @@
...
-# Time out to wait before raising exception
-TIMEOUT = 60
+def _each_prefix_int() -> Iterator[int]:
+    """Get each possible prefix int to check in order.
+
+    Returns:
+        Iterator[int]: Each prefix int.
+    """
+    prefix_int = 0
+    while True:
```
dakinggg, line 31: previously this loop was `range(10**6)`, should we maintain that behavior? Or is it intended for this to be infinite?

streaming/base/dataset.py
```diff
@@ -277,6 +323,14 @@ def __init__(self,
         self.stream_per_shard = np.array(stream_per_shard, np.int64)
         self.num_shards = len(self.shards)
 
+        # Check that cache limit is possible.
+        if self.cache_limit:
+            min_cache_usage = sum(map(lambda stream: stream.get_index_size(), streams))
```
dakinggg, line 328: Isn't the min actually this size + the size of one shard? I guess that is harder to figure out up front?

streaming/base/dataset.py
```diff
@@ -659,79 +763,183 @@ def _get_work(self, world: World, epoch: int, sample_in_epoch: int) -> NDArray[n
...
+            # Get the stream and shard.
             stream_id = self.stream_per_shard[shard_id]
             stream = self.streams[stream_id]
             shard = self.shards[shard_id]
+
+            # If cache_limit is enabled, we first may have to make space for the new shard.
+            if self.cache_limit:
+                # Here, we give the shard a provisional access time so it doesn't get inadvertently
+                # evicted if we need to evict shard(s).
+                self._shard_access_times[shard_id] = time()
```
dakinggg, line 902: at this point, this shard is missing, so its download time should be never, right? This line is safe, just checking my understanding because I think it is unnecessary.

dakinggg (commented):
Could we also add a unit test that tests that the _coldest_ shard is actually the one that gets evicted?

streaming/base/dataset.py
dakinggg: nvm, found the exception that happens in this case

karan6181 (commented):
Can you please also add a test (With and Without Shard eviction) using `torch.utils.data.DataLoader` by varying `num_workers` ?

streaming/base/shared/memory.py
```diff
@@ -47,28 +39,29 @@ def __init__(self,
...
                     self.created_shms.append(shm)
                 except FileExistsError:
                     sleep(TICK)
                     resource_tracker.register = self.fix_register
                     # Attaches to an existing shared memory block
-                    shm = SharedMemory(name, False, size)
+                    shm = BuiltinSharedMemory(name, False, size)
                     self.opened_shms.append(shm)
             self.shm = shm
+            self.buf = shm.buf
```
karan6181, line 64: `SharedMemory` buf is of type `MemoryView` and it is not-pickable when using `Spawn` multiprocessing. Instead of saving `buf` as a class object, create a method which returns a `buf` object. Something like this

```
@property
def buf(self):
    return self.shm.buf
```

streaming/base/dataset.py
```diff
@@ -294,40 +348,79 @@ def __init__(self,
...
+            # Also use shard presence to initialize the shard states array and last access times.
+            for shard_id, is_shard_present in enumerate(are_shards_present):
+                if is_shard_present:
+                    self._shard_states[shard_id] = _ShardState.PRESENT
+                    self._shard_access_times[shard_id] = time()
+                else:
+                    self._shard_states[shard_id] = _ShardState.MISSING
+                    self._shard_access_times[shard_id] = NEVER
+
+        self._shared_barrier(world.workers_per_node)
```
karan6181, line 414: I see a hang when running a model training on more than one GPU. However, if I replace this line of code with torch dist barrier code then my training works. Looks like  `_shared_barrier` is not sync processes.

streaming/base/shared/prefix.py
knighton: infinite

low single digits in practice

streaming/base/shared/prefix.py
knighton: fixed

streaming/base/shared/prefix.py
knighton: eh?

vchiley:
Tried running this with the llm-foundry repo
It runs without issue. Using the 125M model, I do not see any difference.
Screenshot 2023-05-09 at 12 54 25 PM
Disk utilization dropped a bit
Screenshot 2023-05-09 at 12 55 15 PM
but the same thing happened when I updt composer (without installing this branch) so I'm not sure this is from this PR
Screenshot 2023-05-09 at 12 56 23 PM

knighton:
@karan6181 @vchiley I have integrated Karan's changes, please try it again?

- StreamingDataset init: switch shared barrier -> torch dist barrier
- StreamingDataset: _cache_filelock -> torch.multiprocessingManager().Lock()
- Ditto for SharedBarrier
- Exception handling in __iter__ threads

`dataset = StreamingDataset(..., cache_limit=(bytes <  total dataset bytes))`

karan6181 (approved):

PR #255, opened 2023-05-09
Add support for Cloudflare R2 cloud storage

hlky:
## Description of changes:

Add support for Cloudflare R2 cloud storage

```
>>> import numpy as np
>>> from PIL import Image
>>> from streaming import MDSWriter
>>>
>>> # Local or remote directory in which to store the compressed output files
>>> data_dir = 'r2://test'
>>>
>>> # A dictionary mapping input fields to their data types
>>> columns = {
...     'image': 'jpeg',
...     'class': 'int'
... }
>>>
>>> # Shard compression, if any
>>> compression = 'zstd'
>>>
>>> # Save the samples as shards using MDSWriter
>>> with MDSWriter(out=data_dir, columns=columns, compression=compression) as out:
...     for i in range(10000):
...         sample = {
...             'image': Image.fromarray(np.random.randint(0, 256, (32, 32, 3), np.uint8)),
...             'class': np.random.randint(10),
...         }
...         out.write(sample)
...
>>> from torch.utils.data import DataLoader
>>> from streaming import StreamingDataset
>>>
>>> # Remote path where full dataset is persistently stored
>>> remote = 'r2://test'
>>>
>>> # Local working dir where dataset is cached during operation
>>> local = './test'
>>>
>>> # Create streaming dataset
>>> dataset = StreamingDataset(local=local, remote=remote, shuffle=True)
>>>
>>> # Let's see what is in sample #1337...
>>> sample = dataset[1337]
>>> img = sample['image']
>>> cls = sample['class']
>>>
>>> # Create PyTorch DataLoader
>>> dataloader = DataLoader(dataset)
>>> sample
{'class': 6, 'image': <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=32x32 at 0x23A3B83BFA0>}
>>>
```

image

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/storage/download.py
+++ b/streaming/base/storage/download.py
@@ -198,6 +198,37 @@ def download_from_oci(remote: str, local: str) -> None:
     os.rename(local_tmp, local)
 
 
+def download_from_r2(remote: str, local: str) -> None:
+    """Download a file from remote Cloudflare R2 to local.
+
+    Args:
+        remote (str): Remote path (S3).
+        local (str): Local path (local filesystem).
+    """
+    import boto3
+    from botocore.exceptions import ClientError
+
+    obj = urllib.parse.urlparse(remote)
+    if obj.scheme != 'r2':
+        raise ValueError(f'Expected obj.scheme to be "r2", got {obj.scheme} for remote={remote}')
+
+    # Create a new session per thread
+    session = boto3.session.Session()
+    # Create a resource client using a thread's session object
+    r2_client = session.client('s3',
+                               region_name='auto',
+                               endpoint_url=os.environ['R2_ENDPOINT_URL'],
+                               aws_access_key_id=os.environ['R2_ACCESS_KEY_ID'],
+                               aws_secret_access_key=os.environ['R2_SECRET_ACCESS_KEY'])
+    try:
+        r2_client.download_file(obj.netloc, obj.path.lstrip('/'), local)
+    except ClientError as e:
+        if e.response['Error']['Code'] in BOTOCORE_CLIENT_ERROR_CODES:
+            raise FileNotFoundError(f'Object {remote} not found.') from e
+    except Exception:
+        raise
+
+
 def download_from_local(remote: str, local: str) -> None:
     """Download a file from remote to local.
 
@@ -237,6 +268,8 @@ def download_file(remote: Optional[str], local: str, timeout: float):
         download_from_gcs(remote, local)
     elif remote.startswith('oci://'):
         download_from_oci(remote, local)
+    elif remote.startswith('r2://'):
+        download_from_r2(remote, local)
     else:
         download_from_local(remote, local)
 
--- a/streaming/base/storage/download.py
+++ b/streaming/base/storage/download.py
@@ -144,7 +144,7 @@ def download_from_gcs(remote: str, local: str) -> None:
     """Download a file from remote GCS to local.
 
     Args:
-        remote (str): Remote path (S3).
+        remote (str): Remote path (GCS).
         local (str): Local path (local filesystem).
     """
     import boto3
@@ -202,7 +202,7 @@ def download_from_r2(remote: str, local: str) -> None:
     """Download a file from remote Cloudflare R2 to local.
 
     Args:
-        remote (str): Remote path (S3).
+        remote (str): Remote path (R2).
         local (str): Local path (local filesystem).
     """
     import boto3
@@ -217,9 +217,7 @@ def download_from_r2(remote: str, local: str) -> None:
     # Create a resource client using a thread's session object
     r2_client = session.client('s3',
                                region_name='auto',
-                               endpoint_url=os.environ['R2_ENDPOINT_URL'],
-                               aws_access_key_id=os.environ['R2_ACCESS_KEY_ID'],
-                               aws_secret_access_key=os.environ['R2_SECRET_ACCESS_KEY'])
+                               endpoint_url=os.environ['S3_ENDPOINT_URL'])
     try:
         r2_client.download_file(obj.netloc, obj.path.lstrip('/'), local)
     except ClientError as e:
@@ -254,6 +252,7 @@ def download_file(remote: Optional[str], local: str, timeout: float):
     if os.path.exists(local):
         return
 
+    # fix paths for windows
     local = local.replace('\\', '/')
     if remote:
         remote = remote.replace('\\', '/')
--- a/streaming/base/storage/download.py
+++ b/streaming/base/storage/download.py
@@ -254,6 +254,10 @@ def download_file(remote: Optional[str], local: str, timeout: float):
     if os.path.exists(local):
         return
 
+    local = local.replace('\\', '/')
+    if remote:
+        remote = remote.replace('\\', '/')
+
     local_dir = os.path.dirname(local)
     os.makedirs(local_dir, exist_ok=True)
 
--- a/streaming/base/storage/upload.py
+++ b/streaming/base/storage/upload.py
@@ -15,7 +15,9 @@ import tqdm
 
 from streaming.base.storage.download import BOTOCORE_CLIENT_ERROR_CODES
 
-__all__ = ['CloudUploader', 'S3Uploader', 'GCSUploader', 'OCIUploader', 'LocalUploader']
+__all__ = [
+    'CloudUploader', 'S3Uploader', 'GCSUploader', 'OCIUploader', 'R2Uploader', 'LocalUploader'
+]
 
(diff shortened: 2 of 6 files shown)


karan6181:
Thanks @hlky for creating a PR.

hlky:
Let me know if there any changes required, thanks.

karan6181:
Overall, the PR looks great @hlky . Thank you for your contribution. I have some minor comments to it.

Cloudflare r2 is an [s3-compatible](https://developers.cloudflare.com/r2/api/s3/api/) storage system. Users credentials can be set explicitly or it can retrieved implicitly  from `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` , here is the reference [link](https://developers.cloudflare.com/r2/examples/aws/boto3/) for the same. Hence, instead of having a separate environment variables for R2 such as `R2_ACCESS_KEY_ID` and `R2_SECRET_ACCESS_KEY`. Instead, we can rely on the default AWS environment variable names such as `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. The benefit of this is that it can work in both the cases as mentioned [here](https://developers.cloudflare.com/r2/examples/aws/boto3/) where user just have to set the credentials once as part of `~/.aws/credentials` and `~/.aws/config`.

**Here is my proposal:**

- You can remove `R2_ACCESS_KEY_ID` and `R2_SECRET_ACCESS_KEY` and replace `R2_ENDPOINT_URL` with `S3_ENDPOINT_URL`.
- you can update the R2 documentation by pointing it [here](https://docs.mosaicml.com/projects/streaming/en/stable/how_to_guides/configure_cloud_storage_cred.html#others) for the credentials setup. There will be an additional step required to set the flag `S3_ENDPOINT_URL`. Also mentioned this as part of R2 documentation.
- On the streaming dataset side, streaming code will just have to do

```python
r2_client = session.client('s3', region_name='auto', endpoint_url=os.environ['S3_ENDPOINT_URL'])
```

- For the multiple object stores, for example, if user would like to keep an AWS S3 credentials as well as R2 credentials part of `~/.aws/credentials` and `~/.aws/config`, then they can do so by separating the credentials in a different `AWS_PROFILE`. For example, user just need to set all of credentials under `AWS_PROFILE` section and set it either as `os.environ['AWS_PROFILE'] = '<profile_name>'` or `export AWS_PROFILE='<profile_name>'`. You can also add this point as part of R2 documentation. The `~/.aws/credentials `  would look something like this as shown below. In this case, `os.environ['AWS_PROFILE'] = 'r2'`.

```bash
[default]
aws_access_key_id=<your_aws_access_key>
aws_secret_access_key=<your_aws_secret_key>
[r2]
aws_access_key_id=<your_cloudflare_access_key>
aws_secret_access_key=<your_cloudflare_secret_key>
```

The `~/.aws/config` would look something like this as shown below:

```bash
[default]
region=us-west-2
[r2]
region=us-west-2
```

- On the streaming side, there is no change in the API, it will still have to do the same thing since boto3 imports `AWS_PROFILE` during runtime.

```python
r2_client = session.client('s3', region_name='auto', endpoint_url=os.environ['S3_ENDPOINT_URL'])
```

Please let us know if you have any questions. Thank you!

streaming/base/storage/download.py
```diff
@@ -198,6 +198,37 @@ def download_from_oci(remote: str, local: str) -> None:
     os.rename(local_tmp, local)
 
 
+def download_from_r2(remote: str, local: str) -> None:
+    """Download a file from remote Cloudflare R2 to local.
+
+    Args:
+        remote (str): Remote path (S3).
```
karan6181, line 205: nit:`Remote path (S3).` -> `Remote path (R2).`

streaming/base/storage/download.py
```diff
@@ -223,6 +254,10 @@ def download_file(remote: Optional[str], local: str, timeout: float):
     if os.path.exists(local):
         return
 
+    local = local.replace('\\', '/')
```
karan6181, line 257: Is this for windows file system ?

streaming/base/storage/upload.py
```diff
@@ -393,6 +396,89 @@ def check_bucket_exists(self, remote: str):
...
+                obj.path.lstrip('/'),
+                Callback=lambda bytes_transferred: pbar.update(bytes_transferred),
+            )
+        self.clear_local(local=local_filename)
+
+    def check_bucket_exists(self, remote: str):
+        """Raise an exception if the bucket does not exist.
+
+        Args:
+            remote (str): S3 bucket path.
```
karan6181, line 465: nit: S3 bucket path. -> `r2 bucket path.`

tests/conftest.py
```diff
@@ -17,6 +17,7 @@
 MY_BUCKET = 'streaming-test-bucket'
 MY_PREFIX = 'train'
 GCS_URL = 'https://storage.googleapis.com'
+R2_URL = 'https://c4c9097c86eeef198b16fd088f5f99d6.r2.cloudflarestorage.com'
```
karan6181, line 20: Is this a real account id? Can we use any other alphanumeric value such as `abcdefghijkl1234567` ? since we are anyhow mocking the response!!

hlky: Random value gives SSL error and test fails, I've replaced it with just `'https://r2.cloudflarestorage.com'`

streaming/base/storage/download.py
hlky: Yes, os.path.join returns paths with `\` on Windows and remote file name e.g. `s3://test\example` doesn't work

streaming/base/storage/download.py
```diff
@@ -144,7 +144,7 @@ def download_from_gcs(remote: str, local: str) -> None:
     """Download a file from remote GCS to local.
 
     Args:
-        remote (str): Remote path (S3).
+        remote (str): Remote path (GCS).
```
karan6181, line 147: Thank You!

karan6181 (commented):
Thank you for addressing the comments. Do you have a download and upload shard results from R2 ? It would be great if you can add that in the PR description.

tests/conftest.py
karan6181: Addressed as part of https://github.com/mosaicml/streaming/pull/255/commits/3b7e2068f524d6ff2a3b9de4908ff7064a8eb5eb.

streaming/base/storage/upload.py
karan6181: Addressed as part of https://github.com/mosaicml/streaming/pull/255/commits/3b7e2068f524d6ff2a3b9de4908ff7064a8eb5eb.

streaming/base/storage/download.py
karan6181: Addressed as part of https://github.com/mosaicml/streaming/pull/255/commits/3b7e2068f524d6ff2a3b9de4908ff7064a8eb5eb.

hlky:
As suggested I have removed R2 specific environment variables, replaced `R2_ENDPOINT_URL` with `S3_ENDPOINT_URL`, and updated documentation.

There is a fix for paths on Windows in R2Uploader's `upload_file` too, this could be replicated to other upload classes. Tests on Windows are also spotty with many tests failing from `Invalid Cloud provider prefix.` and others
```
Expected: download_from_gcs('gs://streaming-test-bucket\\train\\file.txt', 'C:\\Users\\user\\AppData\\Local\\Temp\\tmp16qu_of_\\file.txt')
Actual: download_from_gcs('gs://streaming-test-bucket/train/file.txt', 'C:/Users/user/AppData/Local/Temp/tmp16qu_of_/file.txt')
```
```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid bucket name "streaming-test-bucket\train\file.txt": Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$" or be an ARN matching the regex "^arn:(aws).*:(s3|s3-object-lambda):[a-z\-0-9]*:[0-9]{12}:accesspoint[/:][a-zA-Z0-9\-.]{1,63}$|^arn:(aws).*:s3-outposts:[a-z\-0-9]+:[0-9]{12}:outpost[/:][a-zA-Z0-9\-]{1,63}[/:]accesspoint[/:][a-zA-Z0-9\-]{1,63}$"
```

I have also updated the description.

karan6181 (approved):
LGTM. Thank You!

bandish-shah:
@hlky thanks for doing this, it's great to see the OSS community get involved!  Please let us know what else we can add or improve.

hlky:
No problem, happy to help, Streaming seems to be a great alternative to WebDataset. I'll check further into Windows support as this would be useful for me.

PR #263, opened 2023-05-12
Bugfix in user_guide.md sample code

tginart:
Bugfix in quick start guide sample code

See docs at https://docs.mosaicml.com/projects/streaming/en/stable/api_reference/generated/streaming.StreamingDataset.html#streaming.StreamingDataset

I think the local & remote need to be kwargs.


knighton (approved):
Awesome, thanks

PR #272, opened 2023-05-23
Fix README slack link

growlix:
README slack link currently requires a @mosaicml.com email address


karan6181 (approved):

karan6181:
Can you please also update the link to other places? you can grep for `join.slack.com` in the codebase.

karan6181:
Slack link with a URL shortener so that if the Slack link dies on us we don't have to change all the slack links everywhere: https://github.com/mosaicml/streaming/pull/273

Issue #300, opened 2023-06-14
LocalDataset bug


tungdq212:
## LocalDataset Bug
`streaming.base.LocalDataset` need override `size` function
 ```
class LocalDataset(Array, Dataset):
    """A streaming dataset whose shards reside locally as a pytorch Dataset.

    Args:
        local (str): Local dataset directory where shards are cached by split.
        split (str, optional): Which dataset split to use, if any. Defaults to ``None``.
    """

    def __init__(self, local: str, split: Optional[str] = None):
        split = split or ''

        self.local = local
        self.split = split

        filename = os.path.join(local, split, get_index_basename())  # pyright: ignore
        obj = json.load(open(filename))
        if obj['version'] != 2:
            raise ValueError('Unsupported version')

        self.shards = []
        for info in obj['shards']:
            shard = reader_from_json(local, split, info)
            self.shards.append(shard)
        self.num_samples = sum([shard.samples for shard in self.shards])

        shard_sizes = np.array([x.samples for x in self.shards])
        self.spanner = Spanner(shard_sizes)

    def __len__(self) -> int:
        """Get the length as an IterableDataset.

        Returns:
            int: Dataset length.
        """
        return self.num_samples
    
    @property
    def size(self) -> int:
        """Get the size of the dataset in samples.

        Returns:
            int: Number of samples.
        """
        return self.length

    def get_item(self, sample_id: int) -> Dict[str, Any]:
        """Get sample by global sample ID.

        Args:
            sample_id (int): Sample ID.

        Returns:
            Dict[str, Any]: Column name with sample data.
        """
        shard_id, index_in_shard = self.spanner[sample_id]
        shard = self.shards[shard_id]
        return shard[index_in_shard]
```

**Comment by karan6181** [comment]
Hi @tungdq212, thanks for identifying the issue in `LocalDataset`. Since you have already provided the fix above, do you mind create a PR for the same ? If don't have the bandwidth, that's totally understandable too!

**Comment by tungdq212** [comment]
@karan6181 I created a PR. But why we need Array beside pytorch Dataset in LocalDataset?

**Comment by knighton** [comment]
Thanks for the PR! You have some kind of linting error. You can do something like this:

```
apt install pre-commit
pre-commit install
pre-commit run --all-files
```

The Array base class just provides fancy numpy-style indexing (https://github.com/mosaicml/streaming/blob/main/streaming/base/array.py#L78)

tungdq212:
[...]

PR #303, opened 2023-06-15
fix: :bug: LocalDataset

tungdq212:
## Description of changes:
FIx LocalDataset bug.
- Related to #300


knighton:
We are working on a patch release and we would like to add this fix as part of our next release. Hence, we have created the same fix (https://github.com/mosaicml/streaming/pull/305) as yours to move ahead with the release. Please let us know if you see any other issues with streaming or you're also welcome to create a PR with the fix directly in future. Thanks!

PR #304, opened 2023-06-16
Propagate exception between threads and processes and improved error message

karan6181:
## Description of changes:
- If the main process dies, let the thread know to stop execution.
- Improved the error message by adding a file location to the exception message.
- Added the unit test for the same.

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -1030,6 +1030,7 @@ class StreamingDataset(Array, IterableDataset):
         shard = self.shards[shard_id]
 
         sample = None
+        errors = []
         for _ in range(1 + retry):
             try:
                 # Shortcut path: just assume the shard is present. Using exceptions as control flow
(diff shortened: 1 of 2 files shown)


knighton (approved):

streaming/base/dataset.py
```diff
@@ -1208,7 +1222,7 @@ def _each_sample_id(self, it: _Iterator) -> Iterator[int]:
 
             # Background thread crashed, terminate the main process
             if self._event.is_set():
-                raise RuntimeError('Background thread failed. Check other traceback.')
```
b-chu, line 1211: Why don't we raise exception anymore for `if self._event.is_set():` and the other ones you added?

karan6181: The event is only set by the thread and if thread fails, we are propagating the error back to the main process to display as stderr. In the main process, we just need to check if the event has been set or not and come out of the loop, if we also raise an exception in the main process then there is a stack of traceback which confuses the user.

Issue #293, opened 2023-06-08
Timeout Error when local=None in StreamingDataset and training in distributed mode


vancoykendall:
While trying to train with the llm-foundry repo on 1 node with 8 gpus, I kept getting a timeout error in all the non-leader processes when using a `StreamingDataset` I uploaded to s3. 

```
File "/usr/lib/python3/dist-packages/streaming/base/dataset.py", line 350, in __init__
    stream_shards = stream.get_shards(world)
  File "/usr/lib/python3/dist-packages/streaming/base/stream.py", line 387, in get_shards
    wait_for_file_to_exist(filename, TICK, 60,
  File "/usr/lib/python3/dist-packages/streaming/base/util.py", line 52, in wait_for_file_to_exist
    raise RuntimeError(f'{err_msg}, bailing out: ' + f'{timeout:.3f} < {dt:.3f} sec.')
RuntimeError: Index file /tmp/tmpqc4652dc/train/index.json took too long to download, bailing out: 60.000 < 60.003 sec.
```

In the training config, I don't set the `data_local` field, so when the `StreamingDataset` is initialized `local=None`. I found the tmp file where the index file is supposed to be downloaded and it exists and downloads in a second so it shouldn't be timing out. I went into the code and commented out the `RuntimeError` above so it would continue running and all the processes would try loading in the index file that definitely existed. This resulted in a new error: 
```
File "/usr/lib/python3/dist-packages/streaming/base/dataset.py", line 350, in __init__
    stream_shards = stream.get_shards(world)
  File "/usr/lib/python3/dist-packages/streaming/base/stream.py", line 392, in get_shards
    obj = json.load(open(filename))
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp6uouofjr/train/index.json'
```
So it seems that since each process is setting the `local` attribute to a different tmp directory and then waits for the index file to be created in that directory, but it never will.

In `streaming/base/stream.py:100`
```python
self.local = local or mkdtemp()
```

karan6181:
Hi @vancoykendall , thanks for your detailed explanation. You are right about each process is setting the `local` attribute to a different tmp directory and this causes the above error. This is indeed a bug and we will fix it soon! Wondering, do you have a fix proposal in mind or open to contribute ? Thanks!

PR #316, opened 2023-06-27
Sync tmp directory

b-chu:
## Description of changes:
Use a temporary directory based on remote and split instead of a randomly created one so the same directory will be used in distributed training

## Issue #, if available:
#293 

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -288,6 +288,9 @@ class StreamingDataset(Array, IterableDataset):
             raise ValueError(
                 'You must provide either `streams` or `remote`/`local`, but not both.')
 
+        # Initialize torch dist ourselves, if necessary.
+        destroy_dist = maybe_init_dist()
+
         # Create a temporary directory for the default stream
         if remote is None and local is None and split is None:
             local = mkdtemp()
@@ -321,9 +324,6 @@ class StreamingDataset(Array, IterableDataset):
         # different values for these fields. We are saving the rank World here because we cannot
         # instantiate a World inside the StreamingDataset destructor.
         self._rank_world = world = World()
-
-        # Initialize torch dist ourselves, if necessary.
-        destroy_dist = maybe_init_dist()
 
         # Download each stream's index, load their shards, and map streams <-> shards.
         self.num_samples = 0
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -294,20 +294,28 @@ class StreamingDataset(Array, IterableDataset):
         # Create a temporary directory for the default stream
         if remote is None and local is None and split is None:
             local = mkdtemp()
-        # Initialize the Stream defaults.
-        default = Stream(remote=remote,
-                         local=local,
-                         split=split,
-                         download_retry=download_retry,
-                         download_timeout=download_timeout,
-                         validate_hash=validate_hash,
-                         keep_zip=keep_zip)
 
-        # Normalize to a list of Streams.
+        # Initialize the Stream defaults and normalize to a list of Streams.
         if streams:
+            default = {
+                'remote': remote,
+                'local': local,
+                'split': split,
+                'download_retry': download_retry,
+                'download_timeout': download_timeout,
+                'validate_hash': validate_hash,
+                'keep_zip': keep_zip,
+            }
             for stream in streams:
                 stream.apply_default(default)
         else:
+            default = Stream(remote=remote,
+                             local=local,
+                             split=split,
+                             download_retry=download_retry,
+                             download_timeout=download_timeout,
+                             validate_hash=validate_hash,
+                             keep_zip=keep_zip)
             streams = [default]
 
         # Validate the stream weighting scheme (relative or absolute) to catch errors before we go
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -9,7 +9,6 @@ from concurrent.futures import ThreadPoolExecutor, wait
 from concurrent.futures._base import Future
 from enum import IntEnum
 from math import ceil
-from tempfile import mkdtemp
 from threading import Event, Lock
 from time import sleep, time_ns
 from typing import Any, Dict, Iterator, Optional, Sequence, Tuple, Union
@@ -290,10 +289,6 @@ class StreamingDataset(Array, IterableDataset):
 
         # Initialize torch dist ourselves, if necessary.
         destroy_dist = maybe_init_dist()
-
-        # Create a temporary directory for the default stream
-        if remote is None and local is None and split is None:
-            local = mkdtemp()
 
         # Initialize the Stream defaults and normalize to a list of Streams.
         if streams:
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -9,6 +9,7 @@ from concurrent.futures import ThreadPoolExecutor, wait
 from concurrent.futures._base import Future
 from enum import IntEnum
 from math import ceil
+from tempfile import mkdtemp
 from threading import Event, Lock
 from time import sleep, time_ns
 from typing import Any, Dict, Iterator, Optional, Sequence, Tuple, Union
@@ -287,6 +288,9 @@ class StreamingDataset(Array, IterableDataset):
             raise ValueError(
                 'You must provide either `streams` or `remote`/`local`, but not both.')
 
+        # Create a temporary directory for the default stream
+        if remote is None and local is None and split is None:
+            local = mkdtemp()
         # Initialize the Stream defaults.
         default = Stream(remote=remote,
                          local=local,
--- a/streaming/base/stream.py
+++ b/streaming/base/stream.py
@@ -161,12 +161,10 @@ class Stream:
     def _get_temporary_directory(self) -> str:
         """Construct a path to a temporary directory based on remote and split."""
         root = tempfile.gettempdir()
-        hash = None
+        hash = ''
         if self.remote is not None:
             hash = hashlib.blake2s(self.remote.encode('utf-8'), digest_size=16).hexdigest()
-        # Removes underscore if self.split is an empty string
-        folder = '_'.join(filter(None, [hash, self.split]))
-        return os.path.join(root, folder)
+        return os.path.join(root, hash, self.split)
 
     def apply_default(self, default: Self) -> None:
         """Apply defaults, setting any unset fields.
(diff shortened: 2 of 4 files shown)


streaming/base/stream.py
```diff
@@ -145,6 +149,15 @@ def __init__(self,
             self.keep_zip = keep_zip
             self.safe_keep_zip = self.keep_zip or self.remote in {None, self.local}
 
+    def _get_temporary_directory(self) -> str:
+        root = tempfile.gettempdir()
```
karan6181, line 153: What would be the value of `root` when running on 2 GPUs? Does each rank gives the same `root` directory ?

streaming/base/stream.py
```diff
@@ -98,8 +99,11 @@ def __init__(self,
                  keep_zip: Optional[bool] = None) -> None:
         self.remote = remote
         self._local = local
-        self.local = local or mkdtemp()
         self.split = split or ''
+        if local is None:
```
karan6181, line 103: What would happen if user instantiate two or more `StreamingDataset` with same `remote` and `local=None` ? Ideal behavior is to throw a valid error message asking user to provide a `local` directory.

streaming/base/stream.py
```diff
@@ -145,6 +149,15 @@ def __init__(self,
             self.keep_zip = keep_zip
             self.safe_keep_zip = self.keep_zip or self.remote in {None, self.local}
 
+    def _get_temporary_directory(self) -> str:
+        root = tempfile.gettempdir()
+        hash = None
+        if self.remote is not None:
+            hashlib.md5(self.remote.encode('utf-8')).hexdigest()
```
karan6181, line 156: What is the max character length output of `hashlib.md5` ?

streaming/base/stream.py
b-chu: Yep they use the same `root`. I've changed the tests to use 2 GPUs and they all pass, but I'm getting a weird error at the end of the pytest saying it didn't exit correctly

```
========================= 4 passed, 1 warning in 1.17s =========================

----------End global rank 1 STDOUT----------
----------Begin global rank 1 STDERR----------

----------End global rank 1 STDERR----------
ERROR:__main__:Global rank 0 (PID 6785) exited with code -15
```

streaming/base/stream.py
b-chu: With multiple `StreamingDataset` and multiple GPUs, how do we know if a directory is being created/accessed by the different GPUs within the same `StreamingDataset` or by different `StreamingDataset`s? Does streaming have functions similar to composer's [broadcast](https://github.com/mosaicml/composer/blob/dev/composer/utils/dist.py#L248C8-L248C8)? Should we use torch.dist?

streaming/base/stream.py
b-chu: 32

streaming/base/stream.py
```diff
@@ -138,13 +140,54 @@ def __init__(self,
                 raise ValueError('`download_timeout` must be positive')
             self.download_timeout = download_timeout
 
+        self.local_directory_timeout = local_directory_timeout
+        if local_directory_timeout <= 0:
+            raise ValueError('`local_directory_timeout` must be positive')
+
         self.validate_hash = validate_hash
 
+        if local is None:
```
karan6181, line 149: There are two cases where `torch.distributed.init_process_group()` is created for a distributed call. If it is initialized before creating the `StreamingDataset` then this solution would work. However, if it is not initialized then Streaming initialize it [here](https://github.com/mosaicml/streaming/blob/main/streaming/base/dataset.py#L322) which is after Streams are created. I would recommend to move the `destroy_dist = maybe_init_dist()` line before [this](https://github.com/mosaicml/streaming/blob/main/streaming/base/dataset.py#L290).

streaming/base/stream.py
```diff
@@ -94,11 +96,11 @@ def __init__(self,
                  choose: Optional[int] = None,
                  download_retry: Optional[int] = None,
                  download_timeout: Optional[float] = None,
+                 local_directory_timeout: float = 60,
```
karan6181, line 99: I would recommend not to add another parameter. Can we use `download_timeout` instead ? I don't think `mkdirs()` can take more than 1 sec, so we don't need to worry much about timeout but it's a good habit to raise an exception which you did.

streaming/base/stream.py
```diff
@@ -138,13 +140,54 @@ def __init__(self,
...
 
+    def _get_temporary_directory(self) -> str:
+        """Construct a path to a temporary directory based on remote and split."""
+        root = tempfile.gettempdir()
+        hash = None
+        if self.remote is not None:
+            hash = hashlib.md5(self.remote.encode('utf-8')).hexdigest()
+        # Removes underscore if self.split is an empty string
+        folder = '_'.join(filter(None, [hash, self.split]))
+        return os.path.join(root, folder)
```
karan6181, line 168: How about something like this ?

```python
import streaming.base.distributed as dist
local = os.path.join(root, folder)
if dist.get_local_rank() == 0:
    if os.path.exists(local):
            raise ValueError(f'Could not create a local directory. Specify a local directory with the `local` value.')
    os.makedirs(local)
dist.barrier()
```

streaming/base/stream.py
```diff
@@ -138,13 +140,54 @@ def __init__(self,
...
 
+        self.local_directory_timeout = local_directory_timeout
+        if local_directory_timeout <= 0:
+            raise ValueError('`local_directory_timeout` must be positive')
+
         self.validate_hash = validate_hash
 
+        if local is None:
+            self.local = self._get_temporary_directory()
+            self._create_or_wait_for_local()
```
karan6181, line 151: you don't need this method. Check my below comment

streaming/base/stream.py
```diff
@@ -138,13 +140,54 @@ def __init__(self,
...
+        """Construct a path to a temporary directory based on remote and split."""
+        root = tempfile.gettempdir()
+        hash = None
+        if self.remote is not None:
+            hash = hashlib.md5(self.remote.encode('utf-8')).hexdigest()
+        # Removes underscore if self.split is an empty string
+        folder = '_'.join(filter(None, [hash, self.split]))
+        return os.path.join(root, folder)
+
+    def _create_or_wait_for_local(self) -> None:
```
karan6181, line 170: You don't need this method based on my above suggestion in another comment !

tests/test_stream.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import Optional
+
+import pytest
+from _pytest.monkeypatch import MonkeyPatch
+
+from streaming import Stream
+from streaming.base.distributed import barrier
+
+
+@pytest.mark.world_size(2)
```
karan6181, line 17: I am not 100% sure that by setting `@pytest.mark.world_size(2)` would create 2 sub-processes. Can you check once by printing the `RANK`, `WORLD_SIZE`, and `LOCAL_WORLD_SIZE` ?

tests/test_stream.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2023 MosaicML Streaming authors
```
karan6181, line 1: Can you also add a unit test for the exception that it generates ?

karan6181 (commented):
Please also try running the basic dataset iteration on more than 1 gpu (maybe 2, 4, or 8) multiple times with `local=None` to ensure it works as expected.

tests/test_stream.py
b-chu: Yep, shows the correct world size.

streaming/base/stream.py
```diff
@@ -138,13 +140,54 @@ def __init__(self,
...
         if keep_zip is not None:
             self.keep_zip = keep_zip
             self.safe_keep_zip = self.keep_zip or self.remote in {None, self.local}
 
+    def _get_temporary_directory(self) -> str:
+        """Construct a path to a temporary directory based on remote and split."""
+        root = tempfile.gettempdir()
+        hash = None
+        if self.remote is not None:
+            hash = hashlib.md5(self.remote.encode('utf-8')).hexdigest()
```
knighton, line 165: we do not want to be seen in public using md5, please use something like `sha3_256`

knighton: probably want to truncate it for length though (say 16 bytes?)

streaming/base/stream.py
```diff
@@ -138,13 +140,54 @@ def __init__(self,
...
+        # Removes underscore if self.split is an empty string
+        folder = '_'.join(filter(None, [hash, self.split]))
```
knighton, line 167: need it to be like

```
root/
    hash/
        split/
```

so just `return os.path.join(root, hash, split)` i presume

streaming/base/dataset.py
```diff
@@ -287,6 +288,9 @@ def __init__(self,
             raise ValueError(
                 'You must provide either `streams` or `remote`/`local`, but not both.')
 
+        # Create a temporary directory for the default stream
+        if remote is None and local is None and split is None:
+            local = mkdtemp()
```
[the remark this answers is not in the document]
knighton: and by `remote`/`local`, we mean our API here is that at least one of them must be set

`mkdtemp` relies on a seed not internal to streaming so is not really usable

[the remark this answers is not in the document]
b-chu: This is added because in the code path where `streams` is provided, a default stream can get created with remote/local/split all as None which raises ValueError(f'Could not create a local directory. Specify a local directory with the `local` value.') in stream.py. We need a valid temporary directory and that directory doesn't get used so it should be fine.

streaming/base/stream.py
```diff
@@ -140,11 +141,31 @@ def __init__(self,
 
         self.validate_hash = validate_hash
 
+        if local is None:
+            self.local = self._get_temporary_directory()
+            if get_local_rank() == 0:
```
knighton, line 146: why is this here? if i am not mistaken, `self.local` will get created incidentally when local rank zero downloads its index

streaming/base/dataset.py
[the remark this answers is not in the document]
knighton: lol that is gnarly

what do you think about passing `default` as a dict instead of instantiating a "default" stream?

streaming/base/stream.py
b-chu: I think we want to make sure that if multiple streaming datasets are created, they use different local directories instead of all being instantiated with None so we require that the temporary directory doesn't exist beforehand.

streaming/base/dataset.py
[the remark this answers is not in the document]
b-chu: Makes sense

karan6181 (approved):

PR #318, opened 2023-06-30
Add a regression test for StreamingDataset instantiation and iteration

b-chu:
## Description of changes:
Add a regression test for StreamingDataset instantiation and iteration

## Issue #, if available:
STR-88

## Merge Checklist:

### General

### Tests


regression/basic_flow.py
```diff
@@ -0,0 +1,165 @@
+# Copyright 2023 MosaicML Streaming authors
```
karan6181, line 1: Sorry, I did not think a proper name when I created a task. Can we rename this file to `iterate_data.py` or something better ?

regression/basic_flow.yaml
```diff
@@ -0,0 +1,32 @@
+name: streaming-regression-test-basic-flow
```
karan6181, line 1: Sorry, I did not think a proper name when I created a task. Can we rename this file to `iterate_data.yaml` or something better ?

regression/basic_flow.py
```diff
@@ -0,0 +1,165 @@
...
+from streaming import StreamingDataset
+from streaming.base.distributed import barrier
+
+_TRAIN_EPOCHS = 2
+
+
+def parse_args() -> Namespace:
+    """Parse command-line arguments.
+
+    Returns:x
```
karan6181, line 22: `x` ??

regression/dataset.py
```diff
@@ -0,0 +1,126 @@
+# Copyright 2023 MosaicML Streaming authors
```
karan6181, line 1: Can we rename this to `synthetic_dataset.py` to avoid confusion with other dataset.py file ?

regression/basic_flow.yaml
```diff
@@ -0,0 +1,32 @@
...
+  composer -n 2 regression/basic_flow.py --shuffle --shuffle_algo py1b --shuffle_seed 12 --shuffle_block_size 10000
+  composer -n 1 regression/dataset.py --delete
+
+image: mosaicml/composer:0.15.0
+scheduling:
+  resumable: true
+  priority: medium
+integrations:
+  - integration_type: git_repo
+    git_repo: b-chu/streaming
```
karan6181, line 30: missed changing it to `mosaicml/streaming` ?

regression/basic_flow.yaml
```diff
@@ -0,0 +1,32 @@
...
+    git_branch: test_basic_flow
```
karan6181, line 31: `main` ?

regression/basic_flow.yaml
```diff
@@ -0,0 +1,32 @@
+name: streaming-regression-test-basic-flow
+compute:
+  cluster: r1z1
+gpu_num: 8
+command: |-
+  cd streaming
+  composer -n 1 regression/dataset.py --create --compression gz
```
karan6181, line 7: Can you add few tests with `-n 4` and `-n 8` ?

regression/basic_flow.py
```diff
@@ -0,0 +1,165 @@
...
+        Namespace: Command-line arguments.
+    """
+    args = ArgumentParser()
+    args.add_argument('--local', default=False, action='store_true')
+    args.add_argument(
```
karan6181, line 27: We can use something like this so that you don't have to copy/paste the argparse from multiple places to here.

```python
def get_kwargs(kwargs: List[str]) -> Dict[str, str]:
    if kwargs.startswith('--'):
        kwargs = kwargs[2:]
    kwargs = kwargs.replace('-','_')
    return kwargs

args, runtime_args = args.parse_known_args()
kwargs = {get_kwargs(k):v for k,v in zip(runtime_args[::2],runtime_args[1::2])}

# one can use it as
print(int(kwargs.get('batch_size', 32)))
print(kwargs.get('cache_limit', '30gb'))
```

This way you don't have to add `args.add_argument(` for `download_retry`, `download_timeout`, `validate_hash`, etc.

regression/basic_flow.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def main(args: Namespace) -> None:
+    """Benchmark time taken to generate the epoch for a given dataset.
+
+    Args:
+        args (Namespace): Command-line arguments.
+    """
+    tmp_dir = tempfile.gettempdir()
+    tmp_upload_dir = os.path.join(tmp_dir, 'regression_upload')
```
karan6181, line 130: Why name `upload` ? Streaming dataset does not upload the shard file !! better to keep `remote` and `local` ??

regression/basic_flow.yaml
```diff
@@ -0,0 +1,32 @@
...
+  composer -n 1 regression/dataset.py --delete
```
karan6181, line 8: Do you know what would happen if anyone of the command fails ? will the job terminate or it will go onto next command to execute ?

b-chu: Job will terminate

regression/basic_flow.yaml
b-chu: Is there a good workflow for this? It needs to be set as the fork for testing and then after I see it passes, I change it, but there's no way to test that right?

karan6181: Yeah right! Just remember to change it before merging a PR by adding a new commit.

regression/iterate_data.yaml
```diff
@@ -0,0 +1,35 @@
+name: streaming-regression-test-iterate-data
+compute:
+  cluster: r1z1
```
karan6181, line 3: Please remove the `cluster` name and do something like [this](https://github.com/mosaicml/examples/blob/main/examples/end-to-end-examples/stable_diffusion/yamls/mcloud_run.yaml).

regression/synthetic_dataset.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def parse_args() -> Namespace:
+    """Parse command-line arguments.
+
+    Returns:
+        Namespace: Command-line arguments.
+    """
+    args = ArgumentParser()
+    args.add_argument('--create', default=False, action='store_true')
```
karan6181, line 35: Is it possible to add one more parameter for the type of dataset that we need to create ? say right now we have `numberAndSay`, in future, we will have synthetic CV dataset !! See if you do with nested subparsers with argparse maybe ?

regression/iterate_data.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def parse_args() -> tuple[Namespace, dict[str, str]]:
+    """Parse command-line arguments.
+
+    Returns:
+        tuple(Namespace, dict[str, str]): Command-line arguments and named arguments.
+    """
+    args = ArgumentParser()
+    args.add_argument('--local', default=False, action='store_true')
```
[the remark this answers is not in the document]
b-chu: Yes, this is the cleanest way to do boolean arguments. It doesn't work well with the kwargs workaround.

regression/synthetic_dataset.py
b-chu: I think we should keep this for now and when there's PR using CV it can charge the argument to specify a dataset type

karan6181 (approved):

PR #319, opened 2023-06-30
Add a regression test for StreamingDataset using cloud providers

b-chu:
## Description of changes:
Add a regression test for StreamingDataset using cloud providers

## Issue #, if available:
STR-90

## Merge Checklist:

### General

### Tests


regression/cloud_providers.yaml
```diff
@@ -0,0 +1,27 @@
+name: streaming-regression-test-cloud-providers
+compute:
+  cluster: r1z1
+gpu_num: 8
+command: |-
+  pip uninstall -y mosaicml-streaming
+  cd streaming
+  pip install -e '.[dev]'
+  composer -n 1 regression/synthetic_dataset.py --create --cloud gs
```
karan6181, line 9: I am thinking if we can add a command line arg for the type of dataset ? Soon, we will have to add synthetic computer vision dataset, and maybe others ?

regression/utils.py
```diff
@@ -0,0 +1,36 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Utility and helper functions for regression testing."""
+
+import os
+import tempfile
+from typing import Optional
+
+_CLOUD_REMOTE_LOCATIONS = {
+    'gs': 'gs://mosaicml-composer-tests/streaming/regression/',
```
karan6181, line 11: Is it possible to fetch the path during runtime ? I mean user would need to provide a remote_dir to run! We would like avoid putting any cloud location publicly.

regression/cloud_providers.yaml
```diff
@@ -0,0 +1,27 @@
...
+  composer -n 2 regression/iterate_data.py --cloud gs
+  composer -n 1 regression/synthetic_dataset.py --delete --cloud gs
+  composer -n 1 regression/synthetic_dataset.py --create --cloud s3
+  composer -n 4 regression/iterate_data.py --cloud s3
```
karan6181, line 13: What are we validating as part of this run ? I am thinking if we can validate the total number of shard files between remote and local directory or validate the total disk size used by the dataset between remote and local directory ?

regression/utils.py
b-chu: Hmm wouldn't it still go in the yaml then? And can be seen publicly anyways? Also it seems like we expose our cloud location elsewhere
https://github.com/mosaicml/streaming/blob/137928c595d503ae3c996ff0ea1980a91f949318/tests/test_streaming_remote.py#L24

regression/cloud_providers.yaml
b-chu: Let's do this in a separate PR where we implement CV dataset creation and add it to the yamls/command line args as necessary.

regression/iterate_data.py
```diff
@@ -35,6 +38,8 @@ def parse_args() -> tuple[Namespace, dict[str, str]]:
         tuple(Namespace, dict[str, str]): Command-line arguments and named arguments.
     """
     args = ArgumentParser()
+    args.add_argument('--cloud', type=str)
+    args.add_argument('--check_download', default=False, action='store_true')
     args.add_argument('--local', default=False, action='store_true')
```
karan6181, line 43: can we have `local` as `str` instead of `bool` and let the user provides the path of local, if it is not provided then streaming dataset would create a tmp directory

karan6181 (approved):

PR #325, opened 2023-07-16
astreaming/base/shared/prefix.py: use a complete command to clean up …

guoyejun:
…stale memory

### DIFF
--- a/streaming/base/shared/prefix.py
+++ b/streaming/base/shared/prefix.py
@@ -115,9 +115,8 @@ def _check_and_find(my_locals_set: Set[str]) -> int:
             raise ValueError(f'Reused local directory: {sorted(my_locals_set)} vs ' +
                              f'{sorted(their_locals_set)}. Provide a different one. If using ' +
                              f'a unique local directory, try deleting the local directory and ' +
-                             f'call `streaming.base.util.clean_stale_shared_memory()` only once ' +
-                             f'in your script to clean up the stale shared memory before ' +
-                             f'instantiation of `StreamingDataset`.')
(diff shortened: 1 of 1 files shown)


streaming/base/shared/prefix.py
```diff
@@ -115,9 +115,8 @@ def _check_and_find(my_locals_set: Set[str]) -> int:
             raise ValueError(f'Reused local directory: {sorted(my_locals_set)} vs ' +
                              f'{sorted(their_locals_set)}. Provide a different one. If using ' +
                              f'a unique local directory, try deleting the local directory and ' +
-                             f'call `streaming.base.util.clean_stale_shared_memory()` only once ' +
-                             f'in your script to clean up the stale shared memory before ' +
-                             f'instantiation of `StreamingDataset`.')
+                             f'call `python -c "import streaming; streaming.base.util.clean_stale_shared_memory()"` ' +
```
karan6181, line 118: Thanks for your contribution. The sentence `only once in your script to clean up the stale shared memory before instantiation of StreamingDataset` is intentional and important to avoid any confusion, such as, if user calls `streaming.base.util.clean_stale_shared_memory()` between train_dataset and val_dataset then it's a wrong thing to do, so we would like user to call this method only once. 

With regards to change `python -c "import streaming; streaming.base.util.clean_stale_shared_memory()`, sometimes user won't have a flexibility to run it as a bash command. When running the script remotely, it would be easier to call `streaming.base.util.clean_stale_shared_memory()` in the main script before instantiation of StreamingDataset.

guoyejun: sure, got it, will close the PR.

PR #330, opened 2023-07-17
Fix init local dir zip-only shard handling

knighton:
## Description of changes:

Fix bug where all the local shards were getting "normalized" (deleted) on init, because StreamingDataset was expecting present shards to have the raw form resident, but the local dir only had the zips. Then, update all the cache usage accounting logic to handle these cases. New init local dir logic:

```
        if self.compression:
            if safe_keep_zip:
                if has_raw:
                    if has_zip:
                        # Present (normalized).
                        pass
                    else:
                        # Missing: there is no natural way to arrive at this state, so drop raw.
                        has_raw = False
                        self._evict_raw()
                else:
                    if has_zip:
                        # Present: but missing raw, so need to decompress upon use.
                        pass
                    else:
                        # Missing (normalized).
                        pass
            else:
                if has_raw:
                    if has_zip:
                        # Present: zip is unnecessary, so evict it.
                        has_zip = False
                        self._evict_raw()
                    else:
                        # Present (normalized).
                        pass
                else:
                    if has_zip:
                        # Present: but missing raw, so need to decompress and evict zip upon use.
                        pass
                    else:
                        # Missing (normalized).
                        pass
```

## Issue #, if available:
STR-102

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -31,7 +31,7 @@ from streaming.base.shared import (SharedArray, SharedBarrier, SharedMemory, Sha
 from streaming.base.shuffle import get_shuffle
 from streaming.base.spanner import Spanner
 from streaming.base.stream import Stream
-from streaming.base.util import bytes_to_int
+from streaming.base.util import bytes_to_int, number_abbrev_to_int
 from streaming.base.world import World
 
 # An arbitrary time in the future, used for cold shard eviction.
@@ -218,10 +218,11 @@ class StreamingDataset(Array, IterableDataset):
         keep_zip (bool): Whether to keep or delete the compressed form when decompressing
             downloaded shards. If ``False``, keep iff remote is local or no remote. Defaults to
             ``False``.
-        epoch_size (int, optional): Number of samples to draw per epoch balanced across all
+        epoch_size (Union[int, str], optional): Number of samples to draw per epoch balanced across all
             streams. If ``None``, takes its value from the total number of underlying samples.
             Provide this field if you are weighting streams relatively to target a larger or
-            smaller epoch size. Defaults to ``None``.
+            smaller epoch size. Defaults to ``None``. Can also take in human-readable number
+            abbreviations (e.g., ``"100k"``, ``"64M"``, ``"77b"``, and so on). Defaults to ``None``.
         predownload (int, optional): Target number of samples ahead to download the shards per
             number of workers provided in a dataloader while iterating. If ``None``, its value
             gets derived using batch size and number of canonical nodes
@@ -263,7 +264,7 @@ class StreamingDataset(Array, IterableDataset):
                  download_timeout: float = 60,
                  validate_hash: Optional[str] = None,
                  keep_zip: bool = False,
-                 epoch_size: Optional[int] = None,
+                 epoch_size: Optional[Union[int, str]] = None,
                  predownload: Optional[int] = None,
                  cache_limit: Optional[Union[int, str]] = None,
                  partition_algo: str = 'orig',
@@ -369,10 +370,17 @@ class StreamingDataset(Array, IterableDataset):
         self.sample_offset_per_shard = self.samples_per_shard.cumsum() - self.samples_per_shard
         self.spanner = Spanner(self.samples_per_shard)
 
+        # Convert epoch size from string to int, if needed. Cannot be negative.
+        epoch_size_value = None
+        if epoch_size:
+            epoch_size_value = number_abbrev_to_int(epoch_size)
+            if epoch_size_value < 0:
+                raise ValueError(f'Epoch size cannot be negative. Received {epoch_size_value}.')
+
         # Now that we know the number of underlying samples of each stream, derive each stream's
         # true proportion/repeat/choose, as well as the total epoch size.
-        self.epoch_size = Stream.apply_weights(self.streams, self.samples_per_stream, epoch_size,
-                                               self.shuffle_seed)
+        self.epoch_size = Stream.apply_weights(self.streams, self.samples_per_stream,
+                                               epoch_size_value, self.shuffle_seed)
 
         # Length (__len__) is the resampled epoch size divided over the number of devices.
         self.length = ceil(self.epoch_size / world.num_ranks)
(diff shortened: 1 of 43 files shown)


streaming/base/dataset.py
```diff
@@ -43,12 +43,12 @@ class _ShardState(IntEnum):
 
     Restrictions:
     - The initial state of INVALID must be zero.
-    - State transitions: MISSING -> DOWNLOADING -> PRESENT -> MISSING.
+    - State transitions: REMOTE -> DOWNLOADING -> LOCAL -> REMOTE.
```
karan6181, line 46: I liked it.

streaming/base/dataset.py
```diff
@@ -43,12 +43,12 @@ class _ShardState(IntEnum):
...
     """
     INVALID = 0
```
karan6181, line 48: Can you please add a one line description for each what each enum means ?

streaming/base/format/base/reader.py
```diff
@@ -75,34 +75,50 @@ def __len__(self) -> int:
...
         """Bring what shard files are present to a consistent state, returning whether present.
 
         Args:
-            filenames_present (Set[str]): The listing of all files under dirname/[split/]. This is
-                listed once and then saved because there could potentially be very many shard
-                files.
-            keep_zip (bool): Whether to keep zip files when decompressing. Possible when
+            listing (Set[str]): The listing of all files under dirname/[split/]. This is listed
+                once and then saved because there could potentially be very many shard files.
+            safe_keep_zip (bool): Whether to keep zip files when decompressing. Possible when
```
karan6181, line 121: isn't `keep_zip` more intuitive ?

karan6181 (commented):
Did a quick pass, will do a second round in sometime. Overall looks good. Few comments:
- Can you update the existing test case in `test_reader.py` and maybe in `test_streaming.py` with compression parameter so that it runs with and without compression? You would need to replace `mds_dataset_dir` with below code 

```
dataset = SequenceDataset(num_samples)
columns = dict(zip(dataset.column_names, dataset.column_encodings))
local, remote = local_remote_dir
write_mds_dataset(out_root=remote, columns=columns, samples=dataset, size_limit=size_limit, compression='<your choice>')
```

Checkout [this](https://github.com/mosaicml/streaming/blob/main/tests/test_streaming.py#L26-L29) lines for reference.

streaming/base/dataset.py
knighton: I mean, if a shard is MISSING, but we didn't need it anyway, is it even missed?

If a shard can be PRESENT, should MISSING be "ABSENT"?

Or should shards be RESIDENT or NONRESIDENT?

Or HERE or THERE? Or LOCAL or REMOTE...

streaming/base/format/base/reader.py
knighton: `safe_keep_zip` is the name of a field in Stream which is derived from Stream arg `keep_zip` but protecting against blowing away the shards when your local is the remote/authoritative source

streaming/base/dataset.py
knighton: done

karan6181: Wait, I did not see a new commit !!

knighton: pushed sorry

tests/test_reader.py
```diff
@@ -293,15 +293,12 @@ def test_empty_shards_index_json_exception(local_remote_dir: Tuple[str, str]):
...
-
-    with pytest.raises(RuntimeError,
-                       match=f'.*Check if the shard file exists in your remote location.*'):
-        for _ in dataset:
-            if os.path.exists(os.path.join(local_dir, filename)):
-                os.remove(os.path.join(local_dir, filename))
-            pass
+    for _ in dataset:
+        if os.path.exists(filename):
+            os.remove(filename)
```
karan6181, line 303: Can you validate that the `filename` is present at the end of the test ?

karan6181 (commented):
Overall looks good. Can you please test certain scenarios that the feature works? We have a unit test for most of the below test case with uncompressed shard file. Can you also add it for compressed shard files?

- With compressed shard files
  - `remote=<cloud location>` and `local=/tmp/streaming`, ensure files are present in `local` directory.
  - `remote=None` and `local=/tmp/streaming`, ensure files are not getting deleted in `local` directory. (This is where the bug surfaced and you created this PR to fix it)
  - `remote=local` and `local=None`, ensure files are copied over from `remote` location to `local` location.
- uncompressed shard files
  - Same step as above.

streaming/base/format/base/reader.py
```diff
@@ -114,51 +130,78 @@ def init_local_dir(self, filenames_present: Set[str], keep_zip: bool) -> bool:
...
+                    else:
+                        # Missing (normalized).
+                        pass
         else:
-            # If we don't keep_zip, drop any zip files.
-            if is_zip_present:
-                is_zip_present = False
-                self._evict_zip()
+            if has_zip:
+                raise ValueError('Shard is invalid: compression was not used, but has a ' +
```
b-chu, line 195: Is there a test for this exception?

tests/test_reader.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/330/commits/380f9627e4fcb8b7b9a1767bc0ca479d1bd2bcab.

karan6181 (approved):

streaming/base/format/base/reader.py
karan6181: This would occur if someone manually edit the index.json file and make it corrupted.

streaming/base/stream.py
```diff
@@ -400,8 +421,7 @@ def get_shards(self, world: World) -> List[Reader]:
         filename = os.path.join(self.local, self.split, basename)  # pyright: ignore
         if world.is_local_leader:
             if self.remote:
-                tmp_filename = self._download_file(basename, basename + '.tmp')
-                os.rename(tmp_filename, filename)
+                filename = self._download_file(basename)
```
shivshandilya, line 424: the following changes cause the following error on a single node, multi-gpu setup(4 gpus):
```
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Index file at /tmp/local_path_1/train/index.json is empty or corrupted. Expecting value: line 1 column 1 (char 0)
```

On re -running the script the issue is fixed. But every time I change the `local` path, it always gives this error in the first run and fixes itself from the second run. I tried running with the earlier code and the code runs fine every time. What could be the reason? Why was this here in the first place? I mean the downloading and renaming part.

PR #349, opened 2023-07-28
Fixed sampling

snarayan21:
## Description of changes:
Added fixed sampling option to `StreamingDataset` which makes sure that the samples are the same across epochs.

Enables users to keep samples consistent across epochs by keeping the random seed the same between epochs.

## Issue #, if available:
STR-36

## Merge Checklist:

### General

### Tests

### DIFF
--- a/regression/cloud_providers.yaml
+++ b/regression/cloud_providers.yaml
@@ -1,0 +1,27 @@
+name: streaming-regression-test-cloud-providers
+compute:
+  gpus: 8 # Number of GPUs to use
+  # cluster: TODO # Name of the cluster to use for this run
+command: |-
+  pip uninstall -y mosaicml-streaming
+  cd streaming
+  pip install -e '.[dev]'
+  composer -n 1 regression/synthetic_dataset.py --create --cloud_url $GS_CLOUD_URL --size_limit 4194304 # 4mb
+  composer -n 2 regression/iterate_data.py --cloud_url $GS_CLOUD_URL --check_download
+  composer -n 1 regression/synthetic_dataset.py --delete --cloud_url $GS_CLOUD_URL
+  composer -n 1 regression/synthetic_dataset.py --create --cloud_url $S3_URL --size_limit 134217728 # 128 mb
+  composer -n 4 regression/iterate_data.py --cloud_url $S3_URL --check_download --local
+  composer -n 1 regression/synthetic_dataset.py --delete --cloud_url $S3_URL
+  composer -n 1 regression/synthetic_dataset.py --create --cloud_url $OCI_URL --size_limit 268435456 # 256 mb
+  composer -n 8 regression/iterate_data.py --cloud_url $OCI_URL --check_download
+  composer -n 1 regression/synthetic_dataset.py --delete --cloud_url $OCI_URL
+
+image: mosaicml/composer:0.15.0
+scheduling:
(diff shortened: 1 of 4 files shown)


streaming/base/dataset.py
```diff
@@ -284,12 +292,25 @@ def __init__(self,
...
         self.shuffle_block_size = shuffle_block_size
+        self.sampling_method = sampling_method.lower().strip()
 
         # Check streams vs remote/local.
         if bool(streams) == (bool(remote) or bool(local)):
             raise ValueError(
                 'You must provide either `streams` or `remote`/`local`, but not both.')
 
+        # Check sampling method is one of "balanced" or "fixed".
+        if sampling_method not in ['balanced', 'fixed']:
```
karan6181, line 303: nit: `sampling_method` -> `self.sampling_method` !!

tests/test_streaming.py
```diff
@@ -12,6 +12,103 @@
...
+        assert samples_seen == epoch_size - (epoch_size % batch_size)
+    else:
+        assert samples_seen == epoch_size
+
+
+@pytest.mark.parametrize('batch_size', [1, 4])
+@pytest.mark.parametrize('seed', [2222])
+@pytest.mark.parametrize('shuffle', [False, True])
+@pytest.mark.parametrize('drop_last', [False])
+@pytest.mark.parametrize('num_canonical_nodes', [1])
```
karan6181, line 60: Can we have a variation in `num_canonical_nodes` ? such as `4` or `8` ? Plus increase the `num_samples` if it is necessary!

tests/test_streaming.py
```diff
@@ -12,6 +12,103 @@
...
+                               batch_size=batch_size,
+                               shuffle_seed=seed,
+                               num_canonical_nodes=num_canonical_nodes,
+                               epoch_size=epoch_size,
+                               sampling_method='fixed')
+
+    # Build DataLoader
+    dataloader = StreamingDataLoader(dataset=dataset,
+                                     batch_size=batch_size,
+                                     num_workers=0,
```
karan6181, line 85: Any reason for setting this `num_workers=0` ? Would recommend setting this to real world value such as `4` or `8` !!

tests/test_streaming.py
```diff
@@ -12,6 +12,103 @@
...
+                                     drop_last=drop_last)
+
+    # iterate once over the dataloader (first epoch)
```
karan6181, line 88: You can simplify the two for loop into one something like this: I haven't ran it, maybe you need to take care of shallow copy / deep copy !!

```
# iterate once over the dataloader (first epoch)
first_samples_seen = {}
second_samples_seen = {}
is_first_epoch = True
for batch in dataloader:
    samples_seen = first_samples_seen if is_first_epoch else second_samples_seen
    for element in batch['sample']:
        int_element = int(element)
        if int_element in samples_seen:
            samples_seen[int_element] += 1
        else:
            samples_seen[int_element] = 1
    samples_seen = {}
```

snarayan21: The dataloader only has enough batches for a single epoch I thought? so you would have to loop twice right? To my understanding the dataloader restarts at the beginning of the epoch when iterated over again. I'll definitely remove the redundancy in the for loops though. The elements are the just ints and dictionary equality should work...

PR #358, opened 2023-08-02
Add iteration time test as part of regression testing

karan6181:
## Description of changes:
- Add iteration time test as part of regression testing

## Issue #, if available:
STR-89

## Merge Checklist:

### General

### Tests

### DIFF
--- a/regression/iterate_data.py
+++ b/regression/iterate_data.py
@@ -4,28 +4,14 @@
 """Create a streaming dataset from toy data with various options for regression testing."""
 
 import os
-import tempfile
+import time
 import urllib.parse
 from argparse import ArgumentParser, Namespace
 
-import utils
 from torch.utils.data import DataLoader
+from utils import get_dataloader_params, get_kwargs, get_streaming_dataset_params
 
 from streaming import StreamingDataset
-
-_TRAIN_EPOCHS = 2
-
-
-def get_kwargs(key: str) -> str:
-    """Parse key of named command-line arguments.
-
-    Returns:
-        str: Key of named arguments.
-    """
-    if key.startswith('--'):
-        key = key[2:]
-    key = key.replace('-', '_')
-    return key
 
 
 def parse_args() -> tuple[Namespace, dict[str, str]]:
@@ -35,23 +21,15 @@ def parse_args() -> tuple[Namespace, dict[str, str]]:
         tuple(Namespace, dict[str, str]): Command-line arguments and named arguments.
     """
     args = ArgumentParser()
-    args.add_argument('--cloud_url', type=str)
-    args.add_argument('--check_download', default=False, action='store_true')
-    args.add_argument('--local', default=False, action='store_true')
-    args.add_argument(
-        '--keep_zip',
-        default=False,
-        action='store_true',
-        help=('Whether to keep or delete the compressed form when decompressing'
-              ' downloaded shards for StreamingDataset.'),
-    )
-    args.add_argument(
-        '--shuffle',
-        default=False,
-        action='store_true',
-        help=('Whether to iterate over the samples in randomized order for'
-              ' StreamingDataset.'),
-    )
+    args.add_argument('--epochs',
+                      type=int,
+                      default=2,
+                      help='Number of epochs to iterate over the data')
+    args.add_argument('--validate-files', default=False, action='store_true', help='Verify files')
+    args.add_argument('--validate-iter-time',
+                      default=False,
+                      action='store_true',
+                      help='Test iter time')
 
     args, runtime_args = args.parse_known_args()
     kwargs = {get_kwargs(k): v for k, v in zip(runtime_args[::2], runtime_args[1::2])}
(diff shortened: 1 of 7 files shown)


regression/iterate_data.py
```diff
@@ -75,12 +54,14 @@ def get_file_count(cloud_url: str) -> int:
 
         bucket = Bucket(gcs_client, obj.netloc)
         files = bucket.list_blobs(prefix=obj.path.lstrip('/'))
+        return sum(1 for f in files if f.key[-1] != '/')
```
b-chu, line 57: What is this doing?

regression/iterate_data.py
```diff
@@ -35,30 +22,22 @@ def parse_args() -> tuple[Namespace, dict[str, str]]:
...
+                      action='store_true',
+                      help='Test iter time')
 
     args, runtime_args = args.parse_known_args()
     kwargs = {get_kwargs(k): v for k, v in zip(runtime_args[::2], runtime_args[1::2])}
     return args, kwargs
 
 
-def get_file_count(cloud_url: str) -> int:
+def get_file_count(cloud_url: str) -> Union[int, None]:
```
b-chu, line 40: Should we error for invalid or unsupported cloud url instead of returning None?

regression/utils.py
```diff
@@ -3,12 +3,191 @@
...
+    Returns:
+        str: Key of named arguments.
+    """
+    if kwargs.startswith('--'):
+        kwargs = kwargs[2:]
+    kwargs = kwargs.replace('-', '_')
+    return kwargs
+
+
+def get_streaming_dataset_params(kwargs: Dict[str, str]) -> Dict[str, Any]:
```
b-chu, line 30: Is there a more scalable way to do argument parsing that also takes into account future changes in dataset/dataloader/writer?

regression/utils.py
```diff
@@ -3,12 +3,191 @@
...
+        writer_params['size_limit'] = str(kwargs['size_limit'])
+    if 'progress_bar' in kwargs:
+        writer_params['progress_bar'] = kwargs['progress_bar'].lower().capitalize() == 'True'
+    if 'max_workers' in kwargs:
+        writer_params['max_workers'] = int(kwargs['max_workers'])
+    logger.debug(f'writer_params: {writer_params}')
+    return writer_params
 
 
 def get_local_remote_dir() -> str:
```
b-chu, line 135: Is this function still being used?

regression/synthetic_dataset.py
```diff
@@ -1,200 +1,306 @@
...
-            delete_s3(remote_dir)
+            delete_s3(args.out)
         elif cloud == 'oci':
-            delete_oci(remote_dir)
+            delete_oci(args.out)
 
 
 if __name__ == '__main__':
-    main(parse_args())
+    DATASET_MAP = {
```
b-chu, line 300: Should this be at the top of the file prepended with _?

regression/synthetic_dataset.py
```diff
@@ -1,200 +1,306 @@
...
-                compression=args.compression,
-                hashes=args.hashes,
-                size_limit=args.size_limit,
-        ) as out:
+        with MDSWriter(out=args.out, columns=columns, **writer_params) as out:
             for sample in dataset:
                 out.write(sample)
-    if args.delete:
-        obj = urllib.parse.urlparse(remote_dir)
+    elif args.delete:
```
b-chu, line 285: If this is elif, probably want to specify that only --create or --delete can be used and add an assert for that

regression/iterate_data.yaml
```diff
@@ -6,25 +6,41 @@ command: |-
...
-  composer -n 8 regression/iterate_data.py --shuffle --shuffle_algo py1b --shuffle_seed 12 --shuffle_block_size 10000
-  composer -n 1 regression/synthetic_dataset.py --delete
-
-image: mosaicml/composer:0.15.0
+  python regression/synthetic_dataset.py --create --name numberandsaydataset --out /tmp/streaming_dataset/ --num_samples 10_000 --size_limit 10240
+  composer -n 2 regression/iterate_data.py --remote /tmp/streaming_dataset/ --batch_size 16
+  composer -n 2 regression/iterate_data.py --remote /tmp/streaming_dataset/ --local /tmp/local_dataset/ --batch_size 32
+  composer -n 4 regression/iterate_data.py --local /tmp/streaming_dataset/
+  python regression/synthetic_dataset.py --delete --out /tmp/streaming_dataset/
+  python regression/synthetic_dataset.py --delete --out /tmp/local_dataset/
```
b-chu, line 14: This seems like a hacky way to delete the folder created by iterate_data.py. synthetic_data.py should only manage the folders it created, so maybe we have a separate cleanup file that deletes directories from iterate_data.py and synthetic_data.py?

b-chu (changes requested):
It seems like the PR includes a refactor and a new regression test. Can you write a description of what you're refactoring (longer) and what iteration time is testing (shorter)?

Also would it be better to separate out this PR into 3 PRs? One for refactor, one for the new datasets, and one for the time test?

regression/iterate_data.py
karan6181: looks like the variable `files` contains the directory as well as files and when counting, it accounts for directory as well which is incorrect. Hence the `f.key[-1] != '/'` ignores the directory and only counts the file.

regression/iterate_data.py
karan6181: I would change this to include the logic for counting the local files which is a TODO.

regression/utils.py
karan6181: I have looked into multiple options, this is the best way I could find without overriding the defaults in dataset/dataloader/writer!!

regression/utils.py
karan6181: Good catch. Will remove it if not being used.

regression/iterate_data.yaml
karan6181: Sure. Another way is to just run `rm -rf /tmp/local_dataset/` !!!

regression/synthetic_dataset.py
karan6181: Done by using `ArgumentParser.add_mutually_exclusive_group()`

regression/synthetic_dataset.py
b-chu: I mean as a private global variable

b-chu (commented):
Was there something that changed the dataloader to return dictionaries of lists in a batch instead of dictionaries of tensors?

regression/synthetic_dataset.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/358/commits/74253df1aeba4e68fb0603dac5b0cc90472c3d06

regression/synthetic_dataset.py
```diff
@@ -260,9 +267,9 @@ def main(args: Namespace, kwargs: Dict[str, str]) -> None:
     """
     dataset_params = get_dataset_params(kwargs)
     writer_params = get_writer_params(kwargs)
-    if args.name.lower() not in DATASET_MAP:
-        raise ValueError(f'Unsupported dataset {args.name}. Supported: {DATASET_MAP.keys()}')
-    dataset = DATASET_MAP[args.name.lower()](**dataset_params)
+    if args.name.lower() not in _DATASET_MAP:
+        raise ValueError(f'Unsupported dataset {args.name}. Supported: {_DATASET_MAP.keys()}')
+    dataset = getattr(sys.modules[__name__], _DATASET_MAP[args.name.lower()])(**dataset_params)
```
b-chu, line 272: Why do you need this getattr + sys.modules? Does `dataset = _DATASET_MAP[args.name.lower()](**dataset_params)` not work?

b-chu (changes requested):

regression/synthetic_dataset.py
karan6181: That's because the values are a string and not a class name. If you put values as a class name then it will say class is not defined since those are getting initialized at later point.

b-chu (approved):

PR #363, opened 2023-08-04
Add delta to mds converter

XiaohanZhangCMU:
## Description of changes:

- Introduced a utility function for converting a Spark DataFrame into an MDS format.
- This utility function is flexible and supports a callable function, allowing modifications to the original data format. The function iterates over the callable, processes the modified data, and writes it in MDS format. For instance, it can be used with a tokenizer callable function that yields tokens as output.
- Included an example Jupyter notebook demonstrating a complete workflow. It illustrates how to transform raw data into a Spark DataFrame and then convert it into the MDS format. 

## Issue #, if available
STR-111

## Merge Checklist:

### General

### Tests


streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,250 @@
...
+    else:
+        mds_index = os.path.join(mds_path[0], 'index.json')
+
+    with open(mds_index, 'w') as out:
+        json.dump(obj, out)
+
+
+def dataframeToMDS(dataframe: DataFrame,
+                   out: Union[str, Tuple[str, str]],
+                   columns: Dict[str, str],
```
Skylion007, line 90: Why can't the type of the columns be inferred by the dataframe if not specified. Having the user specify all the column types manually risks columns being lost in the conversion

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,250 @@
...
+                if udf_iterable is not None:
+                    d = udf_iterable(pdf, **udf_kwargs)  # pyright: ignore
+                else:
+                    d = pdf.to_dict('records')
+                assert is_iterable(
+                    d), f'pandas_processing_fn needs to return an iterable instead of a {type(d)}'
+
+                for row in d:
+                    mds_writer.write(row)
+        yield pd.DataFrame(pd.Series([out_file_path], name='mds_path'))
```
Skylion007, line 180: Wait, a pandas dataframe is not the same as a Spark dataframe, why are we doing a conversion here to the pandas dataframe?

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,250 @@
...
+        with MDSWriter(**mds_kwargs) as mds_writer:
+            for pdf in iterator:
+                if udf_iterable is not None:
+                    d = udf_iterable(pdf, **udf_kwargs)  # pyright: ignore
+                else:
+                    d = pdf.to_dict('records')
+                assert is_iterable(
+                    d), f'pandas_processing_fn needs to return an iterable instead of a {type(d)}'
+
+                for row in d:
```
Skylion007, line 178: You should iterate over the rows and compute each row to a dict line by line, that will be way more memory efficient.

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,250 @@
...
+        for k, v in mds_kwargs.items():
+            if v is None:
+                del mds_kwargs[k]
+
+        with MDSWriter(**mds_kwargs) as mds_writer:
+            for pdf in iterator:
+                if udf_iterable is not None:
+                    d = udf_iterable(pdf, **udf_kwargs)  # pyright: ignore
+                else:
+                    d = pdf.to_dict('records')
```
Skylion007, line 174: This will materialize the entire dataframe in memory, won't it?

Skylion007 (changes requested):
Some performance issues with large datasets

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: yes. d doubles the memory of pdf.

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: iterrows of pandas frame is not computation efficient due to the column-oriented storage pattern.

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: This is only a wrapper of file path, not creating pandas frame of the actual data.

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: In practice, inferring schema from table is very error prone. It's users' responsibility to specify correct input.

streaming/base/converters/dataframeToMDS.py
Skylion007: It's not longer a pandas frame through, it's a huge amount materialized dictionaries?

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,250 @@
...
+                assert is_iterable(
+                    d), f'pandas_processing_fn needs to return an iterable instead of a {type(d)}'
+
+                for row in d:
+                    mds_writer.write(row)
+        yield pd.DataFrame(pd.Series([out_file_path], name='mds_path'))
+
+    if not dataframe:
+        raise ValueError(f'input dataframe is none!')
```
Skylion007, line 183: This mean's it is false, not that it is None. The type should be printed out too.

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,250 @@
...
+    if 0 < sample_ratio < 1:
+        df = dataframe.sample(sample_ratio)
+
+    if partition_size > 0:
+        df = df.repartition(partition_size)
+
+    cu = CloudUploader.get(out, keep_local=keep_local)
+    if os.path.exists(cu.local) and len(os.listdir(cu.local)) != 0:
+        raise ValueError(
+            'Looks like {out} is local folder and it is not empty. MDSwriter needs an empty local folder to proceed.'
```
Skylion007, line 196: Missing fstring here?

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: One example is that a user id column can be interpreted as both integer or str.

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: Good catch~

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+def is_iterable(obj: Any) -> bool:
+    """Check if obj is iterable."""
+    return issubclass(type(obj), Iterable)
+
+
+def do_merge_index(partitions: Iterable,
+                   mds_path: Union[str, Tuple[str, str]],
+                   keep_local: bool = True,
+                   skip: bool = False):
+    """Merge index.json from partitions into one for streaming."""
```
karan6181, line 54: Missing method docs

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+    if not partitions or skip:
+        return
+
+    shards = []
+
+    for row in partitions:
+        mds_partition_index = f'{row.mds_path}/index.json'
```
karan6181, line 61: Can we have a global name for `index.json` and then use that everywhere? such as `MDS_META_FILE` or better name !!

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+        mds_partition_basename = os.path.basename(row.mds_path)
+        obj = json.load(open(mds_partition_index))
+        for i in range(len(obj['shards'])):
```
karan6181, line 64: nit `i` -> something better ?

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: 'record' is not a key but an option parameter for the to_dict function.

setup.py
```diff
@@ -64,20 +64,10 @@
...
-    'yamllint==1.32.0',
-    'moto>=4.0,<5',
-    'fastapi==0.101.0',
-    'pydantic==2.1.1',
-    'uvicorn==0.23.2',
-    'pytest-split==0.8.1',
+    'datasets>=2.4.0,<3', 'docformatter>=1.4', 'jupyter==1.0.0', 'pre-commit>=2.18.1,<4',
+    'pytest==7.4.0', 'pytest_codeblocks==0.16.1', 'pytest-cov>=4,<5', 'toml==0.10.2',
+    'yamllint==1.32.0', 'moto>=4.0,<5', 'fastapi==0.101.0', 'pydantic==2.1.1', 'uvicorn==0.23.2',
+    'pytest-split==0.8.1', 'pyspark==3.4.1'
```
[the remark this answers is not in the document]
XiaohanZhangCMU: yes, it's better to add it to install_requires.

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+                else:
+                    d = pdf.to_dict('records')
+                assert is_iterable(
+                    d), f'pandas_processing_fn needs to return an iterable instead of a {type(d)}'
+
+                for row in d:
+                    mds_writer.write(row)
+        yield pd.DataFrame(pd.Series([out_file_path], name='mds_path'))
+
+    if dataframe is None:
```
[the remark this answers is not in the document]
XiaohanZhangCMU: what to check here?

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+        raise ValueError(f'input dataframe is none!')
+
+    df = dataframe
+
+    if 0 < sample_ratio < 1:
```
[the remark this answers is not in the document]
XiaohanZhangCMU: From the conversation with Cody, we concluded it may be helpful to have the sample_ratio option here as some users may not know it's possible to sample a spark dataframe.

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+    if cu.remote is not None:
+        if merge_index == True:
+            cu.upload_file('index.json')
+        if keep_local == False:
+            shutil.rmtree(cu.local, ignore_errors=True)
+
+    return mds_path
+
+
+if __name__ == '__main__':
```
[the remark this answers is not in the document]
XiaohanZhangCMU: This part will be used for implementing spark data source: it will be called from scala shell.

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: Yes~

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: yes, good catch.

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+
+    df = dataframe
+
+    if 0 < sample_ratio < 1:
+        df = dataframe.sample(sample_ratio)
+
+    if partition_size > 0:
+        df = df.repartition(partition_size)
+
+    cu = CloudUploader.get(out, keep_local=keep_local)
```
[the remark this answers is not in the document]
XiaohanZhangCMU: We are not uploading file here. This is to parse "out" into local and remote. 
The place where file is uploaded is at the end of this function:

    if cu.remote is not None:
        if merge_index == True:
            cu.upload_file(MDS_META)

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+
+    if dataframe is None:
+        raise ValueError(f'input dataframe is none!')
+
+    df = dataframe
+
+    if 0 < sample_ratio < 1:
+        df = dataframe.sample(sample_ratio)
+
+    if partition_size > 0:
```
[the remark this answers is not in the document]
XiaohanZhangCMU: removed.

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+          keyword arguments to the udf_iterable.
+    """
+
+    def write_mds(iterator: Iterable):
+
+        context = TaskContext.get()
+        if context is not None:
+            id = context.taskAttemptId()
+        else:
+            raise ValueError('None TaskContext')
```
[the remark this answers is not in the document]
XiaohanZhangCMU: user should provide correct dataframe and specify correct schema that they want. It avoids a lot of runtime issues. MDSwriter also expects columns to be specified.

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+    if isinstance(mds_path, str):
+        mds_index = os.path.join(mds_path, 'index.json')
+    else:
+        mds_index = os.path.join(mds_path[0], 'index.json')
+
+    with open(mds_index, 'w') as out:
+        json.dump(obj, out)
+
+
+def dataframeToMDS(dataframe: DataFrame,
```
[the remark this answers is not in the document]
XiaohanZhangCMU: done.

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: user should provide correct dataframe and specify correct schema that they want. It avoids a lot of runtime issues. MDSwriter also expects columns to be specified.

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+        with MDSWriter(**mds_kwargs) as mds_writer:
+            for pdf in iterator:
+                if udf_iterable is not None:
+                    d = udf_iterable(pdf, **udf_kwargs or {})
+                else:
+                    d = pdf.to_dict('records')
+                assert is_iterable(
+                    d), f'pandas_processing_fn needs to return an iterable instead of a {type(d)}'
+
+                for row in d:
```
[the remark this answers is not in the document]
XiaohanZhangCMU: done

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,251 @@
...
+                    mds_writer.write(row)
+        yield pd.DataFrame(pd.Series([out_file_path], name='mds_path'))
+
+    if dataframe is None:
+        raise ValueError(f'input dataframe is none!')
+
+    df = dataframe
```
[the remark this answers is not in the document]
XiaohanZhangCMU: removed df.

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: done.

streaming/base/converters/dataframeToMDS.py
XiaohanZhangCMU: yes, it's possible. I enabled that. From what I seen before, inferring schema from dataframe is error prone when table has more complicated structure.

streaming/base/converters/jsonToMDS.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2023 MosaicML Streaming authors
```
knighton, line 1: 1. We should make it clear in the (1) filename, (2) method name, and (3) docstrings that this goes through spark, not directly as a user would probably assume
2. Is there some powerful reason by this is in camel case? Method/file should be named in snake case

streaming/base/converters/jsonToMDS.py
```diff
@@ -0,0 +1,46 @@
...
+from streaming.base.converters.dataframeToMDS import dataframeToMDS
+
+
+def jsonToMDS(input_path: str,
+              schema: StructType,
+              merge_index: bool = True,
+              sample_ratio: float = -1.0,
+              mds_kwargs: Dict[str, Any] = {},
+              udf_iterable: Optional[Callable] = None,
+              udf_kwargs: Dict[str, Any] = {}):
```
knighton, line 20: Default arg must not be a mutable type, please default it to `None` and normalize to empty dict inside the method.

streaming/base/converters/jsonToMDS.py
```diff
@@ -0,0 +1,46 @@
...
+        schema (StructType): User specified schema for spark.read.csv
+        merge_index (bool): Whether to merge MDS index files. Default is True.
+        sample_ratio (float): The fraction of data to randomly sample during conversion.
+            Should be in the range (0, 1). Default is -1.0 (no sampling).
+        mds_kwargs (Dict): same arguments that would be passed to mdswriter
+        udf_iterable (Callable or None): A user-defined function that returns an iterable over the dataframe. ppfn_kwargs is the k-v args for the method. Default is None.
+        udf_kwargs (Dict): Additional keyword arguments to pass to the pandas processing
+            function if provided. Default is an empty dictionary.
+
+    Returns:
```
knighton, line 34: Don't need to note Returns if it's None.

tests/base/converters/test_csvToMDS.py
```diff
@@ -0,0 +1,66 @@
...
+            'hashes': ['sha1', 'xxh64'],
+            'size_limit': 1 << 26
+        }
+        csvToMDS(temp_csv_file.name,
+                 schema=schema,
+                 merge_index=True,
+                 sample_ratio=-1.0,
+                 mds_kwargs=mds_kwargs)
+
+        assert (os.path.exists(os.path.join(out, 'index.json'))), 'No merged index found'
```
knighton, line 61: You can get `index.json` from streaming.base.format.index.get_index_basename`

streaming/base/converters/dataframeToMDS.py
knighton: Also let's fix the source, if need be

streaming/base/converters/__init__.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2023 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Base module for downloading/uploading files from/to cloud storage."""
```
knighton, line 4: cc: @karan6181

Do we have strong opinions about where we draw the lines between `scripts/` and `streaming/base/converters/`?

I have been pining for a `streaming/cli/` directory containing our collection of tools to operate on streaming datasets, which would have included downloading/uploading to remote...

knighton (commented):
I've taken a first pass, various minor nits to pick, very neat

streaming/base/converters/dataframeToMDS.py
```diff
@@ -0,0 +1,304 @@
...
+    'tokenizer': 'EleutherAI/gpt-neox-20b',
+    'eos_text': '<|endoftext|>',
+    'compression': 'zstd',
+    'split': 'train',
+    'no_wrap': False,
+    'bos_text': '',
+    'key': 'content',
+}
+
+MDS_META = 'index.json'
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Didn't know that! Fixed.

streaming/base/converters/__init__.py
XiaohanZhangCMU: Any comments? @karan6181

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+def is_iterable(obj: Any) -> bool:
+    """Check if obj is iterable."""
+    return issubclass(type(obj), Iterable)
+
+
+def infer_dataframe_schema(dataframe: DataFrame,
+                           user_defined_cols: Optional[Dict[str, Any]] = None) -> Optional[Dict]:
+    """Takes a pyspark dataframe and retrives the schema information, constructing a dictionary."""
```
karan6181, line 39: Full method comments with args is missing. Please check other methods too!

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+        MapType: None,
+        StructType: None,
+        StructField: None
+    }
+
+    def map_spark_dtype(spark_data_type: Any):
+        for k, v in dtype_mapping.items():
+            if isinstance(spark_data_type, k) and v is not None:
+                return v
+        raise ValueError(f'{spark_data_type} is not supported by MDSwrite')
```
karan6181, line 65: - nit: `MDSwrite` -> `MDSwriter`

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+
+    if user_defined_cols is not None:  # user has provided schema, we just check if mds supports the dtype
+        mds_supported_dtypes = list(dtype_mapping.values())
+        for k, v in user_defined_cols.items():
+            if k not in dataframe.columns:
+                raise ValueError(f'{k} is not a column of input dataframe')
```
karan6181, line 71: Can we also print `dataframe.columns` so that users get more context about the exception ?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+            if v not in mds_supported_dtypes:
+                raise ValueError(f'{v} is not supported by MDSwrite')
```
karan6181, line 73: nit: `MDSwrite` -> `MDSwriter`

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+        dtype = map_spark_dtype(field.dataType)
+        if dtype in _encodings:
+            schema_dict[field.name] = dtype
+        else:
+            print(_encodings)
+
+    return schema_dict
+
+
+def do_merge_index(partitions: Iterable,
```
karan6181, line 89: Wondering, can we unify a method for spark and non-spark users ? We can have a non-spark user method `merge_index()` which gets called inside `do_merge_index()` ? Your thoughts ?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+                   udf_kwargs: Optional[Dict[str, Any]] = None) -> Tuple[Any, int]:
+    """Execute a spark dataframe to MDS conversion process.
+
+    This method orchestrates the conversion of a spark dataframe into MDS format by
+    processing the input data, applying a user-defined iterable function if
+    provided, and writing the results to MDS-compatible format. The converted data is saved to mds_path.
+
+    Args:
+        dataframe (pyspark.sql.DataFrame or None): A DataFrame containing Delta Lake data.
+        merge_index (bool): Whether to merge MDS index files. Default is True.
```
karan6181, line 145: nit: `Default is True` -> "Default to ``True``"

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+        sample_ratio (float): The fraction of data to randomly sample during conversion.
+            Should be in the range (0, 1). Default is -1.0 (no sampling).
```
karan6181, line 147: `Default to` instead of `Default is` since we do this way in our codebase. And wrap the default value in double back tick mark so that the documentation renders correctly. You can refer to any file in streaming code base.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+
+        mds_kwargs (dict): arguments for MDSwrite.
```
karan6181, line 149: I think you can point the MDSWriter link here instead of copy/paste. Reason, if there is any change in the MDSWriter parameter then we don't have to update it here plus looks slick.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+            or remove it after uploading. Defaults to ``False``.
+        compression (str, optional): Optional compression or compression:level. Defaults to
+            ``None``.
+        hashes (List[str], optional): Optional list of hash algorithms to apply to shard files.
+            Defaults to ``None``.
+        size_limit (Union[int, str], optional): Optional shard size limit, after which point to start a new
+            shard. If ``None``, puts everything in one shard. Can specify bytes
+            human-readable format as well, for example ``"100kb"`` for 100 kilobyte
+            (100*1024) and so on. Defaults to ``1 << 26``
+        udf_iterable (Callable or None): A user-defined function that returns an iterable over the dataframe. ppfn_kwargs is the k-v args for the method. Default is None.
```
karan6181, line 168: I don't see `ppfn_kwargs` as the method parameter !!!

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+        udf_kwargs (Dict): Additional keyword arguments to pass to the pandas processing
+            function if provided. Default is an empty dictionary.
+
+    Return:
+        mds_path: actual local and remote path were used
```
karan6181, line 173: data type is missing

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+        fail_count: number of records failed to be converted
```
karan6181, line 174: same here

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+    Raises:
+        ValueError: If dataframe is not provided
```
karan6181, line 176: Do we need this since dataframe is a required arg ?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+                    records
+                ), f'pandas_processing_fn needs to return an iterable instead of a {type(records)}'
+
+                for sample in records:
+                    try:
+                        mds_writer.write(sample)
+                    except:
+                        logger.warning(f'failed to write sample: {sample}')
+                        count += 1
+                        raise Exception(f'failed to write sample: {sample}')
```
karan6181, line 227: Looks we are throwing the exception again here which breaks the conversion script, is it intentional ?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+        yield pd.concat(
+            [pd.Series([out_file_path], name='mds_path'),
+             pd.Series([count], name='fail_count')],
+            axis=1)
+
+    if dataframe is None or dataframe.count() == 0:
+        raise ValueError(f'input dataframe is none or empty!')
+
+    if not mds_kwargs:
+        mds_kwargs = collections.defaultdict(str)
```
karan6181, line 238: Does this also handles the datatype which is not string ? For example, `size_limit` or `max_workers` is an `int`, similarly `keep_local` and `progress_bar` is a bool, do we need some mapping which converts the str to their respective data type ?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+        infer_dataframe_schema(dataframe, mds_kwargs['columns'])
+
+    if 0 < sample_ratio < 1:
+        dataframe = dataframe.sample(sample_ratio)
+
+    out = mds_kwargs['out']
+    keep_local = False if 'keep_local' not in mds_kwargs else mds_kwargs['keep_local']
+    cu = CloudUploader.get(out, keep_local=keep_local)
+    if os.path.exists(cu.local) and len(os.listdir(cu.local)) != 0:
+        raise ValueError(
```
karan6181, line 261: Doesn't MDSWriter handle this ?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,326 @@
...
+
+    if 'columns' not in mds_kwargs:
+        logger.warning(
+            "User's discretion required: columns arg is missing from mds_kwargs. Will be auto inferred"
+        )
+        mds_kwargs['columns'] = infer_dataframe_schema(dataframe)
+    else:
+        infer_dataframe_schema(dataframe, mds_kwargs['columns'])
+
+    if 0 < sample_ratio < 1:
```
karan6181, line 254: Why do need `sample_ratio` ? Do you think it should be part of this method or let user provide the final dataframe that they wanted to convert to MDS ?

streaming/base/converters/csv_to_mds.py
```diff
@@ -0,0 +1,43 @@
...
+
+from typing import Any, Callable, Dict, Optional
+
+from pyspark.sql import SparkSession
+from pyspark.sql.types import StructType
+
+from streaming.base.converters.dataframe_to_mds import dataframeToMDS
+
+
+def csvToMDS(input_path: str,
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Agreed. I will just remove these two methods, users will figure out read csv or json to a dataframe anyway.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: Good catch. python3 would still allow following to run. but it is better to change it to {} I think. 

d = collections.defaultdict(str)
d['columns'] = {'new':'value'}

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: Cody mentioned this may be useful for non-experienced spark users. And it was handy during debugging. So I would rather keep this option for the users.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: This was for debugging. Removing "Raise Exception()" so the call will keep running and adding warning printout to spark log.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: Yes, removing.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: Not sure I understand. Can you expand it a little bit?

streaming/base/converters/dataframe_to_mds.py
karan6181: @codestar12 Do you think the `dataframeToMDS()` method should support `sample_ratio` ? Or we can expect user to provide a final dataframe ?

What is the use-case of `sample_ratio` here ?

codestar12: I wouldn't mind a sample_ratio. I think @zankner has found it can be non-trival sometimes to properly sample with spark in ways that map to the mental model of how I would expect it to in streaming

XiaohanZhangCMU: I removed sample_ratio option.

streaming/base/converters/dataframe_to_mds.py
kurosch: As the method does not have `spark.sql.DataFrame` but `pandas.DataFrame` dependencies, the question seems to be whether the code can be moved to a generic place. That would allow reusability through multiple multi-processing approaches.

karan6181: @XiaohanZhangCMU Currently, there is utility method to merge index.json for non-spark user dataset conversion usage. I am thinking if we can create a generic function for non-spark users (where they do not use dataframeToMDS() method) and the spark users.

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+def remote_local_dir() -> Any:
+    """Creates a temporary directory and then deletes it when the calling function is done."""
+
+    os.environ[
+        'GOOGLE_APPLICATION_CREDENTIALS'] = '/Users/xiaohan.zhang/.mosaic/mosaicml-research-nonprod-027345ddbdfd.json'
+
+    def _method(cloud_prefix: str = 'gs://') -> Tuple[str, str]:
+        mock_local_dir = mkdtemp()
+        mock_remote_dir = os.path.join(cloud_prefix, MY_BUCKET, MY_PREFIX)
+        return mock_local_dir, mock_remote_dir
```
karan6181, line 35: is it possible to use this https://github.com/mosaicml/streaming/blob/main/tests/common/utils.py#L19 ? On the other hand, please delete the directory once test finishes. Check [this](https://github.com/mosaicml/streaming/blob/main/tests/common/utils.py#L19) which does the same.

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+        mds_kwargs = {
+            'out': out,
+            'keep_local': keep_local,
+            'compression': 'zstd:7',
+            'hashes': ['sha1', 'xxh64'],
+            'size_limit': 1 << 26
+        }
+
+        with pytest.raises(ValueError):
+            _, _ = dataframeToMDS(dataframe.select(col('id'), col('dept'), col('properties')),
```
karan6181, line 93: What value error this would throw ? You can add a `match` argument as part of `pytest.raises` which validates the error message, for example, something like [this](https://github.com/mosaicml/streaming/blob/main/tests/test_streaming.py#L283)

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+                              merge_index=merge_index,
+                              mds_kwargs=mds_kwargs)
+
+        assert (len(os.listdir(out)) > 0), f'{out} is empty'
+        for d in os.listdir(out):
+            if os.path.isdir(os.path.join(out, d)):
+                assert (os.path.exists(os.path.join(
+                    out, d, 'index.json'))), f'No index.json found in subdirectory {d}'
+
+        if merge_index == True:
```
karan6181, line 107: just `if merge_index` !!

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+    }, {
+        'id': 'strr',
+        'dept': 'str'
+    }])
+    def test_user_defined_columns(self, dataframe: Any, user_defined_columns: Dict):
+        out = mkdtemp()
+        mds_kwargs = {
+            'out': out,
+            'columns': user_defined_columns,
+            'keep_local': True,
```
karan6181, line 153: You can skip the below args for some of the test for variation

```python
'keep_local': True,
'compression': 'zstd:7',
'hashes': ['sha1', 'xxh64'],
'size_limit': 1 << 26
```

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+            'compression': 'zstd:7',
+            'hashes': ['sha1', 'xxh64'],
+            'size_limit': 1 << 26
+        }
+        with pytest.raises(ValueError):
```
karan6181, line 158: Use the `match` arg to validate the error message

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+                                         merge_index: bool):
+        out = mkdtemp()
+        mds_kwargs = {
+            'out': out,
+            'columns': {
+                'id': 'str',
+                'dept': 'str'
+            },
+            'keep_local': keep_local,
+            'compression': 'zstd:7',
```
karan6181, line 173: sometimes, you can skip `compression` args, sometimes, `hashes` for variation

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+
+        _, _ = dataframeToMDS(dataframe, merge_index=merge_index, mds_kwargs=mds_kwargs)
+
+        assert (len(os.listdir(out)) > 0), f'{out} is empty'
+        for d in os.listdir(out):
+            if os.path.isdir(os.path.join(out, d)):
+                assert (os.path.exists(os.path.join(
+                    out, d, 'index.json'))), f'No index.json found in subdirectory {d}'
+
+        if merge_index == True:
```
karan6181, line 186: just `if merge_index` !!

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+            assert (os.path.exists(os.path.join(out, 'index.json'))), 'No merged index found'
```
karan6181, line 187: `No merged index.json found`. Check the similar error message to other places as well.

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        mds_path, fail_count = dataframeToMDS(dataframe,
+                                              merge_index=merge_index,
+                                              mds_kwargs=mds_kwargs)
+
+        assert (fail_count == 0), 'some records were not converted correctly'
+        assert out == mds_path, f'returned mds_path: {mds_path} is not the same as out: {out}'
+
+        if keep_local == False:
```
karan6181, line 233: `if not keep_local` !!

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+
+    @pytest.mark.parametrize('keep_local', [True, False])
+    @pytest.mark.parametrize('merge_index', [True, False])
+    @pytest.mark.usefixtures('remote_local_dir')
+    def test_integration_conversion_local_and_remote(self, dataframe: Any, remote_local_dir: Any,
+                                                     merge_index: bool, keep_local: bool):
+        if not LOCAL_MANUAL_TEST:
+            pytest.skip('run local only. CI cluster does not have GCS service acct set up.')
+        out = remote_local_dir(cloud_prefix='gs://')
+        # bucket_name = 'mosaicml-composer-tests'
```
karan6181, line 258: extra ?

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+        assert len(mds_path) == 2, 'returned mds is a str but should be a tuple (local, remote)'
+        assert not (os.path.exists(os.path.join(
+            mds_path[0], 'index.json'))), 'Local merged index was not removed successfully'
+        assert (len(os.listdir(mds_path[0])) > 0), f'{mds_path[0]} is not empty'
+
+    def test_simple_remote(self, dataframe: Any):
+        if not LOCAL_MANUAL_TEST:
+            pytest.skip('run local only. CI cluster does not have GCS service acct set up.')
+
+        out = 'gs://mosaicml-composer-tests/test_df2mds'
```
karan6181, line 317: Does this actually upload shard file to GCS ? I think unit test does not perform service authentication since they don't have credentials and it's expensive since the test runs for every commit.

XiaohanZhangCMU: Yes, I have set LOCAL_MANUAL_TEST to be false. All the integration tests will skipped on CI nodes.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,301 @@
...
+                else:
+                    records = pdf.to_dict('records')
+                assert is_iterable(
+                    records
+                ), f'pandas_processing_fn needs to return an iterable instead of a {type(records)}'
+
+                for sample in records:
+                    try:
+                        mds_writer.write(sample)
+                    except:
```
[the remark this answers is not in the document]
Skylion007: Yeah, I have no way of knowing hte user error in my code and it's actually majorly blocking.

setup.py
```diff
@@ -59,6 +59,7 @@
     'azure-storage-blob>=12.0.0,<13',
     'azure-storage-file-datalake>=12.11.0,<13',
     'azure-identity>=1.13.0',
+    'pyspark==3.4.1',
```
Skylion007, line 62: Does this need to be a hard import? We have so many dependencies already...

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,301 @@
...
+                return mdsDtype
+        raise ValueError(f'{spark_data_type} is not supported by MDSwriter')
+
+    if user_defined_cols is not None:  # user has provided schema, we just check if mds supports the dtype
+        mds_supported_dtypes = list(dtype_mapping.values())
+        for col_name, mdsDtype in user_defined_cols.items():
+            if col_name not in dataframe.columns:
+                raise ValueError(
+                    f'{col_name} is not a column of input dataframe: {dataframe.columns}')
+            if mdsDtype not in mds_supported_dtypes:
```
Skylion007, line 82: You should check that it actually matches the dtype schema too, not just hat its' supported

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    def map_spark_dtype(spark_data_type: Any):
+        for sparkDtype, mdsDtype in dtype_mapping.items():
+            if isinstance(spark_data_type, sparkDtype) and mdsDtype is not None:
+                return mdsDtype
+        raise ValueError(f'{spark_data_type} is not supported by MDSwriter')
+
+    if user_defined_cols is not None:  # user has provided schema, we just check if mds supports the dtype
+        mds_supported_dtypes = list(dtype_mapping.values())
```
Skylion007, line 77: should be a set

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,301 @@
...
+        for col_name, mdsDtype in user_defined_cols.items():
+            if col_name not in dataframe.columns:
+                raise ValueError(
+                    f'{col_name} is not a column of input dataframe: {dataframe.columns}')
+            if mdsDtype not in mds_supported_dtypes:
+                raise ValueError(f'{mdsDtype} is not supported by MDSwriter')
```
Skylion007, line 83:
```suggestion
                raise ValueError(f'{mdsDtype} is not supported by MDSwriter')
            actual_spark_dtype = schema[col_name].dataType
            mapped_mds_dtype = map_spark_dtype(actual_spark_dtype)
            if user_dtype != mapped_mds_dtype:
                raise ValueError(f"Mismatched types: {col_name} is {mapped_mds_dtype} in DataFrame but {user_dtype} in user_defined_cols")

File: streaming/base/converters/dataframe_to_mds.py [hunk 59]
```diff
@@ -0,0 +1,301 @@
...
+    dtype_mapping = {
+        ByteType: 'bytes',
+        ShortType: 'uint64',
+        IntegerType: 'int',
+        LongType: 'int64',
+        FloatType: 'float32',
+        DoubleType: 'float64',
+        DecimalType: 'str_decimal',
+        StringType: 'str',
+        BinaryType: None,
```
Skylion007, line 58:
```suggestion
        BinaryType: 'bytes',
```

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,301 @@
...
+        BooleanType: None,
+        TimestampType: None,
+        TimestampNTZType: None,
+        DateType: None,
+        DayTimeIntervalType: None,
+        ArrayType: None,
```
Skylion007, line 64: Don't we support ndarray natively now>

streaming/base/converters/dataframe_to_mds.py
knighton: thought it should be MDS' `uint8` type

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: It can be a one to many mapping. not possible to forsee all combinations. e.g., user can choose to encode id column with either int64, int32, or str.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: Thanks!

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: From the spark dtype doc, looks like arraytype represents list of objects of ANY types. It's not the same ndarray in streaming. Please confirm. @karan6181 @knighton

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: @knighton can you confirm that? We thought BinaryType is a byte array, so it should map to bytes.

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,322 @@
...
+from pyspark.sql.types import DecimalType, IntegerType, StringType, StructField, StructType
+
+from streaming import MDSWriter
+from streaming.base.converters import dataframeToMDS
+
+MY_PREFIX = 'train'
+MY_BUCKET = 'mosaicml-composer-tests'
+LOCAL_MANUAL_TEST = False
+os.environ[
+    'OBJC_DISABLE_INITIALIZE_FORK_SAFETY'] = 'YES'  # set to yes to all fork process in spark calls
```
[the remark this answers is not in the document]
XiaohanZhangCMU: This is only for local integration test with extra cloud uploading. If it is No, then the upload process cannot be forked on mac.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+                    records = pdf.to_dict('records')
+                assert is_iterable(
+                    records
+                ), f'pandas_processing_fn needs to return an iterable instead of a {type(records)}'
+
+                for sample in records:
+                    try:
+                        mds_writer.write(sample)
+                    except:
+                        logger.warning(f'failed to write sample: {sample}')
```
Skylion007, line 221: If you are going to do this use logger.exception or use taskContext callback.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+                        count += 1
+
+        yield pd.concat(
```
Skylion007, line 224: We should use PySpark series and concat, not Pandas here.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+    if not partitions:
+        logger.warning("No partitions exist, no mergei index")
+        return
+
+    shards = []
+
+    for row in partitions:
+        mds_partition_index = f'{row.mds_path}/{get_index_basename()}'
+        mds_partition_basename = os.path.basename(row.mds_path)
+        obj = json.load(open(mds_partition_index))
```
Skylion007, line 128: You should let pyspark handle opening and merging the JSON. This would allow it to transparently handle compressed JSONs etc. If you are going to have merge_index like this, you should abstract this out into a function that can be called as a module script.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+    """Merge index.json from partitions into one for streaming.
+
+    Args:
+        partitions (Iterable): partitions that contain pd.DataFrame
+        mds_path (Union[str, Tuple[str, str]]): (str,str)=(local,remote), str = local or remote based on parse_uri(url) result
+    Return:
+        None
+    """
+    if not partitions:
+        logger.warning("No partitions exist, no mergei index")
```
Skylion007, line 120: It is suppose to be mergei index?

streaming/base/converters/dataframe_to_mds.py
Skylion007: You also never close the file object from json.load, please use a context.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+                        logger.warning(f'failed to write sample: {sample}')
+                        count += 1
+
+        yield pd.concat(
+            [pd.Series([out_file_path], name='mds_path'),
+             pd.Series([count], name='fail_count')],
+            axis=1)
+
+    if dataframe is None or dataframe.count() == 0:
+        raise ValueError(f'input dataframe is none or empty!')
```
Skylion007, line 230: This is an empty f-string, which seems like an error.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+                return mds_dtype
+        raise ValueError(f'{spark_data_type} is not supported by MDSwriter')
+
+    if user_defined_cols is not None:  # user has provided schema, we just check if mds supports the dtype
+        mds_supported_dtypes = set(mapping_spark_to_mds.values())
+        for col_name, user_dtype in user_defined_cols.items():
+            if col_name not in dataframe.columns:
+                raise ValueError(
+                    f'{col_name} is not a column of input dataframe: {dataframe.columns}')
+            if user_dtype not in mds_supported_dtypes:
```
Skylion007, line 88: The user_dtype should never be None, but this would be valid right now because None is listed as an `mds_supported_dtype`.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+        ShortType: 'uint64',
+        IntegerType: 'int',
+        LongType: 'int64',
+        FloatType: 'float32',
+        DoubleType: 'float64',
+        DecimalType: 'str_decimal',
+        StringType: 'str',
+        BinaryType: 'bytes',
+        BooleanType: None,
+        TimestampType: None,
```
Skylion007, line 66: We should do an error check if there are any None or have the user drop any rows that have None values automatically.

streaming/base/converters/dataframe_to_mds.py
Skylion007: @kurosch @karan6181 It would be great if it could be in a module script. IE. `python -m streaming.merge_indices index_01.json ... index_99.json index_merged.json`.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+
+from streaming import MDSWriter
+from streaming.base.format.index import get_index_basename
+from streaming.base.format.mds.encodings import _encodings
+from streaming.base.storage.upload import CloudUploader
+
+logger = logging.getLogger(__name__)
+
+
+def is_iterable(obj: Any) -> bool:
```
Skylion007, line 31:
```suggestion
def is_iterable(obj: object) -> bool:
```

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+                raise ValueError(
+                    f'{col_name} is not a column of input dataframe: {dataframe.columns}')
+            if user_dtype not in mds_supported_dtypes:
+                raise ValueError(f'{user_dtype} is not supported by MDSwriter')
+
+            actual_spark_dtype = dataframe.schema[col_name].dataType
+            mapped_mds_dtype = map_spark_dtype(actual_spark_dtype)
+            if user_dtype != mapped_mds_dtype:
+                raise ValueError(f"Mismatched types: {col_name} is {mapped_mds_dtype} in DataFrame but {user_dtype} in user_defined_cols")
+        return None
```
Skylion007, line 95: Why don't just we return the user_defined_columns to make the call path simpler?

streaming/base/converters/dataframe_to_mds.py
Skylion007: count() is really expensive to compute, so we should log the count for the user so the user can instantly know if there is a bug in their process.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+        udf_kwargs = {}
+
+    if 'out' not in mds_kwargs:
+        raise ValueError(f'out and columns need to be specified in mds_kwargs')
+
+    if 'columns' not in mds_kwargs:
+        logger.warning(
+            "User's discretion required: columns arg is missing from mds_kwargs. Will be auto inferred"
+        )
+        mds_kwargs['columns'] = infer_dataframe_schema(dataframe)
```
Skylion007, line 245: log what the autoinfered dataschema is.

streaming/base/converters/dataframe_to_mds.py
Skylion007: This shoul dbe moved after infer_dataframe_schema

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+        StructField: None
+    }
+
+    def map_spark_dtype(spark_data_type: Any):
+        for spark_dtype, mds_dtype in mapping_spark_to_mds.items():
+            if isinstance(spark_data_type, spark_dtype) and mds_dtype is not None:
+                return mds_dtype
+        raise ValueError(f'{spark_data_type} is not supported by MDSwriter')
+
+    if user_defined_cols is not None:  # user has provided schema, we just check if mds supports the dtype
```
Skylion007, line 82: Can we also log which columns are being dropped?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+
+    shards = []
+
+    for row in partitions:
+        mds_partition_index = f'{row.mds_path}/{get_index_basename()}'
+        mds_partition_basename = os.path.basename(row.mds_path)
+        obj = json.load(open(mds_partition_index))
+        for i in range(len(obj['shards'])):
+            shard = obj['shards'][i]
+            for key in ['raw_data', 'zip_data']:
```
Skylion007, line 131:
```suggestion
            for key in ('raw_data', 'zip_data'):
```
Tuples are faster to construct

streaming/base/converters/dataframe_to_mds.py
Skylion007: @XiaohanZhangCMU Can't numpy array support an object dtype which equivalent? Althogh we probably do not support that anyhow.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+            if col_name not in dataframe.columns:
+                raise ValueError(
+                    f'{col_name} is not a column of input dataframe: {dataframe.columns}')
+            if user_dtype not in mds_supported_dtypes:
+                raise ValueError(f'{user_dtype} is not supported by MDSwriter')
+
+            actual_spark_dtype = dataframe.schema[col_name].dataType
+            mapped_mds_dtype = map_spark_dtype(actual_spark_dtype)
+            if user_dtype != mapped_mds_dtype:
+                raise ValueError(f"Mismatched types: {col_name} is {mapped_mds_dtype} in DataFrame but {user_dtype} in user_defined_cols")
```
Skylion007, line 94: You may want to make this a TypeError?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+        MapType: None,
+        StructType: None,
+        StructField: None
+    }
+
+    def map_spark_dtype(spark_data_type: Any):
+        for spark_dtype, mds_dtype in mapping_spark_to_mds.items():
+            if isinstance(spark_data_type, spark_dtype) and mds_dtype is not None:
+                return mds_dtype
+        raise ValueError(f'{spark_data_type} is not supported by MDSwriter')
```
Skylion007, line 80:
```suggestion
        mds_type = mapping_spark_to_mds.get(spark_dtype, None)
        if mds_type is None:
            raise ValueError(f'{spark_data_type} is not supported by MDSwriter')
        return mds_type
```
This should be way easier to read / more efficient.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+        TimestampNTZType: None,
+        DateType: None,
+        DayTimeIntervalType: None,
+        ArrayType: None,
+        MapType: None,
+        StructType: None,
+        StructField: None
+    }
+
+    def map_spark_dtype(spark_data_type: Any):
```
Skylion007, line 76:
```suggestion
    def map_spark_dtype(spark_data_type: Any) -> 'str':
```

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+    else:
+        mds_index = os.path.join(mds_path[0], get_index_basename())
+
+    with open(mds_index, 'w') as out:
+        json.dump(obj, out)
+
+
+def dataframeToMDS(dataframe: DataFrame,
+                   merge_index: bool = True,
+                   mds_kwargs: Optional[Dict[str, Any]] = None,
```
Skylion007, line 154: Instead of nesting kw_args, why not just do `**mds_kwargs` at the end, and then any args that are passed in the end are routed to MDSWriter? As long as no kwargs conflict with dataframeToMDS it should be fine.

streaming/base/converters/dataframe_to_mds.py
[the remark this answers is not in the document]
Skylion007: You should really propogate the exception unless the user wants to explicitly ignore them.

setup.py
Skylion007: @XiaohanZhangCMU Why import only if the relevant functions are called? Like do we to install a hard coded version of PySpark for all of our users?

Skylion007: I am also worried about this downgrading a users databricks' version of PySpark and breaking everything later. At the very least we shouldn't be tying it to a specific minor version, espically given how stable PySpark is.

Skylion007 (changes requested):
I have a lot of usability concerns / optimizations/ improvements I have added in comments.

streaming/base/converters/dataframe_to_mds.py
Skylion007: @knighton You have it backwards I think. ByteType should be a uint8.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: but mds_kwargs[out] is different for each partition. @Skylion007

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    def map_spark_dtype(spark_data_type: Any):
+        for spark_dtype, mds_dtype in mapping_spark_to_mds.items():
+            if isinstance(spark_data_type, spark_dtype) and mds_dtype is not None:
+                return mds_dtype
+        raise ValueError(f'{spark_data_type} is not supported by MDSwriter')
+
+    if user_defined_cols is not None:  # user has provided schema, we just check if mds supports the dtype
+        mds_supported_dtypes = set(mapping_spark_to_mds.values())
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Good catch. set cannot pop None, I eliminated None when creating the set.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,283 @@
...
+        user_defined_cols (Optional[Dict[str, Any]]): user specified schema for MDSWriter
+    Return:
+        If user_defined_cols is None, return schema_dict (dict): column name and dtypes that supported by MDSWriter
+        Else, return None
+    Exceptions:
+        Any of the datatype found to be unsupported by MDSWriter, then raise ValueError
+    """
+    mapping_spark_to_mds = {
+        ByteType: 'bytes',
+        ShortType: 'uint64',
```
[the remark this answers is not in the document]
XiaohanZhangCMU: You are right, this is a typo when copy over!

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: It is up to the user to check nan values.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: count() is not expensive. it just collects meta from the footers.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: no index merged :(

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: Yes, uint8 makes sense. thanks~

setup.py
XiaohanZhangCMU: I moved pyspark to extra_deps['databricks']. I also cut loose the spark versions to next release, it should be fine.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,302 @@
...
+        ArrayType: None,
+        MapType: None,
+        StructType: None,
+        StructField: None
+    }
+
+    def map_spark_dtype(spark_data_type: Any) -> str:
+        mds_type = mapping_spark_to_mds.get(type(spark_data_type), None)
+        if mds_type is None:
+            raise ValueError(f'{spark_data_type} is not supported by MDSwriter')
```
karan6181, line 84: nit: `MDSWriter` with `W` as caps. Check other places too!

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,302 @@
...
+    for field in schema:
+        dtype = map_spark_dtype(field.dataType)
+        if dtype in _encodings:
+            schema_dict[field.name] = dtype
+        else:
+            raise ValueError(f'{dtype} is not supported by MDSwriter')
+    return schema_dict
+
+
+def do_merge_index(partitions: Iterable, mds_path: Union[str, Tuple[str, str]]):
```
karan6181, line 118: nit: `-> None`

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,302 @@
...
+                        raise RuntimeError(f'failed to write sample: {sample}')
+                        count += 1
+
+        yield pd.concat(
+            [pd.Series([out_file_path], name='mds_path'),
+             pd.Series([count], name='fail_count')],
+            axis=1)
+
+    if dataframe is None or dataframe.count() == 0:
+        raise ValueError(f'input dataframe is none or empty!')
```
karan6181, line 240: `Input dataframe is None or Empty!`

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,302 @@
...
+
+    if not mds_kwargs:
+        mds_kwargs = {}
+
+    if not udf_kwargs:
+        udf_kwargs = {}
+
+    if 'out' not in mds_kwargs:
+        raise ValueError(f'out and columns need to be specified in mds_kwargs')
```
karan6181, line 249: "`out` and `columns` need to be specified in `mds_kwargs`"

streaming/base/converters/__init__.py
```diff
@@ -0,0 +1,8 @@
...
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Good catch. copied it from another place.

Skylion007:
You may want to use a jupytext pre-commit or some other formatiing the Jupyter notebook to make it easy to view in Github / easy to check PRs with it. Jupytext can roundtrip between notebooks and python files.

streaming/base/converters/dataframe_to_mds.py
Skylion007: Any reason not use a with context here?

karan6181 (commented):
Please clean all the output of jupyter notebook before you add it here.

docs/source/index.md
```diff
@@ -77,6 +77,7 @@ If you have any questions, please feel free to reach out to us on [Twitter](htt
    examples/facesynthetics.ipynb
    examples/synthetic_nlp.ipynb
    examples/multiprocess_dataset_conversion.ipynb
+   examples/spark_dataframe.ipynb
```
karan6181, line 80: Can we change the name to `spark_dataframe_to_MDS.ipynb` ?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,302 @@
...
+    Args:
+        dataframe (spark dataframe): dataframe to inspect schema
+        user_defined_cols (Optional[Dict[str, Any]]): user specified schema for MDSWriter
+    Return:
+        If user_defined_cols is None, return schema_dict (dict): column name and dtypes that supported by MDSWriter
+        Else, return None
+    Exceptions:
+        Any of the datatype found to be unsupported by MDSWriter, then raise ValueError
+    """
+    mapping_spark_to_mds = {
```
karan6181, line 60: Should we keep this global ? For example, after the import with name as `MAPPING_SPARK_TO_MDS` ?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,302 @@
...
+        DateType: None,
+        DayTimeIntervalType: None,
+        ArrayType: None,
+        MapType: None,
+        StructType: None,
+        StructField: None
+    }
+
+    def map_spark_dtype(spark_data_type: Any) -> str:
+        mds_type = mapping_spark_to_mds.get(type(spark_data_type), None)
```
karan6181, line 82: Missing doc string !

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+@pytest.fixture(scope='class', autouse=True)
+def remote_local_dir() -> Any:
+    """Creates a temporary directory and then deletes it when the calling function is done."""
+
+    os.environ[
+        'GOOGLE_APPLICATION_CREDENTIALS'] = '~/.mosaic/mosaicml-research-nonprod-027345ddbdfd.json'
```
karan6181, line 30: Can we have some dummy path here ? I am sure there is no credentials in this path!

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,313 @@
...
+MY_PREFIX = 'train'
+MY_BUCKET = 'mosaicml-composer-tests'
+LOCAL_MANUAL_TEST = False
+os.environ[
+    'OBJC_DISABLE_INITIALIZE_FORK_SAFETY'] = 'YES'  # set to yes to all fork process in spark calls
+
+
+@pytest.fixture(scope='class', autouse=True)
+def remote_local_dir() -> Any:
+    """Creates a temporary directory and then deletes it when the calling function is done."""
```
karan6181, line 27: Where is delete operation happening ?

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,313 @@
...
+        ])
+
+        df = spark.createDataFrame(data=data, schema=schema).repartition(3)
+        yield df
+
+    @pytest.mark.parametrize('keep_local', [True, False])
+    @pytest.mark.parametrize('merge_index', [True, False])
+    def test_end_to_end_conversion_local_nocolumns(self, dataframe: Any, keep_local: bool,
+                                                   merge_index: bool):
+        out = mkdtemp()
```
karan6181, line 83: Can we use [this](https://github.com/mosaicml/streaming/blob/main/tests/common/utils.py#L19) method ?

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,313 @@
...
+                    if shards:
+                        nsamples += shards[0]['samples']
+            assert (nsamples == sum([a['samples'] for a in mgi['shards']]))
+        else:
+            assert not (os.path.exists(os.path.join(
+                out, 'index.json'))), 'merged index is created when merge_index=False'
+
+    @pytest.mark.parametrize('use_columns', [True, False])
+    def test_end_to_end_conversion_local_decimal(self, decimal_dataframe: Any, use_columns: bool):
+        out = mkdtemp()
```
karan6181, line 121: Same here. Please try to use [this](https://github.com/mosaicml/streaming/blob/main/tests/common/utils.py#L19) method and all the other places in this file.

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,313 @@
...
+        assert (len(os.listdir(out)) > 0), f'{out} is empty'
+
+    def test_user_defined_columns(self, dataframe: Any):
+        out = mkdtemp()
+        user_defined_columns = {'idd': 'str', 'dept': 'str'}
+        mds_kwargs = {
+            'out': out,
+            'columns': user_defined_columns,
+        }
+        with pytest.raises(ValueError):  # ,match=f'*is not a column of input dataframe*'):
```
karan6181, line 141: All the `match` keyword is commented which should not be, For example, it should be like this

```
with pytest.raises(ValueError, match=f'.*is not a column of input dataframe.*'):
```

xiaohanzhan-db: Yes, I tried those, but somehow it didn't work. trying to figure why.

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -0,0 +1,313 @@
...
+
+@pytest.fixture(scope='class', autouse=True)
+def remote_local_dir() -> Any:
+    """Creates a temporary directory and then deletes it when the calling function is done."""
+
+    os.environ[
+        'GOOGLE_APPLICATION_CREDENTIALS'] = '~/.mosaic/mosaicml-research-nonprod-027345ddbdfd.json'
+
+    def _method(cloud_prefix: str = 'gs://') -> Tuple[str, str]:
+        mock_local_dir = mkdtemp()
```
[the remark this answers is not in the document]
xiaohanzhan-db: I switched local tests to use that existing method. I have modified the remote integration tests dir to enable auto deletion.

tests/base/converters/test_dataframe_to_mds.py
xiaohanzhan-db: plz see above reply.

streaming/base/converters/dataframe_to_mds.py
xiaohanzhan-db: good idea!

tests/base/converters/test_dataframe_to_mds.py
xiaohanzhan-db: Done.

tests/base/converters/test_dataframe_to_mds.py
xiaohanzhan-db: Please see reply below.

tests/base/converters/test_dataframe_to_mds.py
xiaohanzhan-db: Please see reply above.

streaming/base/converters/dataframe_to_mds.py
xiaohanzhan-db: do we need doc string for every method? I normally skip doc string for internal functions shorter than 5 lines....

docs/source/index.md
xiaohanzhan-db: done.

tests/base/converters/test_dataframe_to_mds.py
XiaohanZhangCMU: nvmd, it was a typo. resolved.

Skylion007 (approved):

streaming/base/converters/README.md
```diff
@@ -0,0 +1,5 @@
+### [Spark Dataframe Conversion](https://github.com/mosaicml/streaming/blob/main/streaming/base/converters/dataframe_to_mds.py)
```
karan6181, line 1: nit: you can remove the link and instead add it in the below line where you are mentioned `dataframeToMDS` method, just link the definition path there.

streaming/base/converters/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2023 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Utility function for converting spark dataframe to MDS dataset."""
+
+from streaming.base.converters.dataframe_to_mds import MAPPING_SPARK_TO_MDS, dataframeToMDS
+
+__all__ = ['dataframeToMDS', 'MAPPING_SPARK_TO_MDS']
```
karan6181, line 8: Does `MAPPING_SPARK_TO_MDS` getting use anywhere apart from `dataframe_to_mds.py` ? If not, then we don't need to include here !!

XiaohanZhangCMU: It is not used in the code, but I think it is useful for users to import it and inspect the mapping from a notebook?

karan6181 (approved):

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -0,0 +1,312 @@
...
+                assert is_iterable(
+                    records
+                ), f'pandas_processing_fn needs to return an iterable instead of a {type(records)}'
+
+                for sample in records:
+                    try:
+                        mds_writer.write(sample)
+                    except:
+                        raise RuntimeError(f'failed to write sample: {sample}')
+                        count += 1
```
Skylion007, line 242: I just realized that while my conversation was marked resolved, nothing actually changed here. The count +=1 is unreachable here now so the error counting is kinda useless.

PR #382, opened 2023-08-15
Fixed fake AWS credentials

karan6181:
## Description of changes:
- Fixed fake AWS credentials to fixed the secret scan complaints. Previously, it was fake too based on [this official doc](https://docs.aws.amazon.com/IAM/latest/UserGuide/security-creds.html), however, I think the secret scan the few initial character plus the length for vulnerability without actually verifying the secrets via server API calls. I have changed the few initial characters to satisfy this.


knighton (approved):
Lmao

PR #385, opened 2023-08-17
The Simulator!

snarayan21:
## Description of changes:
Adding a simulator for streaming.

## Issue #, if available:
Simulator story: https://mosaicml.atlassian.net/browse/STR-123

## Merge Checklist:

### General

### Tests


setup.py
```diff
@@ -93,6 +93,14 @@
     'sphinx-tabs==3.4.1',
 ]
 
+extra_deps['simulator'] = [
```
karan6181, line 96: Can you restrict the package version to a major version with `<` operator?

simulation/core/create_index.py
```diff
@@ -0,0 +1,83 @@
+# Copyright 2023 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Create a dataset index file from input parameters."""
+
+import os.path
+import sys
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
```
karan6181, line 9: Why do we need this?

simulation/core/create_index.py
```diff
@@ -0,0 +1,83 @@
...
+            'hashes': [],
+            'samples': samples_per_shard,
+            'size_limit': avg_raw_shard_size,
+            'version': 2,
+            'zip_data': None,
+            'compression': None
+        }
+        if avg_zip_shard_size is not None:
+            shard_data['zip_data'] = {
+                'basename': 'shard.' + str(shard_id) + '.mds.zstd',
```
karan6181, line 61: Any reason why compression extension is hard coded?

simulation/core/create_index.py
```diff
@@ -0,0 +1,83 @@
...
+    index_data['shards'] = shards_list
+
+    # Try making the directory for the stream's index.json file
+    foldername = get_random_foldername() + '_indexcreated'
+    try:
+        os.mkdir(foldername)
+    except FileExistsError:
+        print('Folder already exists, trying again...')
+        foldername = get_random_foldername()
+        os.mkdir(foldername)
```
karan6181, line 77: What happens if there is a collision here? Do we want to loop through this logic until we find a unique directory? Maybe try 100/1000 times or so?

simulation/core/create_index.py
```diff
@@ -0,0 +1,83 @@
...
+
+    with open(foldername + '/index.json', 'w') as f:
+        json.dump(index_data, f)
+        f.close()
```
karan6181, line 81: You don't need to call `f.close()` since that's handled in `with` call.

simulation/core/create_index.py
```diff
@@ -0,0 +1,83 @@
...
+    # Try making the directory for the stream's index.json file
+    foldername = get_random_foldername() + '_indexcreated'
+    try:
+        os.mkdir(foldername)
+    except FileExistsError:
+        print('Folder already exists, trying again...')
+        foldername = get_random_foldername()
+        os.mkdir(foldername)
+
+    with open(foldername + '/index.json', 'w') as f:
```
karan6181, line 79: For `index.json`, you can use the method [get_index_basename()](https://github.com/mosaicml/streaming/blob/6b638518294f7338f7bb54e9bb529bfad8e5fa50/streaming/base/format/index.py#L9)

simulation/core/create_index.py
snarayan21: Oh yeah I put this here by mistake. Within the `core` directory this isn't needed, but when we want to import functionality from the `core` directory in the `interfaces` or `testing` directories then it's necessary. Otherwise we get import errors.

simulation/core/create_index.py
snarayan21: Removed all hardcoding from shard basenames in `raw_data` and `zip_data` when creating this index file. All the fields have to be present because otherwise StreamingDataset will throw an error if the index file is not structured correctly, but no hardcoding / making fake shard names is actually needed so I've removed that now.

simulation/core/create_index.py
snarayan21: So the folders get generated and deleted right after they are processed, during `SimulationDataset` instantiation. So there are only a few of these folders that exist at a time. The `get_random_foldername()` function returns a random alphanumeric string of length 16. characters can be lowercase, uppercase, and numeric, so that's 26+26+10 = 62 choices for each character, so overall 62^16 ~= 4.7e28 possible filenames, so the collision probability is super super low. I just added this `FileExistsError` in the very slim off-chance there's a collision, but if you'd want I can increase the number of tries?

simulation/core/create_index.py
```diff
@@ -0,0 +1,77 @@
...
+        shards_list.append(shard_data)
+
+    index_data['shards'] = shards_list
+
+    # Try making the directory for the stream's index.json file
+    foldername = get_random_foldername() + '_indexcreated'
+    try:
+        os.mkdir(foldername)
+    except FileExistsError:
+        print('Folder already exists, trying again...')
```
karan6181, line 68: Should we convert all the `print()` statements to `logger.info()`?

simulation/core/node_tracker.py
```diff
@@ -0,0 +1,224 @@
...
+from sortedcollections import OrderedSet
+
+from streaming.base.spanner import Spanner
+
+
+class NodeTracker():
+    """Tracker for node information during simulation.
+
+    Args:
+        workers (int): The number of workers.
```
karan6181, line 21: Are these CPU workers? The `num_workers` parameter in Dataloader?

simulation/core/create_index.py
```diff
@@ -0,0 +1,77 @@
...
+import json
+import os
+import random
+import string
+from typing import Optional
+
+from streaming.base.format import get_index_basename
+
+
+def get_random_foldername():
```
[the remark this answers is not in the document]
snarayan21: Just tried using this and there's problems with using the `TemporaryDirectory()` because it falls out of scope once the function completes. Even returning the `TemporaryDirectory` from the function causes it to fall out of scope, meaning that it will get cleaned up before accessed in `SimulationDataset`. Because the index file creation functionality is separated from the instantiation of the StreamingDataset, I had to use actual folders instead of temp dirs.

simulation/core/create_index.py
snarayan21: changed all `print` statements to `logger.info()` or `logger.warning()` except for those in `sim_cli.py` and `sim_script.py` since it looks nicer being kept as print statements there.

simulation/core/main.py
```diff
@@ -0,0 +1,238 @@
...
+    for _ in range(physical_nodes):
+        nodes.append(
+            NodeTracker(workers, devices, predownload, device_batch_size, total_shards,
+                        cache_limit))
+
+    # Time for the global batch is just device batch size * time per sample.
+    # We assume all devices process their microbatch perfectly in parallel.
+    avg_batch_process_time = device_batch_size * time_per_sample
+
+    notification_batches = int(batches_per_epoch) / 20
```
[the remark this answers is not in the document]
snarayan21: it's just so we log batches completed so far 20 times every simulation run. So the notification batch interval is 1/20th the total number of steps in the simulation.

simulation/core/node_tracker.py
snarayan21: yeah it's the same.

simulation/core/node_tracker.py
```diff
@@ -0,0 +1,224 @@
...
+    def evict_shard(self) -> int:
+        """Evict a shard.
+
+        Returns:
+            int: The evicted shard.
+        """
+        evicted_shard = self.shards.popLRU()
+        return evicted_shard
+
+    def evict_until_satisfied(self, incoming_shard_size: int, raw_shard_sizes: NDArray[np.int64]):
```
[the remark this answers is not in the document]
snarayan21: No, this happens for the overall node. But even in the real simulation dataset, shard evictions happen through the worker that is trying to download a shard that pushes the cache usage past the cache limit. So the overall effect is essentially the same, right?

simulation/core/shuffle_quality.py
```diff
@@ -0,0 +1,160 @@
...
+        device_batch_size (int): The batch size per device.
+        samples_per_shard (int): Average number of samples per shard.
+        remove_padding (bool): Whether to remove padding samples. Defaults to ``False``.
+
+    Returns:
+        Tuple[NDArray, NDArray, NDArray]: The partition, in order, the
+            sizes of each shard, and the mapping of sample id to shard id.
+    """
+    num_samples = epoch_size
+    if num_samples > 100000000:
```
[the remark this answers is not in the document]
snarayan21: oh cool I didn't know you could do that in regular python code!

simulation/core/sim_time.py
```diff
@@ -0,0 +1,366 @@
...
+                             r'|'.join([fr'{time_unit.value}' for time_unit in TimeUnit]) + r')$',
+                             flags=re.IGNORECASE)
+
+TValue = TypeVar('TValue', int, float)
+
+
+class Time(Generic[TValue]):
+    """Time represents static durations of training time in terms of a :class:`TimeUnit` enum.
+
+    See the :doc:`Time Guide </trainer/time>` for more details on tracking time during training.
```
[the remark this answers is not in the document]
snarayan21: Oh yeah this is straight from composer. changed the docstring to be simpler and just refer to composer docs.

simulation/requirements.txt
```diff
@@ -0,0 +1,10 @@
+altair==5.1.1
```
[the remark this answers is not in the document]
snarayan21: oh whoops i'll delete this. it's the same and the requirements have already been moved to setup.py.

simulation/interfaces/widgets.py
```diff
@@ -0,0 +1,446 @@
...
+def param_inputs(component: DeltaGenerator, input_params: dict, defaults: dict = {}):
+    """Define parameter input area widget.
+
+    Args:
+        component (DeltaGenerator): Streamlit component to write to.
+        input_params (dict): Dictionary to store input parameters.
+        defaults (dict): Dictionary of default values for input params. Defaults to empty dict, {}.
+    """
+    # split the input column component into left, middle, and right sub columns.
+    colL, colM, colR = component.columns(3)
```
karan6181, line 178: nit: underscore naming convention?

simulation/interfaces/sim_ui.py
```diff
@@ -0,0 +1,397 @@
...
+    immediate_throughput_data = []
+    network_data = []
+    steps = []
+    time_to_first_batch = 0
+    futures = []
+    shuffle_quality_graphed = False
+    # Initialize min_cache_limit to be 0. Will be replaced by the simulated value.
+    min_cache_limit = 0
+    # Define partial function to pass to executor map for simulation.
+    with ProcessPoolExecutor(max_workers=8) as executor:
```
karan6181, line 81: nit: Do we need `max_workers=8`? I think it is by default set as a number of logical CPU cores in your system.

simulation/interfaces/sim_ui.py
```diff
@@ -0,0 +1,397 @@
...
+            # Use multiprocessing to get the shuffle quality results.
+            canonical_nodes = input_params['canonical_nodes']
+            physical_nodes = input_params['physical_nodes']
+            devices = input_params['devices']
+            workers = input_params['workers']
+            device_batch_size = input_params['device_batch_size']
+            shuffle_block_size = number_abbrev_to_int(input_params['shuffle_block_size'])
+            samples_per_shard = dataset.get_avg_samples_per_shard()
+            epoch_size = dataset.get_epoch_size()
+            if epoch_size > 100000000:
```
karan6181, line 95: nit: `100000000` -> `100_000_000` ?

simulation/interfaces/sim_ui.py
```diff
@@ -0,0 +1,397 @@
...
+    # Initialize min_cache_limit to be 0. Will be replaced by the simulated value.
+    min_cache_limit = 0
+    # Define partial function to pass to executor map for simulation.
+    with ProcessPoolExecutor(max_workers=8) as executor:
+        # Submit shuffle quality job to executor.
+        if shuffle_quality:
+            col1.write('Starting shuffle quality analysis...')
+            input_params = st.session_state['input_params']
+            # Use multiprocessing to get the shuffle quality results.
+            canonical_nodes = input_params['canonical_nodes']
```
karan6181, line 87: Question: Is there a case where a key is not present in the dictionary?

simulation/core/yaml_processing.py
```diff
@@ -0,0 +1,187 @@
...
+    max_duration = ensure_time(max_duration, TimeUnit.EPOCH)
+    time_unit = max_duration.unit
+    if time_unit != TimeUnit.EPOCH and time_unit != TimeUnit.BATCH:
+        raise ValueError('Simulator currently only supports max_duration in epochs or batches.')
+
+    # convert train_dataset to dict, if it isn't already
+    if isinstance(train_dataset, DictConfig):
+        train_dataset = om.to_container(train_dataset)
+
+    assert isinstance(workers, int), 'workers must be an integer.'
```
karan6181, line 110: I would recommend raising a `ValueError` instead of an `assert` statement in the source code.

simulation/README.md
```diff
@@ -0,0 +1,48 @@
...
+- Find optimal run configurations
+- Debug issues with underperforming runs
+- Better understand the impact of different configurations
+
+## 🚀 Getting Started
+Run the following to install simulator-specific dependencies, if they don't already exist:
+```
+pip install --upgrade "mosaicml-streaming[simulator]"
+```
+Then, simply run `simulator` in your command line to open the Web UI and get simulating!
```
karan6181, line 13: Should we rename this to `sd_simulator` or `sd-simulator` or something else other than `simulator`? Wondering, if there is a name collision with other application or not?

snarayan21: google search for a `simulator` command on linux and mac doesn't turn up any results...so we should probably be fine? I think the simpler the better right

simulation/interfaces/sim_ui.py
snarayan21: ooh nice. checked and confirmed you're right!

simulation/interfaces/sim_ui.py
snarayan21: nope, this dict is always initialized in `get_input_params_initial()`

simulation/core/yaml_processing.py
snarayan21: done for all lines!

simulation/README.md
snarayan21: discussed and resolved. if issues, will change later!

karan6181 (approved):
LGTM. Thank You!

PR #390, opened 2023-08-21
Plug hole in MDS type system: add arbitrary-precision decimal

knighton:
## Problem

Streaming PR [363](https://github.com/mosaicml/streaming/pull/363/) seeks to add delta to MDS conversion to Streaming. This means MDS must have an answer for each of the types in `pyspark.sql.types`.

## Non-solutions

Currently, we don't: arbitrary-precision decimals (and for that matter, ints) are not natively supported. Several workarounds are available using our current type system:
1. Use `pkl` encoding, which will correctly encode and decode, but with explosively bad data usage.
2. Use `json` encoding to dump the digits as str, then custom decode the field(s) in your subclass of StreamingDataset back to your arbitrary-precision int/float/Decimal type.
3. Use `bytes` encoding and roll your own custom encoding /and/ decoding.

Using Pickle will make me look incompetent, you say? How bad can it be?

```
for ty in [int, float, Decimal]:
    a = ty(0)
    b = pickle.dumps(a)
    c = pickle.loads(b)
    assert a == c
    print(str(type(a)), len(b))
```

```
<class 'int'> 5
<class 'float'> 21
<class 'decimal.Decimal'> 42
```

Oh. That is, if you were expecting serializing 0 to require 4 bytes, pickle is *5x* worse for floats and *10x* worse for decimals. At scale, that really adds up.

## Solution

This PR adds three new MDS types that are small, simple, and general: `str_int`, `str_float`, and `str_decimal`. We take the JSON approach of just dumping digits to str and decoding it back, providing arbitrary precision without any hassle or complexity.

Of course, ASCII digits are not the most efficient way to serialize. To support possible future work, we also include a script analyzing what the most useful configurations of fixed-size binary decimal types would be. The actual fixed-size decimal binary type(s) are not in this PR, as I would like to study the problem a bit more before committing to a certain API for them and that work is off the critical path.

```
Columns:
- int bits: bits used to hold the digits as a scalar.
- pow bits: bits used to hold the exponent (location of decimal place).
- precision: Digits of precision.
- dec range: Range of decimal places (half left, half right).

4-byte unsigned decimal:
    int bits | pow bits | precision | dec range
          23 |        8 |         6 |       512
          24 |        7 |         7 |       256
          27 |        4 |         8 |        32

4-byte signed decimal:
    int bits | pow bits | precision | dec range
          22 |        8 |         6 |       512
          24 |        6 |         7 |       128

8-byte unsigned decimal:
    int bits | pow bits | precision | dec range
          55 |        8 |        16 |       512
          57 |        6 |        17 |       128

8-byte signed decimal:
    int bits | pow bits | precision | dec range
          54 |        8 |        16 |       512
          57 |        5 |        17 |        64

16-byte unsigned decimal:
    int bits | pow bits | precision | dec range
         119 |        8 |        35 |       512
         120 |        7 |        36 |       256
         123 |        4 |        37 |        32

16-byte signed decimal:
    int bits | pow bits | precision | dec range
         118 |        8 |        35 |       512
         120 |        6 |        36 |       128
```

### DIFF
--- a/streaming/base/format/mds/encodings.py
+++ b/streaming/base/format/mds/encodings.py
@@ -402,12 +402,10 @@ class StrInt(StrEncoding):
 
     def encode(self, obj: int) -> bytes:
         self._validate(obj, int)
-        text = str(obj)
-        return text.encode('utf-8')
+        return str(obj).encode('utf-8')
 
     def decode(self, data: bytes) -> int:
-        text = data.decode('utf-8')
-        return int(text)
+        return int(data.decode('utf-8'))
 
 
 class StrFloat(Encoding):
@@ -415,12 +413,10 @@ class StrFloat(Encoding):
 
     def encode(self, obj: float) -> bytes:
         self._validate(obj, float)
-        text = str(obj)
-        return text.encode('utf-8')
+        return str(obj).encode('utf-8')
 
     def decode(self, data: bytes) -> float:
-        text = data.decode('utf-8')
-        return float(text)
+        return float(data.decode('utf-8'))
 
 
 class StrDecimal(Encoding):
@@ -428,12 +424,10 @@ class StrDecimal(Encoding):
 
     def encode(self, obj: Decimal) -> bytes:
         self._validate(obj, Decimal)
-        text = str(obj)
-        return text.encode('utf-8')
+        return str(obj).encode('utf-8')
 
     def decode(self, data: bytes) -> Decimal:
-        text = data.decode('utf-8')
-        return Decimal(text)
+        return Decimal(data.decode('utf-8'))
 
 
 class PIL(Encoding):
--- a/streaming/base/format/mds/encodings.py
+++ b/streaming/base/format/mds/encodings.py
@@ -6,6 +6,7 @@
 import json
 import pickle
 from abc import ABC, abstractmethod
+from decimal import Decimal
 from io import BytesIO
 from typing import Any, Optional, Set, Tuple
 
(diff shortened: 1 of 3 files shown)


XiaohanZhangCMU:
Fixed a test failure, otherwise all look legitimate. Please add any comments here. Idea is to merge this PR before 363. 
@knighton @karan6181

streaming/base/format/mds/encodings.py
```diff
@@ -385,6 +386,56 @@ def __init__(self):
...
+    """
+
+    pass
+
+
+class StrInt(StrEncoding):
+    """Store int as variable-length digits str."""
+
+    def encode(self, obj: int) -> bytes:
+        self._validate(obj, int)
```
karan6181, line 404: Missing method comments. Check all the other places too!

tests/test_encodings.py
```diff
@@ -422,8 +422,8 @@ def test_mds_float64(self, decoded: float, encoded: bytes):
 
     def test_get_mds_encodings(self):
```
karan6181, line 423: Can you please add the test case for each additional data types ? You will find an example in the same file.

karan6181:
Geat PR Description

streaming/base/format/mds/encodings.py
XiaohanZhangCMU: @karan6181  Looks like for simple methods with less than 3 lines in this file, encode and decode tend to not have doc strings. Shall I add doc strings to all of them?

tests/test_encodings.py
```diff
@@ -420,10 +421,43 @@ def test_mds_float64(self, decoded: float, encoded: bytes):
         assert isinstance(dec, np.floating)
         assert dec == decoded
 
+    @pytest.mark.parametrize(('decoded', 'encoded'), [(42, b'42')])
```
karan6181, line 424: What would happen when it is a negative integer ? Can you please also add the test case for the same?

tests/test_encodings.py
```diff
@@ -420,10 +421,43 @@ def test_mds_float64(self, decoded: float, encoded: bytes):
...
+        enc = coder.encode(decoded)
+        assert isinstance(enc, bytes)
+        assert enc == encoded
+
+        dec = coder.decode(encoded)
+        assert isinstance(dec, int)
+        assert dec == decoded
+
+    @pytest.mark.parametrize(('decoded', 'encoded'), [(42.0, b'42.0')])
+    def test_mds_StrFloat(self, decoded: float, encoded: bytes):
```
karan6181, line 436: What would happen when it is a negative float ? Can you please also add the test case for the same?

tests/test_encodings.py
```diff
@@ -420,10 +421,43 @@ def test_mds_float64(self, decoded: float, encoded: bytes):
...
+        coder = mdsEnc.StrFloat()
+        enc = coder.encode(decoded)
+        assert isinstance(enc, bytes)
+        assert enc == encoded
+
+        dec = coder.decode(encoded)
+        assert isinstance(dec, float)
+        assert dec == decoded
+
+    @pytest.mark.parametrize(('decoded', 'encoded'), [(Decimal('4E15'), b'4E+15')])
```
karan6181, line 446: Same here

karan6181 (approved):

PR #391, opened 2023-08-21
Improve shard efficiency of sampling for fractional stream repeats.

knighton:
## Description of changes:

Adds a StreamingDataset argument `sampling_granularity: int = 1`, which determines how many samples are drawn from the same shard at a time when sampling a stream.

At `1`, each sample is drawn independently for the partial repeat if any.

At `100`, up to a hundred samples are pulled at a time from the same shard, limited by the number of un-picked samples left in the shard, for the partial repeat if any.

If shard usage efficiency is especially a concern, you would set `sampling_granularity` to the size of the biggest shard (`500`? `50000`?) in order to fully exhaust one shard a time.

## Merge Checklist:

### General

### Tests


streaming/base/sampling.py
```diff
@@ -23,6 +23,18 @@ def get_sampling(samples_per_shard: NDArray[np.int64], choose: int, granularity:
     Returns:
         NDArray[np.int64]: Array of ephemeral samples chosen per shard.
     """
+    if choose < 0:
```
karan6181, line 26: Why do we need this since we already have a check [here](https://github.com/mosaicml/streaming/blob/main/streaming/base/stream.py#L128) ?

streaming/base/sampling.py
```diff
@@ -23,6 +23,18 @@ def get_sampling(samples_per_shard: NDArray[np.int64], choose: int, granularity:
...
+        raise ValueError('Choose must be a non-negative integer.')
+
+    if granularity <= 0:
```
karan6181, line 29: Can you please add this check in dataset.py since it is a user args and name here is confusing in comparison to `sampling_granularity`?

streaming/base/sampling.py
```diff
@@ -23,6 +23,18 @@ def get_sampling(samples_per_shard: NDArray[np.int64], choose: int, granularity:
...
+        raise ValueError('Granularity must be a positive integer.')
+
+    if seed < 0:
```
karan6181, line 32: Same here. Have it in dataset.py

streaming/base/sampling.py
```diff
@@ -23,6 +23,18 @@ def get_sampling(samples_per_shard: NDArray[np.int64], choose: int, granularity:
...
+        raise ValueError('Seed must be a non-negative integer.')
+
+    if epoch < 0:
```
karan6181, line 35: Can this ever be zero since it is not a user provided args ?

streaming/base/dataset.py
```diff
@@ -265,6 +267,9 @@ class StreamingDataset(Array, IterableDataset):
             of this size, and samples within each block are shuffled. Defaults to ``1 << 18``.
         sampling_method (str): Which sampling method to use, either ``balanced`` or ``fixed``.
             Defaults to ``balanced``.
+        sampling_granularity (int): When picking samples for a stream's final partial repeat,
```
karan6181, line 270: Can you please add more details on what other value means apart from `1` for the users so that they know how to tweak this parameter based on their need?

streaming/base/sampling.py
```diff
@@ -0,0 +1,75 @@
...
+        raise ValueError('Epoch must be a non-negative integer.')
+
+    # Handle whole integer repeat case.
+    num_samples = sum(samples_per_shard)
+    if not choose % num_samples:
```
karan6181, line 40: General comments: Can you please add one line comment as much as possible in this file ? Sometimes, it little hard to understand what each line is doing. Thanks!

snarayan21: +1

streaming/base/sampling.py
```diff
@@ -0,0 +1,75 @@
...
+        shard_ids = np.full(num_granules, shard_id)
+        counts = np.full(num_granules, granularity)
+        if shard_samples % granularity:
+            counts[-1] = shard_samples % granularity
+        pair = shard_ids, counts
+        pairs.append(pair)
+    shard_ids, counts = zip(*pairs)
+    shard_ids = np.concatenate(shard_ids)
+    counts = np.concatenate(counts)
+    num_granules = len(shard_ids)
```
snarayan21, line 58: there are some duplicate variable names that make the code harder to understand/follow. could you change those as well as add comments? otherwise lgtm!!

snarayan21 (changes requested):
just some clarification stuff like comments and duplicate variable names, other than that lgtm

streaming/base/sampling.py
xiaohanzhan-db: Should the boundary check be: granularity <=0 or granularity > samples_per_shard * choose // num_samples?

streaming/base/sampling.py
xiaohanzhan-db: This for loop can be shorten with np-ops, and (hopefully) it's faster, e.g., 

  sp = np.array(samples_per_shard)
  ng, n_shards = (sp + granularity -1)//granularity, len(samples_per_shard)
  shard_ids = np.repeat(list(range(n_shards)), ng)
  counts = np.repeat([granularity]*n_shards, ng)
  counts[np.cumsum(ng)-1] = sp - (ng-1) * granularity  # iiuc, only the last element has a different value which is this difference.

streaming/base/sampling.py
knighton: I am speculating that in the future, a user may want to do something custom that we didn't anticipate, resulting in them calling this method directly instead of indirectly through StreamingDataset.

In that case, it would be correct to validate all its arguments.

That may seem too speculative, but it's practically free to validate all args, so let's play it safe.

knighton: Put another way, if a method exists in the repo, someone somewhere is going to call it (while expecting all args to be validated)

streaming/base/sampling.py
knighton: @xiaohanzhan-db can you explain the second option? The way it's implemented, sampling granularity larger than samples per shard is fine; it will just eat the rest of the shard.

knighton: @karan6181

> Can you please add this check in dataset.py since it is a user args

check added in dataset.py, nice catch

> name here is confusing in comparison to sampling_granularity?

Without having the time to go and verify this has been done uniformly in everywhere, this is actually following the naming convention that we have been using. Also, there is no other kind of granularity in streaming.

Said naming convention:
- StreamingDataset has groups of args to configure the different stages of epoch generation (i.e., sampling, partitioning, shuffling, and now batching)
- In StreamingDataset, these args' names are prefixed with what stage they belong to (because we have a lot of args)
- In the stages themselves, the module and the method are already named after the stage, so we drop the redundant arg name stage prefix

streaming/base/sampling.py
knighton: done

future work: check the required bounds of python random, numpy, torch, etc. "set seed" methods. would be a barrel of monkeys if they are different. then tighten bounds as appropriate. wondering if it's 0 <= seed < (1 << 31)

streaming/base/sampling.py
knighton: The first epoch is zero

For why a check for negative epoch, `epoch` is passed as an arg to this public method and it costs us near-nothing

streaming/base/dataset.py
knighton: added another example

streaming/base/sampling.py
knighton: code is now rewritten to better use numpy and be commented better, please do another round

streaming/base/sampling.py
knighton: incorporated your changes, please have another look

scripts/samples/bench_and_plot.py
```diff
@@ -341,11 +341,11 @@ def bench(args: Namespace, bench_name: str, desc: str, generate: Callable,
     plt.grid(which='minor', ls=':', c='#ddd', lw=0.5)
     ax = plt.gca()
     ax.xaxis.set_major_formatter(ScalarFormatter())
-    ax.xaxis.get_major_formatter().set_scientific(False)
-    ax.xaxis.get_major_formatter().set_useOffset(False)
+    ax.xaxis.get_major_formatter().set_scientific(False)  # pyright: ignore
```
karan6181, line 344: Why do we need `# pyright: ignore` now since the you did not changed anything in this line ?

streaming/base/sampling.py
```diff
@@ -0,0 +1,80 @@
...
+        seed (int): Seed for shuffling sampling granules.
+        epoch (int): Which epoch we are sampling for.
+        use_epoch (bool): Whether to factor epoch into the base seed, or use the same seed across
+            epochs.
+
+    Returns:
+        NDArray[np.int64]: Array of ephemeral samples chosen per shard.
+    """
+    if choose < 0:
+        raise ValueError('Choose must be a non-negative integer.')
```
karan6181, line 27: Can you update the error message with "`choose` must be a positive integer, but got: {choose}"

streaming/base/sampling.py
```diff
@@ -0,0 +1,80 @@
...
+
+    if granularity <= 0:
+        raise ValueError('Granularity must be a positive integer.')
+
+    if seed < 0:
+        raise ValueError('Seed must be a non-negative integer.')
+
+    if epoch < 0:
+        raise ValueError('Epoch must be a non-negative integer.')
```
karan6181, line 36: Can you update the error message with "`epoch` must be a non-negeative integer, but got: {epoch}"

scripts/samples/bench_and_plot.py
knighton: Removed. Something to do about versions

streaming/base/sampling.py
knighton: done

streaming/base/sampling.py
knighton: done

karan6181 (approved):

PR #394, opened 2023-08-23
Expanded range shuffle

snarayan21:
## Description of changes:
Implements a new shuffle - the "expanded range" shuffle, where the range that a shard's samples can appear is expanded, and samples are randomly placed within that range.

Helpful slides: https://docs.google.com/presentation/d/1UHijFFgA0IPUxiOVv4aevGclSc83HKfJ6ae3PZtzc0c/edit?usp=sharing

Suppose we have 10 shards in a canonical node, each shard has 100 samples, and our shuffle block size is set to 500 samples. With py1e, for each shard, its samples will now be distributed over a window of maximum size 500 (equal to shuffle block size). However, these windows cannot cross canonical node boundaries because we don't want overlap between samples from different canonical nodes. So shard 1 will have a window of 300 samples (the max window is 500 samples, but we can't cross the CN boundary, so our window is shortened to 300 samples), shard 2, 400 samples (again, max window is 500 samples, but we can't cross the CN boundary, so our window is shortened to 400 samples), and shard 3-8 will have windows of 500 samples, shard 9, similar to shard 2, has a window of 400 samples (because it can't cross the end of the canonical node), and shard 10 has a window of 300 samples.

Within a canonical node, the total number of shards we need (assuming all shards are the same size) is given by SBS/(# samples per shard). This is the same for algorithms like py1b and py1br. However, when py1b and py1br cross canonical node boundaries, because there is some predownload, as training approaches the end of a canonical node, the predownload will look ahead into the next canonical node. Because the first shuffle block in the next canonical node is shuffled, this means our predownload likely will need to download many shards for the upcoming shuffle block. This results in a spike in downloading, and requires a higher cache limit to store these shards without any negative impact to throughput.

In contrast, with py1e shuffling, as you approach the end of a canonical node, the number of shards you need to proceed with training through the current canonical node approaches 0.5*(SBS/(# samples per shard)). Similarly, at the start of the next canonical node, the number of shards needed to fulfill the first few batches also starts at 0.5*(SBS/(# samples per shard)). This means that I can maintain a small predownload to look ahead into the next canonical node with a lower cache limit, since in total, I need 0.5*(SBS/(# samples per shard)) + 0.5*(SBS/(# samples per shard)) = SBS/(# samples per shard), meaning that the number of shards I need to store per node is constant throughout training.

Additionally, downloading is also more balanced since the shards I need to download steadily ramps up from 0.5*SBS/(# samples per shard) in the beginning of a canonical node to SBS/(# samples per shard). This results in a more smooth downloading curve than algorithms like py1b or even py1br, which both download all the shards needed for a shuffle block within the span of a few batches.

## Issue #, if available:

https://mosaicml.atlassian.net/browse/STR-127

## Merge Checklist:

### General

### Tests


streaming/base/shuffle/py1e.py
```diff
@@ -0,0 +1,121 @@
+# Copyright 2023 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shuffling algorithm that shuffles intra-shard in one place.
```
knighton, line 4: this is the description for py1s, need to tweak it for py1e

streaming/base/shuffle/py1e.py
```diff
@@ -0,0 +1,121 @@
...
+
+Bias in this case merely refers to how we assign samples when we split shards at canonical node
+boundaries, which is non-random in this algorithm. In practice, we found this does not matter to
+convergence, while making us faster.
+"""
+
+import numpy as np
+from numpy.typing import NDArray
+
+from streaming.base.shuffle.py1s import divide_spans
```
knighton, line 16: TODO (eventually): pull out common code

streaming/base/shuffle/py1e.py
```diff
@@ -0,0 +1,121 @@
...
+    # Create each shard's sample ID span (begin, end excl).
+    # also get max shard size to calculate the
+    spans = []
+    num_samples = 0
+    max_shard_size = 0
+    for shard_size in shard_sizes:
+        span = num_samples, num_samples + shard_size
+        spans.append(span)
+        num_samples += shard_size
+        if shard_size > max_shard_size:
```
knighton, line 50: i have generally been following the convention `if LT < GT:` in this repo but w/e

streaming/base/shuffle/py1e.py
```diff
@@ -0,0 +1,121 @@
...
+    run_rng.shuffle(spans)
+
+    # Break the shard spans at canonical node boundaries.
+    # super_spans are the indices of spans that correspond to each canonical node
+    spans, super_spans = divide_spans(spans, num_samples, num_canonical_nodes)
+
+    # Shuffle the span ordering within each canonical node uniquely to this epoch.
+    epoch_rng = np.random.default_rng(seed + epoch)
+    for begin, end in super_spans:
+        # retrieving the spans (shard parts) associated with this canonical node
```
knighton, line 64: conventionally we make comments formal-looking (capitalizing and periods and so on)

streaming/base/shuffle/py1e.py
```diff
@@ -0,0 +1,121 @@
...
+            epoch_rng.shuffle(cn_span_samples)
+            cn_samples[samples_inserted:samples_inserted + (end - begin)] = cn_span_samples
+            samples_inserted += (end - begin)
+
+        # iterate over each span and shift sample indices by gaussian noise
+        cn_sample_offset = 0
+        shifted_samples = np.arange(num_cn_samples).astype(np.float64)
+        for span_size in cn_span_sizes:
+
+            # cutoff is (block_size - span_size)/2, so the span samples
```
knighton, line 93: ` / `

streaming/base/shuffle/py1e.py
```diff
@@ -0,0 +1,121 @@
...
+        for begin, end in cn_spans:
+            # insert span samples into cn_samples array
+            cn_span_samples = np.arange(begin, end)
+            epoch_rng.shuffle(cn_span_samples)
+            cn_samples[samples_inserted:samples_inserted + (end - begin)] = cn_span_samples
+            samples_inserted += (end - begin)
+
+        # iterate over each span and shift sample indices by gaussian noise
+        cn_sample_offset = 0
+        shifted_samples = np.arange(num_cn_samples).astype(np.float64)
```
knighton, line 90: `sample_positions`?

knighton (approved):
some minor nits then LFG

streaming/base/shuffle/py1e.py
snarayan21: done!

streaming/base/shuffle/py1e.py
snarayan21: max_shard_size wasnt even being used, so deleted it

streaming/base/shuffle/py1e.py
snarayan21: will fix these!

snarayan21: i had also added comments on py1b and py1br before. will fix those too.

streaming/base/shuffle/py1e.py
snarayan21: done!

streaming/base/shuffle/py1e.py
snarayan21: yeah that's def more informative

PR #407, opened 2023-08-31
Per Stream Batching

snarayan21:
## Description of changes:
Enables batching where each batch consists of samples from just one stream! Users can access this functionality by setting `batching_method` to `per_stream`.

How it works:
- make partition per stream
- shuffle each partition
- reshape to get global batches per stream, in order of traversal
- discard batches with any padding at all
- construct new overall batch ordering by selecting from stream batches
- pad new batches if necessary
- reshape/transpose back to get final partition

This is a useful feature requested mainly by advanced vision users. For example, each stream can be images of a different aspect ratio and resolution. Using `per_stream` batching means that pytorch now does not throw errors when trying to train with data of different dimensions now, as long as data of the same dimension is in its own stream.

Ran multi-GPU tests with toy datasets to make sure that samples in the same batch were only from a single stream. Ran regression tests over 200 batches using different text datasets to ensure that elastic determinism and resumption still work when using `per_stream` batching. These ran without issue and the results are here: https://wandb.ai/mosaic-ml/batching_methods/reports/new-batching-methods-testing--Vmlldzo1MzIxNjI1

## Issue #, if available:
https://mosaicml.atlassian.net/browse/STR-124

## Merge Checklist:

### General

### Tests


streaming/base/dataset.py
```diff
@@ -311,10 +321,16 @@ def __init__(self,
...
+            raise ValueError(
+                f'Invalid batching method: {batching_method}. Must be one of `random`, `stratified`, or `per_stream.'
+            )
+
         # issue deprecation warning for py1b shuffle algorithm.
         if self.shuffle_algo == 'py1b':
-            warn(
-                'The \'py1b\' shuffle algorithm will soon be deprecated. Please use the more performant \'py1br\' algorithm instead.',
+            warnings.warn(
+                'The \'py1b\' shuffle algorithm will soon be deprecated. Please use the more performant \'py1e\' or \'py1br\' algorithms instead.',
```
karan6181, line 333: Wondering, should we stick to recommending just `py1br` since it is a better alternative of `py1b` because `py1e` shuffling is great but not at par with `py1br` ??

streaming/base/dataset.py
```diff
@@ -765,6 +787,128 @@ def _resample_streams(self, epoch: int) -> Tuple[NDArray[np.int64], NDArray[np.i
         sample_ids = np.concatenate(sample_ids).astype(np.int64)
         return shuffle_units, sample_ids
 
+    def _generate_work_per_stream_batching(self, world: World, epoch: int,
```
karan6181, line 790: Do you know how much time does this method operation takes with today's standard dataset ?

streaming/base/dataset.py
```diff
@@ -765,6 +787,128 @@ def _resample_streams(self, epoch: int) -> Tuple[NDArray[np.int64], NDArray[np.i
...
+                                           sample_in_epoch: int) -> NDArray[np.int64]:
+        """Generate this epoch's arrangement of samples for ``per_stream`` batching.
```
karan6181, line 792: Can we also move this method to `partition` folder since this basically partition the samples across 5D (num physical nodes, ranks per node, workers per rank, batches per worker, batch size) ? this way we keep specific functionality in their respective modules.

tests/test_streaming.py
```diff
@@ -59,6 +59,65 @@ def test_dataloader_epoch_size_no_streams(local_remote_dir: Tuple[str,
...
+    # Build StreamingDataset
+    dataset = StreamingDataset(streams=[stream1, stream2],
+                               shuffle=shuffle,
+                               batch_size=batch_size,
+                               shuffle_seed=seed,
+                               num_canonical_nodes=num_canonical_nodes,
+                               batching_method='per_stream')
+
+    # Build DataLoader
+    dataloader = StreamingDataLoader(dataset=dataset,
```
karan6181, line 103: You can just use `Dataloader` since you are not testing MER !!

tests/test_streaming.py
```diff
@@ -59,6 +59,65 @@ def test_dataloader_epoch_size_no_streams(local_remote_dir: Tuple[str,
...
 
 
+@pytest.mark.parametrize('batch_size', [4])
+@pytest.mark.parametrize('seed', [2222])
+@pytest.mark.parametrize('shuffle', [True])
+@pytest.mark.parametrize('drop_last', [False, True])
+@pytest.mark.parametrize('num_workers', [3, 6])
+@pytest.mark.parametrize('num_canonical_nodes', [4, 8])
+@pytest.mark.usefixtures('local_remote_dir')
+def test_dataloader_per_stream_batching(local_remote_dir: Tuple[str, str], batch_size: int,
```
karan6181, line 69: Please also add a test case of invalid use-case (The exception that your code is throwing)!!

tests/test_streaming.py
```diff
@@ -59,6 +59,65 @@ def test_dataloader_epoch_size_no_streams(local_remote_dir: Tuple[str,
...
+                                     batch_size=batch_size,
+                                     num_workers=num_workers,
+                                     drop_last=drop_last)
+
+    # Ensure that the samples seen in each batch are from the same stream.
+    # Stream 1 has samples 0 to 600, stream 2 has samples 600 and above.
+    for batch in dataloader:
+        samples = batch['sample']
+        first_sample = samples[0]
+        print(first_sample)
```
karan6181, line 113: Please avoid `print` statement in test

streaming/base/dataset.py
```diff
@@ -765,6 +787,128 @@ def _resample_streams(self, epoch: int) -> Tuple[NDArray[np.int64], NDArray[np.i
...
+        for i, partition in enumerate(partition_per_stream):
+            # reshape the partition to be global batches in order of traversal, and count only batches without -1 in them.
+            global_batches_inorder = partition.transpose(3, 2, 0, 1, 4).reshape(
+                -1, batch_size * world.ranks_per_node * world.num_nodes)
+            num_full_batches = np.count_nonzero(np.min(global_batches_inorder, axis=1) >= 0)
+            batches_per_stream.append(num_full_batches)
+            if num_full_batches != global_batches_inorder.shape[0]:
+                logger.warning(
+                    'Because of the `per_stream` batching method, some batches with an inadequate number of samples '
+                    + 'from stream with index ' + str(i) + ' are being dropped.')
```
karan6181, line 857: is `str(i)` a sample index ? I am trying to understand, if this error message gets displays with `str(i)` information, how much can user infer what it is saying !!

streaming/base/dataset.py
```diff
@@ -765,6 +787,128 @@ def _resample_streams(self, epoch: int) -> Tuple[NDArray[np.int64], NDArray[np.i
...
+        # Combine all global batches from all streams into one array.
+        all_partition_batches = np.concatenate(batches_from_partitions)
+
+        # Shuffle seed changes with every epoch so that the order of streams in our batches changes as well.
+        epoch_rng = np.random.default_rng(self.shuffle_seed + epoch)
+
+        # stream_origins is an array that tells us which stream each batch is using.
+        stream_origins = np.concatenate(
+            [np.full(n_batch, i) for i, n_batch in enumerate(batches_per_stream)])
+        epoch_rng.shuffle(stream_origins)
```
karan6181, line 869: Does this shuffling logic downloads extra shard files when training runs for more than one epoch ?

tests/test_streaming.py
```diff
@@ -59,6 +59,65 @@ def test_dataloader_epoch_size_no_streams(local_remote_dir: Tuple[str,
...
+                   num_samples=200,
+                   size_limit=1 << 8)
+    # stream 2 has samples 600 and above. This lets us differentiate between the samples from each stream
+    convert_to_mds(out_root=remote2,
+                   dataset_name='sequencedataset',
+                   num_samples=300,
+                   offset=600,
+                   size_limit=1 << 8)
+
+    stream1 = Stream(local=local1, remote=remote1)
```
karan6181, line 91: Does other streams args `proportion`, `choose`, and `repeat` have any impact or in other words, does the behavior still remain intact ? Does you think adding a test case with other stream args is valuable or no ?

tests/test_streaming.py
```diff
@@ -59,6 +59,65 @@ def test_dataloader_epoch_size_no_streams(local_remote_dir: Tuple[str,
             assert samples_seen == epoch_size
 
 
+@pytest.mark.parametrize('batch_size', [4])
```
karan6181, line 62: Can you add a `batch_size` which is non-divisible of `num_samples/stream` and validate if the last samples are being dropped or not or is this already being handled by other params such as `num_workers` and `num_canonical_nodes` ?

streaming/base/dataset.py
snarayan21: the shuffle quality is just a bit lower, imo in most cases py1e > py1br > py1b?

streaming/base/dataset.py
snarayan21: no i'm not sure. want me to measure? probably like a couple seconds more on large datasets, if at all.

streaming/base/dataset.py
snarayan21: yeah sure!

tests/test_streaming.py
snarayan21: Ah true, will change!

streaming/base/dataset.py
snarayan21: `str(i)` is the index of the stream's partition in the partition_per_stream array. i'll change the variable name to reflect that

streaming/base/dataset.py
snarayan21: No -- we won't be downloading more shard files when training on more than one epoch. This shuffle is to determine which stream each batch will come from. So if we have 3 streams, originally the `stream_origins` array will look something like `[0,0,0,1,1,1,1,2,2]`, assuming we have 3 batches from stream 0, 4 batches from stream 1, and 2 batches from stream 2. Then we shuffle the `stream_origins` array to get the epoch batch ordering of `[1,2,1,0,1,1,0,2,0]`, for example. This ordering will change between epochs but will use the same samples/shards from each stream.

tests/test_streaming.py
snarayan21: That logic is unchanged by the new batching methods, so i don't think it's worth to test. Before batching methods are applied, the `proportion`, `choose`, and `repeat` of each stream is already calculated, and the batching methods use the number of samples per stream corresponding to these param values.

tests/test_streaming.py
snarayan21: done!

streaming/base/dataset.py
snarayan21: moving to separate `batching` folder instead.

streaming/base/dataset.py
snarayan21: fixed!

streaming/base/batching/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import TYPE_CHECKING
+
+import numpy as np
+from numpy.typing import NDArray
+
+from streaming.base.batching.per_stream import generate_work_per_stream_batching
+from streaming.base.batching.random import generate_work_random_batching
+from streaming.base.world import World
+
+if TYPE_CHECKING:
```
karan6181, line 16: Wondering, why do we need this? Were you seeing any exception or was it circular imports ?

karan6181 (approved):

streaming/base/batching/__init__.py
snarayan21: was getting a circular imports error because I needed `StreamingDataset` for type checking in `generate_work` but then `StreamingDataset` also needed to import `generate_work`. So this takes care of that.

Issue #417, opened 2023-09-06
StreamingDataset deadlocks when cache_limit is smaller than the size of the shards


Hubert-Bonisseur:
## To reproduce_

1. Create a dataset with size_limit = 10Mb for instance
2. Load it with cache_limit = 1Mb for instance
3. Access the first item
```python
    dataset = StreamingDataset(local="cache", remote="path_to_remote",
                                  cache_limit="1Mb")
    i = iter(dataset)
    print("accessing the iterator...")
    print(next(i))
    print("exiting...")
```
Nothing will be printed after "accessing the iterator..."
## Expected behavior
An error message should tell the user to increase the cache_limit

karan6181:
Hi @Hubert-Bonisseur , thanks for filing the issue. We would fix this soon!

PR #420, opened 2023-09-08
Raise an exception if cache limit is too low

karan6181:
## Description of changes:
- If `cache_limit` is lower than a shard file size or lower than a few shard files sizes, raise an exception to raise a cache_limit. This is more of a better user experience. Hard to see a practical use-case where the cache_limit is this much low.

## Issue #, if available:
#417 

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -395,7 +395,6 @@ class StreamingDataset(Array, IterableDataset):
         self.shards_per_stream = np.zeros(self.num_streams, np.int64)
         self.sample_offset_per_stream = np.zeros(self.num_streams, np.int64)
         self.samples_per_stream = np.zeros(self.num_streams, np.int64)
-        self.max_shard_size_across_all_streams = 0
         for stream_id, stream in enumerate(self.streams):
             stream_shards = stream.get_shards(world)
             num_stream_samples = sum(map(len, stream_shards))
@@ -403,9 +402,6 @@ class StreamingDataset(Array, IterableDataset):
                 index_filename = os.path.join(stream.local, stream.split, get_index_basename())
                 raise RuntimeError(f'Stream contains no samples: {index_filename}.')
             stream_per_shard += [stream_id] * len(stream_shards)
-            # Includes the size of raw (decompressed) and zip (compressed) shard files.
-            self.max_shard_size_across_all_streams = max(self.max_shard_size_across_all_streams,
-                                                         stream_shards[0].get_max_size())
             self.shard_offset_per_stream[stream_id] = len(self.shards)
             self.shards_per_stream[stream_id] = len(stream_shards)
             self.sample_offset_per_stream[stream_id] = self.num_samples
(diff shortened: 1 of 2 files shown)


streaming/base/dataset.py
```diff
@@ -395,13 +395,17 @@ def __init__(self,
...
         for stream_id, stream in enumerate(self.streams):
             stream_shards = stream.get_shards(world)
             num_stream_samples = sum(map(len, stream_shards))
             if not num_stream_samples:
                 index_filename = os.path.join(stream.local, stream.split, get_index_basename())
                 raise RuntimeError(f'Stream contains no samples: {index_filename}.')
             stream_per_shard += [stream_id] * len(stream_shards)
+            # Includes the size of raw (decompressed) and zip (compressed) shard files.
+            self.max_shard_size_across_all_streams = max(self.max_shard_size_across_all_streams,
+                                                         stream_shards[0].get_max_size())
```
snarayan21, line 408: there's a weird edge case here that honestly might not be worth addressing -- you might merge the index.json files from two different streams where stream A has shard file size 5x smaller than stream B, and the first shard in the combined index.json file is from stream A and it becomes the max shard in this loop. Then, we will still hang when trying to download any shard from stream B. This is pretty obscure though... see later comment for a fix

streaming/base/dataset.py
```diff
@@ -420,6 +424,13 @@ def __init__(self,
                 raise ValueError(f'Minimum cache usage ({min_cache_usage} bytes) is larger than ' +
                                  f'the cache limit ({self.cache_limit} bytes). Please raise ' +
                                  f'`cache_limit`.')
+            if self.cache_limit < 4 * self.max_shard_size_across_all_streams:
+                raise ValueError(f'Cache limit ({self.cache_limit} bytes) is too low. ' +
+                                 f'Increase the `cache_limit` to at-least four times the ' +
+                                 f'largest shard size ({self.max_shard_size_across_all_streams} ' +
+                                 f'bytes) which includes raw (decompressed) and zip ' +
+                                 f'(compressed) file size. Recommendation is to provide a ' +
+                                 f'`cache_limit` as high as possible to avoid thrashing.')
```
snarayan21, line 434: so to get the true max shard size, here you could add the line:

```
self.max_shard_size_across_all_streams = np.array([shard.get_max_size() for shard in self.shards])
```

and remove the other logic right?

snarayan21 (commented):
other than the logic for getting the max shard size, lgtm!

streaming/base/dataset.py
karan6181: No, that's a valid use-case. The only reason I did not wanted to iterate all the shards is because of the time it takes to perform the whole operation. But your use-case is a valid one.

streaming/base/dataset.py
karan6181: Done in https://github.com/mosaicml/streaming/pull/420/commits/9f7f57fbc64f9c5fa68a4a3ace1b6da70ce508d5. Also, wrap it with `max` to get the maximum value.

snarayan21: yeah forgot that! ty!

snarayan21 (approved):
lgtm!

PR #423, opened 2023-09-08
Remove torchtext

mvpatel2000:
## Description of changes:

Move torchtext to docs as it is only used by docs
image

This is causing issues with torch nightly


mvpatel2000:
Update: removing based on @karan6181 's suggestion

karan6181 (approved):

PR #425, opened 2023-09-09
Fix deadlock

### LINKED ISSUE #386: deadlock when process exits before end of epoch
## Environment

- OS: AlmaLinux 8.8
- Hardware (GPU, or instance type): no gpu

## To reproduce
1. take any StreamingDataset
2. create an iterator
3. do one iteration of the iterator, but do NOT finish an epoch.
4. try to exit the program.

The following script will deadlock on exit:
```python
from streaming import MDSWriter
from streaming import StreamingDataset

### generate a sample dataset...
# directory for a sample dataset
out_root = 'deadlock_data'

# A dictionary of input fields to an Encoder/Decoder type
columns = {
    'anumber': 'int'
}

# some sample data
samples = [
    {
        'anumber': 123123
    }
    for _ in range(1000)
]

# write out the sample data
print("writing data...")
with MDSWriter(out=out_root, columns=columns) as out:
    for sample in samples:
        out.write(sample)
### finished generating the sample dataset


print("creating dataset object...")
remote_dir = 'deadlock_data'
local_dir = 'local_cache'
dataset = StreamingDataset(remote=remote_dir, local_dir=local_dir, split=None)
i = iter(dataset)
print("accessing the iterator...")

print(next(i))

print("exiting...")

# process will now hang on exit...

```

## Expected behavior
The program should exit.

## Additional context
I believe this is caused by the threads spawned by the iterator (`_prepare_thread`  and `_ready_thread`) never exiting. When the interpreter exits, it waits for these threads and so will hang. One hack to fix it is to add the following:
```python
import threading
from streaming.base.dataset import _IterState
def kill_iterators(dataset):
    if hasattr(dataset, "_iterator"):
        with dataset._iterator._lock:
            if dataset._iterator._state == _IterState.ITERATING:
                dataset._iterator._state = _IterState.EXITING
threading._register_atexit(kill_iterators, dataset)
```
Note that we need to use `threading._register_atexit` rather than the `atexit` module because the code that waits on the threads happens before `atexit`. Also, it does not work to call `dataset._iterator.exit()` inside `kill_iterators` because this will wait for two threads to exit, which causes its own deadlock if the process has been forked because one fork will not have any threads to exit.

acutkosky:
## Description of changes:

Add a hook that runs at process exit telling the worker threads in StreamingDataset to exit.
This prevents these threads from blocking process exit.

pyright does not like using the "private" method `threading._register_atexit`, but we cannot use the regular `atexit` module because it would run after the deadlock. So there is a pyright ignore directive added.

## Issue #386

- Fixes #386

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -11,7 +11,7 @@ from concurrent.futures import ThreadPoolExecutor, wait
 from concurrent.futures._base import Future
 from enum import IntEnum
 from math import ceil
-from threading import Event, Lock, _register_atexit
+from threading import Event, Lock, _register_atexit  # pyright: ignore
 from time import sleep, time_ns
 from typing import Any, Dict, Iterator, Optional, Sequence, Tuple, Union
 
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -6,12 +6,13 @@
 import json
 import logging
 import os
+import sys
 import warnings
 from concurrent.futures import ThreadPoolExecutor, wait
 from concurrent.futures._base import Future
 from enum import IntEnum
 from math import ceil
-from threading import Event, Lock, _register_atexit  # pyright: ignore
+from threading import Event, Lock
 from time import sleep, time_ns
 from typing import Any, Dict, Iterator, Optional, Sequence, Tuple, Union
 
@@ -103,7 +104,12 @@ class _Iterator:
         self._state = 0
         self._num_exited = 0
 
-        _register_atexit(self.non_blocking_exit)
+        if sys.version_info[1] <= 8:
+            import atexit
+            atexit.register(self.non_blocking_exit)
+        else:
+            from threading import _register_atexit  # pyright: ignore
+            _register_atexit(self.non_blocking_exit)
 
     def non_blocking_exit(self) -> None:
         """Signal threads to exit without blocking.
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -104,7 +104,14 @@ class _Iterator:
         self._state = 0
         self._num_exited = 0
 
-        if sys.version_info[1] <= 8:
+        # python will attempt to join all threads on shutdown.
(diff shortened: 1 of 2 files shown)


karan6181:
@acutkosky looks like python 3.8 does not have `_register_atexit()` method since the installation is failing `ImportError: cannot import name '_register_atexit' from 'threading' (/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/threading.py)`

acutkosky:
Ah, looks like starting from 3.9 on the threading module stopped using `atexit` to register cleanup of threads in favor of the new `_register_atexit`, which now gets called before `atexit`. So, for 3.8 we can use `atexit` and later we can use `threading._register_atexit`. It's a bit hacky though, happy to hear more elegant ideas.

I added a simple test too.

streaming/base/dataset.py
```diff
@@ -103,6 +104,22 @@ def __init__(self, sample_ids: NDArray[np.int64]) -> None:
         self._state = 0
         self._num_exited = 0
 
+        if sys.version_info[1] <= 8:
```
karan6181, line 107: Can you please add a comment here why we had to made this changes ? This will make a code more readable. Thank You!

karan6181:
> Ah, looks like starting from 3.9 on the threading module stopped using `atexit` to register cleanup of threads in favor of the new `_register_atexit`, which now gets called before `atexit`. So, for 3.8 we can use `atexit` and later we can use `threading._register_atexit`. It's a bit hacky though, happy to hear more elegant ideas.
> 
> I added a simple test too.

It looks good to me. We will remove the if/else logic once streaming stops supporting python 3.8.

streaming/base/dataset.py
acutkosky: sure, added.

acutkosky:
great, thanks!

karan6181 (approved):

PR #448, opened 2023-09-26
Add a retry logic with backoff and jitter

karan6181:
## Description of changes:
- Add a retry logic with backoff and jitter to upload and download functionality
- Enqueue the index.json upload after all the shard uploads has been finished.
- Kill the main thread if one of the upload threads receives Exception.

## Issue #, if available:
STR-136
https://github.com/mosaicml/streaming/issues/451

### Integration testing
**- UC Volume**
  - Verified that I am able to see the `REQUEST_LIMIT_EXCEEDED` exception by spinning approx. 500 processes where it hits the 400 request limit per secs. 
  - Having more than one attempts doesn't crash the job since it retries downloading the shard file again.
  - Verified FileNotFound Exception from UC Volume by providing a corrupt path.
 
**- Other cloud providers**
  - Tested invalid path with S3 and OCI and able to see the relevant FileNotFound Exception.
  - Ran single node model training on S3 and OCI with correct dataset path and I don't see an Exception.

## Merge Checklist:

### General

### Tests


streaming/base/dataset.py
```diff
@@ -1284,6 +1284,9 @@ def _ready_thread(self, it: _Iterator) -> None:
                 self.prepare_shard(shard_id, False)
             # Wait for a shard file to download completely.
             while self._shard_states[shard_id] != _ShardState.LOCAL:
+                # Background thread or a main process crashed, terminate this thread.
```
snarayan21, line 1287: if i'm understanding correctly, the bug here is that in the case when we're waiting for a shard file to download completely, if there is an exception we won't terminate because we're not checking `self._event.is_set()` which is set on exception. In other cases we do check for this so the threads terminate in other cases. Right?

streaming/base/storage/upload.py
```diff
@@ -222,23 +229,29 @@ def upload_file(self, filename: str):
...
-                local_filename,
-                obj.netloc,
-                obj.path.lstrip('/'),
-                Callback=lambda bytes_transferred: pbar.update(bytes_transferred),
-            )
-        self.clear_local(local=local_filename)
+
+        @retry(num_attempts=self.retry)
+        def _upload_file():
+            print('number of retries')
```
snarayan21: did you mean to print the number of retries here? otherwise may want to delete the line.

streaming/base/util.py
```diff
@@ -222,7 +225,8 @@ def retry(exc_class: TCallable) -> TCallable:
     ...
 
 
-# error: Type "(TCallable@retry) -> TCallable@retry" cannot be assigned to type "(func: Never) -> Never"
+# error: Type "(TCallable@retry) -> TCallable@retry" cannot be assigned to type
+# "(func: Never) -> Never"
 def retry(  # type: ignore
```
snarayan21, line 230: Do you want to use this package instead? https://pypi.org/project/retry/

snarayan21 (approved):
Some minor comments, other than that lgtm. Using the package might be easier down the road but this also works

streaming/base/storage/upload.py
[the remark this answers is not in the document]
karan6181: Removed since I accidentally pushed it since I added for testing.

karan6181: Removed since I accidentally pushed it since I added for testing.

streaming/base/util.py
karan6181: It is a stale project where last update was in 2016. Hence, it would be risky to use it.

streaming/base/dataset.py
karan6181: First of all, the `ThreadPoolExecutor` does not provide a utility to stop running tasks, we can update our target task functions to stop running when a thread-safe flag is set and this can be achieved using [**threading.Event**](https://docs.python.org/3/library/threading.html) call. There are two threads that streaming uses, `_prepare_thread()` which downloads the shard file, and `_ready_thread()` which ensures the samples are ready to yield. If samples are not ready to yield, it waits for the shard file to present locally (which `_prepare_thread()` is working on to download the shard file from remote location to a local directory). Now, take a case where the `_prepare_thread()`  crashed due to any reason, in that case, `_prepare_thread()`  sets the threading event and exit. Now, since the `_ready_thread()` is in the infinite loop, there is no way to notify `_ready_thread()` that the `_prepare_thread()` is crashed. One way of doing this is that we check for the `threading.Event` in that infinite while loop and exit as soon as the event is set.

knighton (approved):
Noice

PR #449, opened 2023-09-26
Add merge index file utility

XiaohanZhangCMU:
## Description of changes:

- Add merge index file utility to streaming.base.util. Users can use it to merge subdirectories index files from a MDS dataset. If an index file is present locally, it will be used directly, otherwise, a download from remote will be carried out to get the remote file first and then merge locally. 
- Add an exist_ok parameter to CloudUploader to allow the instantiation of CloudUploader on an non-empty directory
- Add list_objects to S3, GCS, OCI.
- Change dataframeToMDS internally, to use the new merge index function

### Integration testing

- Tests coverage on databricks notebook platform 
with ``out`` being the merged index output location, the integration tests cover
1. out = '/tmp/folder': expected "FileNotFoundError", because driver cannot directly access workers' local disk.
2. out = '/Volumes/bucket_name/folder': Merged. Driver and workers both have access to Volume
3. out = ('/tmp/folder', 's3://streaming-test-bucket/folder'):  Merged. Driver cannot access workers' local disk but download the files from remote buckets to merge
4. out = 's3://streaming-test-bucket/folder': Merged. Driver download all remote files from cloud to merge then upload back to s3. 
A shard integrity check is performed to verify the merged index works as intended. 

## Issue #, if available:
STR-134

## Merge Checklist:

### General

### Tests


streaming/base/util.py
```diff
@@ -196,3 +203,126 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+
+    # This is the index json file name, e.g., it is index.json as of 0.6.0
+    index_basename = get_index_basename()
+
+    if os.path.exists(os.path.join(out, index_basename)) and overwrite:
+        logger.warning('Merged index already exists. no index merged if overwrite=False')
+        return
+
+    # Prepare a temp folder to download index.json rom remote if necessary. Removed in the end.
+    temp_root = mkdtemp()
```
Skylion007, line 250: You should use a construct like
`with tempfile.TemporaryDirectory() as tmp:` 
to ensure the temporary directory (and it's contents are always deleted even on exceptions)

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -274,6 +231,7 @@ def write_mds(iterator: Iterable):
     out = mds_kwargs['out']
     keep_local = False if 'keep_local' not in mds_kwargs else mds_kwargs['keep_local']
     cu = CloudUploader.get(out, keep_local=keep_local)
+    print('cu.local = ', cu.local)
```
karan6181, line 234: nit: remove print statement

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -202,11 +160,10 @@ def write_mds(iterator: Iterable):
             raise RuntimeError('TaskContext.get() returns None')
 
         if isinstance(mds_path, str):  # local
-            output = os.path.join(mds_path, f'{id}')
-            out_file_path = output
+            output_path = output = os.path.join(mds_path, f'{id}')
         else:
             output = (os.path.join(mds_path[0], f'{id}'), os.path.join(mds_path[1], f'{id}'))
-            out_file_path = output[0]
+            output_path = ','.join(output)
```
karan6181, line 166: Why this? Pandas doesn't take a list of strings ?

streaming/base/storage/upload.py
```diff
@@ -155,7 +161,7 @@ def __init__(self,
             self.local = out[0]
             self.remote = out[1]
 
-        if os.path.exists(self.local) and len(os.listdir(self.local)) != 0:
+        if not exist_ok and os.path.exists(self.local) and len(os.listdir(self.local)) != 0:
```
karan6181, line 164: Are we emptying the `out` directory if `exist_ok` is `True` ? or what is behavior when `exist_ok` is `True` ? There would be some stale shard files if we don't empty it before writing right ?

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+        out (Union[str, Tuple[str, str]]): path to put the merged index file
+        keep_local (bool): Keep local copy of the merged index file. Defaults to ``True``
+        overwrite (bool): Overwrite merged index file in out if there exists one.Defaults to ``True``
+        download_timeout (int): The allowed time for downloading each json file
+            defaults to 60, same as streaming.download_file
+
+    Returns:
+        int: count of files downloaded during function call
+    """
+    # Import here to avoid circular import error
```
karan6181, line 238: See if you can use below code to avoid circular imports

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from streaming.base.storage.download import download_file
```

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+
+            for each  url in folder_urls, if url is
+                1. tuple (local, remote): check if local is accessible.
+                    -> Yes: use local index to merge
+                    -> No:  download from remote first, then merge
+                2. str (local path): use local path to merge.
+                    raise FileNotFoundError if any local index is not accessible
+                3. str (remote url): download to a temp directory first, then merge
+
+        out (Union[str, Tuple[str, str]]): path to put the merged index file
```
karan6181, line 229: Wondering, why we do need `out` argument in this method? Generally, for merging index, user would provide either `merge_index(local_dir, remote_dir)` path or `merge_index(local_dir)` or `merge_index(remote_dir)`. Depending on which use-case out of above all three, the final merged index.json would either store in local_dir or remote_dir or both. We don't expect the final merged index.json to be saved else apart from above three location.

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+
+    cu = CloudUploader.get(out, keep_local=True, exist_ok=True)
+    if os.path.exists(os.path.join(cu.local, index_basename)) and overwrite:
+        logger.warning('Merged index already exists locally. no index merged if overwrite=False')
+        return 0
+
+    # Remove '/' from right, so os.path.basename gives relative path to each folder
+    urls = []
+    for url in folder_urls:
+        if type(url) is str:
```
karan6181, line 257: nit: this also works but I would recommend `isinstance(url, str)` since it is faster than `type` check.

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -96,6 +96,8 @@ def dataframe(self):
     def test_end_to_end_conversion_local_nocolumns(self, dataframe: Any, keep_local: bool,
                                                    merge_index: bool,
                                                    local_remote_dir: Tuple[str, str]):
+        print('keep local = ', keep_local)
```
karan6181, line 99: nit: remove all print statements

tests/resources/naive_MDSdataset/25/index.json
```diff
@@ -0,0 +1,31 @@
+{
```
karan6181, line 1: Do you need this index.json and shard files ?

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+                3. str (remote url): download to a temp directory first, then merge
+
+        out (Union[str, Tuple[str, str]]): path to put the merged index file
+        keep_local (bool): Keep local copy of the merged index file. Defaults to ``True``
+        overwrite (bool): Overwrite merged index file in out if there exists one.Defaults to ``True``
+        download_timeout (int): The allowed time for downloading each json file
+            defaults to 60, same as streaming.download_file
+
+    Returns:
+        int: count of files downloaded during function call
```
karan6181, line 236: Do we really care about any return from this method ? Would the `count of files downloaded` useful to users ? My 2 cents would be not return anything, if there is an issue, then the method should raise an exception.

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+                # If download is needed, download url from remote to temp_root
+                path = urllib.parse.urlparse(remote).path
+                local = os.path.join(temp_root, path.lstrip('/'))
+                try:
+                    remote_url = os.path.join(remote, index_basename)
+                    local_path = os.path.join(local, index_basename)
+                    download_file(remote_url, local_path, download_timeout)
+                    n_downloads += 1
+                except Exception as ex:
+                    raise RuntimeError(f'failed to download index.json {url}') from ex
```
karan6181, line 299: nit: `failed` -> `Failed`

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+
+            if not (os.path.exists(local)):
+                raise FileNotFoundError(
+                    'Folder {local} does not exit or cannot be acceessed by the current process')
```
karan6181, line 303: `f` string is missing. `f'Folder {local} does not exit or cannot be acceessed by the current process'`

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+                    -> No:  download from remote first, then merge
+                2. str (local path): use local path to merge.
+                    raise FileNotFoundError if any local index is not accessible
+                3. str (remote url): download to a temp directory first, then merge
+
+        out (Union[str, Tuple[str, str]]): path to put the merged index file
+        keep_local (bool): Keep local copy of the merged index file. Defaults to ``True``
+        overwrite (bool): Overwrite merged index file in out if there exists one.Defaults to ``True``
+        download_timeout (int): The allowed time for downloading each json file
+            defaults to 60, same as streaming.download_file
```
karan6181, line 233: value `60` not in sync with the default value of `100`. Also, we don't need to mention `same as streaming.download_file`.

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+                1. tuple (local, remote): check if local is accessible.
+                    -> Yes: use local index to merge
+                    -> No:  download from remote first, then merge
+                2. str (local path): use local path to merge.
+                    raise FileNotFoundError if any local index is not accessible
+                3. str (remote url): download to a temp directory first, then merge
+
+        out (Union[str, Tuple[str, str]]): path to put the merged index file
+        keep_local (bool): Keep local copy of the merged index file. Defaults to ``True``
+        overwrite (bool): Overwrite merged index file in out if there exists one.Defaults to ``True``
```
karan6181, line 231:
```suggestion
        overwrite (bool): Overwrite merged index file in out if there exists one. Defaults to ``True``
```

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+        else:
+            # If url is a remote, download = True, False otherwise
+            download = urllib.parse.urlparse(url).scheme != ''
+
+        # As long as one index file needs download, we download them all to keep it simple
+        if download:
+            break
+
+    print('download = ', download)
+    # Prepare a temp folder to download index.json rom remote if necessary. Removed in the end.
```
karan6181, line 277:
```suggestion
    # Prepare a temp folder to download index.json from remote if necessary. Removed in the end.
```

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+    with tempfile.TemporaryDirectory() as temp_root:
```
karan6181, line 278: Would recommend logging the value of `temp_root` path so that the user aware of the path if passing just remote path.

streaming/base/util.py
karan6181, line 303:
```suggestion
                    f'Folder {local} does not exit or cannot be accessed.')
```

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+            json.dump(obj, outfile)
+
+        # Upload merged index to remote if out has remote part
+        # Otherwise, move it from temp root to out location
+        shutil.move(merged_index_path, cu.local)
+        if cu.remote is not None:
+            cu.upload_file(index_basename)
+
+        # Clean up
+        # shutil.rmtree(temp_root, ignore_errors=True)
```
karan6181, line 337: Extra line ?

tests/test_util.py
```diff
@@ -105,3 +107,68 @@ def test_clean_stale_shared_memory():
...
+
+
+@pytest.mark.parametrize('folder_urls', [
+    'local_accessible', 'remote', 'local_unaccessible', 'tuple_local_accessible',
+    'tuple_partial_local_accessible'
+])
+@pytest.mark.parametrize('out', ['local_str', 'remote_str', 'tuple'])
+@pytest.mark.usefixtures('local_remote_dir')
+@pytest.mark.parametrize('keep_local', [True, False])
+def test_merge_index(local_remote_dir: Tuple[str, str], keep_local: Any, folder_urls: Any,
```
karan6181, line 119: Suggestion: To test the method functionality, I would not worried about cloud provider URLs. Just test whether merge index.json happens correctly or not using local directory. For example:

Test: `folder_urls` is just a `local_dir`, create two to three sub-dataset and apple `merge_index` method. Validate the number of shard files of the final merged index.json. For validation number of shard files, you can get the `basename` field in the `set` and count the length of `set`.

streaming/base/converters/dataframe_to_mds.py
xiaohanzhan-db: It's because mapInPandas cannot pickle a tuple to return it to the driver.

streaming/base/storage/upload.py
xiaohanzhan-db: yes, this Is the most concerning part. 
with "exist_ok=True", it allows merge_index utility to create UploadClouder on an existing folder without an error is thrown. But it also introduces the risk where the user accidentally set exist_ok=True, so the stale shards are not removed before writing new shard files to it.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -289,11 +247,16 @@ def write_mds(iterator: Iterable):
     partitions = dataframe.mapInPandas(func=write_mds, schema=result_schema).collect()
 
     if merge_index:
-        do_merge_index(partitions, mds_path)
+        folder_urls = []
+        for row in partitions:
+            if ',' in row['mds_path']:
```
snarayan21, line 252: this if-else block is also because of the tuple pickling?

streaming/base/util.py
xiaohanzhan-db: Done.

tests/base/converters/test_dataframe_to_mds.py
xiaohanzhan-db: Done.

streaming/base/storage/upload.py
```diff
@@ -72,6 +75,7 @@ def get(cls,
                 shard file or remove it after uploading. Defaults to ``False``.
             progress_bar (bool): Display TQDM progress bars for uploading output dataset files to
                 a remote location. Default to ``False``.
+            exist_ok (bool): Throw error if out already exists and not empty. Defaults to ``False``.
```
snarayan21, line 78: Can you clarify the description for the `exist_ok` arg for the behavior when `True` and `False`? Right now it seems like if I set this to `True` then it will throw error, but that's not actually the case right?

streaming/base/storage/upload.py
```diff
@@ -134,6 +139,7 @@ def __init__(self,
                 shard file or remove it after uploading. Defaults to ``False``.
             progress_bar (bool): Display TQDM progress bars for uploading output dataset files to
                 a remote location. Default to ``False``.
+            exist_ok (bool): Throw error if out already exists and has contents. Defaults to ``False``.
```
snarayan21, line 142: same comment as above, wherever `exist_ok` is in a docstring

streaming/base/util.py
xiaohanzhan-db: Good point~ Removed.

streaming/base/util.py
xiaohanzhan-db: Done

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
     return f'Streaming was installed without {package_name} support. ' + \
             f'To use {package_name} related packages with Streaming, run ' + \
             f'`pip install \'mosaicml-streaming[{package_name}]\'`.'
+
+
+def merge_index(folder_urls: List[Union[str, Tuple[str, str]]],
```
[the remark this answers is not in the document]
snarayan21: Does this also encompass cases where we want to merge index.json files from multiple different folders that aren't necessarily subdirectories of the parent directory? I know some customer was asking for this functionality -- @karan6181 we had discussed it right?

streaming/base/util.py
snarayan21: @karan6181 I think the current default `download_timeout` is 60 in StreamingDataset

xiaohanzhan-db: Done.

streaming/base/util.py
xiaohanzhan-db: Done

streaming/base/util.py
xiaohanzhan-db: Done

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+    for url in folder_urls:
+        if type(url) is str:
+            urls.append(url.rstrip('/').strip())
+        else:
+            urls.append((url[0].rstrip('/').strip(), url[1].rstrip('/').strip()))
+
+    # Determine if we need to call download_file.
+    download = False
+    for url in urls:
+        if type(url) is tuple:
```
snarayan21, line 265: nit: similar to Karan's comment, `isinstance(url, tuple)` also works

streaming/base/util.py
xiaohanzhan-db: Done.

streaming/base/util.py
xiaohanzhan-db: Done.

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+            download = not os.path.exists(url[0])
+        else:
+            # If url is a remote, download = True, False otherwise
+            download = urllib.parse.urlparse(url).scheme != ''
+
+        # As long as one index file needs download, we download them all to keep it simple
+        if download:
+            break
+
+    print('download = ', download)
```
snarayan21, line 276: you can delete this print statement unless you want to log something?

tests/test_util.py
```diff
@@ -105,3 +107,68 @@ def test_clean_stale_shared_memory():
...
+                     out: Any):
+    """Example input urls to test based on folder accessibility to the driver
+        ['pwd()/tests/resources/naive_MDSdataset/25/', ...] --> all accessible locally
+        ['gs://mybucket/mdsdata/25/'...] --> all remote urls, assume all accessible remotely
+        ['/path/never/exists/25',... ] --> none accessible locally
+        [('/path/never/exists/25', 'gs://mybucket/mdsdata/25/'), ...] --> all accessible remotely but not locally
+        [('pwd()/tests/resources/naive_MDSdataset/25/', 'gs://mybucket/mdsdata/25/'), ...] --> all accessible both locally and remotely
+        [('pwd()/tests/resources/naive_MDSdataset/25/', 'gs://mybucket/mdsdata/25/'),
```
snarayan21, line 127: So if I'm merging index files from both a local and a remote, then my result will have some shards that are local and some that are remote. I think we do hypothetically support this functionality in streaming since each shard gets its own Reader object, but would be great if you could run a test that confirms this!

snarayan21 (commented):
quite a hefty PR! some clarification comments and changes in addition to what Karan pointed out.

streaming/base/converters/dataframe_to_mds.py
xiaohanzhan-db: Yes, it is.

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+
+        # merge index files into shards
+        shards = []
+        for partition in partitions:
+            partition_index = f'{partition}/{index_basename}'
+            mds_partition_basename = os.path.basename(partition)
+            obj = json.load(open(partition_index))
+            for i in range(len(obj['shards'])):
+                shard = obj['shards'][i]
+                for key in ('raw_data', 'zip_data'):
```
knighton, line 314: raw_data, zip_data, raw_meta, zip_meta

streaming/base/converters/dataframe_to_mds.py
Skylion007: Are we sure the name can never have a comma? Should we use a more exotic unicode character to reduce collisions?

XiaohanZhangCMU: hmm, good point. Any suggestion what unicode character to use?

streaming/base/util.py
XiaohanZhangCMU: Will keep this method as a helper as it is convenient for dataframeToMDS. Adding a wrapper merge_index function with just a str path as argument.

streaming/base/util.py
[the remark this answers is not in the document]
XiaohanZhangCMU: As we dicussed, this is not going to be covered by this PR.

streaming/base/util.py
XiaohanZhangCMU: Done!

streaming/base/util.py
XiaohanZhangCMU: Changed to isinstance.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -201,12 +159,10 @@ def write_mds(iterator: Iterable):
         else:
             raise RuntimeError('TaskContext.get() returns None')
 
-        if isinstance(mds_path, str):  # local
-            output = os.path.join(mds_path, f'{id}')
-            out_file_path = output
+        if mds_path[1] == '':
```
karan6181, line 162: Can you please add a comment when will `mds_path[1]` be empty ?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -237,10 +193,13 @@ def write_mds(iterator: Iterable):
                         raise RuntimeError(f'failed to write sample: {sample}') from ex
                         count += 1
 
-        yield pd.concat(
-            [pd.Series([out_file_path], name='mds_path'),
-             pd.Series([count], name='fail_count')],
-            axis=1)
+        yield pd.concat([
```
karan6181, line 196: Is this per partition ?

streaming/base/storage/upload.py
```diff
@@ -73,6 +74,7 @@ def get(cls,
                 shard file or remove it after uploading. Defaults to ``False``.
             progress_bar (bool): Display TQDM progress bars for uploading output dataset files to
                 a remote location. Default to ``False``.
+            exist_ok (bool): When exist_ok = False, raise error if out already exists and has contents. Defaults to ``False``.
```
karan6181, line 77: What `out` value are you referring to ? Out can local, remote, or both. Can you please clarify that in the method args comment !

streaming/base/storage/upload.py
karan6181: Can we add a warning message for this to let users about the behavior ?

streaming/base/storage/upload.py
```diff
@@ -175,6 +179,17 @@ def upload_file(self, filename: str):
         """
         raise NotImplementedError('Override this method in your sub-class')
```
karan6181, line 180:
```suggestion
        raise NotImplementedError(f'{type(self).__name__}.upload_file is not implemented')
```

streaming/base/storage/upload.py
```diff
@@ -57,6 +57,7 @@ def get(cls,
             out: Union[str, Tuple[str, str]],
             keep_local: bool = False,
             progress_bar: bool = False,
+            exist_ok: bool = False,
```
karan6181, line 60: Can you put `exist_ok` as the last argument ? Otherwise, [this](https://github.com/mosaicml/streaming/blob/main/streaming/base/format/base/writer.py#L119) line will fail.

streaming/base/storage/upload.py
```diff
@@ -272,6 +289,29 @@ def check_bucket_exists(self, remote: str):
                               f'or check the bucket permission.',)
             raise error
 
+    def list_objects(self, prefix: Optional[str] = None) -> Optional[List[str]]:
+        """List all objects in the S3 object store with the given prefix.
+
+        Args:
+            prefix (Optional[str], optional): The prefix to search for. Defaults to None.
```
karan6181, line 296:
```suggestion
            prefix (Optional[str], optional): The prefix to search for. Defaults to ``None``.
```

streaming/base/storage/upload.py
```diff
@@ -388,6 +430,36 @@ def check_bucket_exists(self, remote: str) -> None:
         elif self.authentication == GCSAuthentication.SERVICE_ACCOUNT:
             self.gcs_client.get_bucket(bucket_name)
 
+    def list_objects(self, prefix: Optional[str] = None) -> Optional[List[str]]:
+        """List all objects in the GCS object store with the given prefix.
+
+        Args:
+            prefix (Optional[str], optional): The prefix to search for. Defaults to None.
```
karan6181, line 437:
```suggestion
            prefix (Optional[str], optional): The prefix to search for. Defaults to ``None``.
```

streaming/base/storage/upload.py
```diff
@@ -478,6 +552,39 @@ def check_bucket_exists(self, remote: str):
                               f'Check the bucket permission or create the bucket.',)
             raise error
 
+    def list_objects(self, prefix: Optional[str] = None) -> Optional[List[str]]:
+        """List all objects in the OCI object store with the given prefix.
+
+        Args:
+            prefix (Optional[str], optional): The prefix to search for. Defaults to None.
```
karan6181, line 559:
```suggestion
            prefix (Optional[str], optional): The prefix to search for. Defaults to ``None``.
```

streaming/base/storage/upload.py
```diff
@@ -841,3 +960,21 @@ def _upload_file():
...
+
+        Returns:
+            List[str]: A list of object names that match the prefix.
+        """
+        if prefix is None:
+            prefix = '.'
+
+        ans = []
```
karan6181, line 976: nit: `ans` -> `files_path` or something better !!

streaming/base/util.py
```diff
@@ -209,6 +215,120 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
             f'`pip install \'mosaicml-streaming[{package_name}]\'`.'
 
 
+def do_merge_index(index_file_urls: List[Any],
+                   out: Union[str, Tuple[str, str]],
+                   keep_local: bool = True,
+                   download_timeout: int = 60) -> None:
+    """Merge index.json from a list of index.json. Write to `out`, overwriting if exists.
```
karan6181, line 222: Can you please add more details ? something like: `Merge index.json from a list of subset of dataset's index.json to create a one joined dataset`. You can come up with better wordings lol

streaming/base/converters/dataframe_to_mds.py
snarayan21, line 162: there's duplicate naming between `mds_path` defined in this `write_mds` function and the returned `mds_path` from the larger `dataframeToMDS` function. could you change the naming so this is more clear?

streaming/base/util.py
```diff
@@ -209,6 +215,120 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+
+    Args:
+        index_file_urls (Union[str, Tuple[str,str]]): index.json from all the partitions
+            each element can take the form of a single path string or a tuple string.
+
+            The pattern of index_file_urls and corresponding reaction is one of:
+            1. All urls are str (local). All urls are accessible locally -> no download
+            2. All urls are tuple (local, remote). All urls are accessible locally -> no download
+            3. All urls are tuple (local, remote). At least one url is not accessible locally -> download all
```
snarayan21, line 231: could you clarify why we need to download all index files if just one url is not accessible locally?

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -28,12 +31,17 @@ def manual_integration_dir() -> Any:
...
+        os.environ.pop('AWS_SECURITY_TOKEN', None)
+        os.environ.pop('AWS_SESSION_TOKEN', None)
+        os.environ['AWS_PROFILE'] = 'temporary'
 
     tmp_dir = mkdtemp()
 
     def _method(cloud_prefix: str = 'gs://') -> Tuple[str, str]:
         mock_local_dir = tmp_dir  # mkdtemp()
-        mock_remote_dir = os.path.join(cloud_prefix, MY_BUCKET, MY_PREFIX)
+        mock_remote_dir = os.path.join(cloud_prefix, MY_BUCKET[cloud_prefix], MY_PREFIX)
```
snarayan21, line 44: are we going to test all cloud providers or just gcp and s3?

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -111,23 +119,30 @@ def test_end_to_end_conversion_local_nocolumns(self, dataframe: Any, keep_local:
...
+                    sub_dir = os.path.join(out, d)
+                    if os.path.isdir(sub_dir):
+                        shards = json.load(open(os.path.join(sub_dir, 'index.json'),
+                                                'r'))['shards']
+                        if shards:
+                            nsamples += shards[0]['samples']
+                assert (nsamples == sum([a['samples'] for a in mgi['shards']]))
+            if not keep_local:
+                assert (not os.path.exists(os.path.join(
+                    out, 'index.json'))), 'merged index.json is found even keep_local = False'
```
snarayan21, line 145:
```suggestion
                    out, 'index.json'))), 'merged index.json is found even though keep_local = False'
```

tests/test_util.py
```diff
@@ -107,6 +156,143 @@ def test_clean_stale_shared_memory():
...
+            if expected_n_shard_files == -1:
+                expected_n_shard_files = get_expected(cu.local)
+
+        if not keep_local:
+            assert not os.path.exists(os.path.join(cu.local, 'index.json'))
+            return
+
+        assert os.path.exists(local_merged_index_path)
+        merged_index = json.load(open(local_merged_index_path, 'r'))
+        n_shard_files = len({b['raw_data']['basename'] for b in merged_index['shards']})
```
snarayan21, line 200: can this just be `len(merged_index['shards')`?

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -111,23 +119,30 @@ def test_end_to_end_conversion_local_nocolumns(self, dataframe: Any, keep_local:
...
-            nsamples = 0
-            for d in os.listdir(out):
-                sub_dir = os.path.join(out, d)
-                if os.path.isdir(sub_dir):
-                    shards = json.load(open(os.path.join(sub_dir, 'index.json'), 'r'))['shards']
-                    if shards:
-                        nsamples += shards[0]['samples']
-            assert (nsamples == sum([a['samples'] for a in mgi['shards']]))
+            if keep_local:
+                assert (os.path.exists(os.path.join(out,
```
knighton, line 131: asserts don't need parens (here and elsewhere)

tests/resources/naive_MDSdataset/25/index.json
knighton, line 1: could we call this dir `naive_mds_dataset`, mixing underscores and caps gives me the heebie jeebies

streaming/base/util.py
karan6181: Yeah good point. Can we just download if it is not present locally?

tests/base/converters/test_dataframe_to_mds.py
karan6181: I think the plan is to add S3, GCS, and OCI in phase 1 and merge this PR. In phase 2, add a support for other cloud providers as well.

karan6181 (commented):
Why do we need to commit `tests/resources/naive_MDSdataset/` directory? Can we create the sub-directories and shard files on the fly in the test like how we do in other tests ?

tests/resources/naive_MDSdataset/25/index.json
XiaohanZhangCMU: removed the dataset. creating test mds dataset on the fly.

tests/base/converters/test_dataframe_to_mds.py
XiaohanZhangCMU: removed

streaming/base/util.py
XiaohanZhangCMU: This is to keep things simple so I don't mess up the folder structure local-remote. so it's either "no download" or "download them all".

streaming/base/storage/upload.py
XiaohanZhangCMU: Added.

streaming/base/util.py
XiaohanZhangCMU: This won't work.

streaming/base/util.py
```diff
@@ -196,3 +204,138 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+            'version': 2,
+            'shards': shards,
+        }
+        merged_index_path = os.path.join(temp_root, index_basename)
+        with open(merged_index_path, 'w') as outfile:
+            json.dump(obj, outfile)
+
+        # Upload merged index to remote if out has remote part
+        # Otherwise, move it from temp root to out location
+        shutil.move(merged_index_path, cu.local)
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Please double check this @karan6181

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: Add. It is for cases when out is local-only.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: yes.

streaming/base/storage/upload.py
XiaohanZhangCMU: changed.

streaming/base/storage/upload.py
XiaohanZhangCMU: changed (and a few other places)

streaming/base/storage/upload.py
XiaohanZhangCMU: Thanks. didn't realize that! changed.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: They are actually the same variables picked to the executor.

streaming/base/util.py
XiaohanZhangCMU: rewrite to allow "only download missing ones to local".

tests/base/converters/test_dataframe_to_mds.py
XiaohanZhangCMU: dataframeToMDS, merge_index are both validated with s3, oci, gcs as manual integration tests.

streaming/base/storage/upload.py
XiaohanZhangCMU: Done.

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+
+    Args:
+        out (Union[str, Tuple[str,str]]): folder that contain MDS partitions.
+            :a local directory, merge index happens locally
+            :a remote directory, download all the sub-directories index.json in a temporary sub-directories, merge locally, and then upload it to out location
+            :a (local_dir, remote_dir), check if sub-directories index.json file present locally
+                If yes, then merge locally and upload to remote_dir .
+                If not, download all the sub-directories index.json from remote to local , merge locally, and upload to remote_dir .
+        keep_local (bool): Keep local copy of the merged index file. Defaults to ``True``
+        download_timeout (int): The allowed time for downloading each json file. Defaults to 60s.
```
karan6181, line 343: There is no `download_timeout`

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+        if not keep_local:
+            shutil.rmtree(cu.local, ignore_errors=True)
+
+
+def _merge_index_from_root(out: Union[str, Tuple[str, str]], keep_local: bool = True) -> None:
+    """Merge index.json given the root of MDS dataset. Write merged index to the root folder.
+
+    Args:
+        out (Union[str, Tuple[str,str]]): folder that contain MDS partitions.
+            :a local directory, merge index happens locally
```
karan6181, line 337:
```suggestion
            :A local directory, merge index happens locally
```

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+            :a remote directory, download all the sub-directories index.json in a temporary sub-directories, merge locally, and then upload it to out location
```
karan6181, line 338:
```suggestion
            : A remote directory, download all the sub-directories index.json in a temporary sub-directories, merge locally, and then upload it to out location
```

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+            :a (local_dir, remote_dir), check if sub-directories index.json file present locally
```
karan6181, line 339:
```suggestion
            : A (local_dir, remote_dir), check if the sub-directories index.json file is present locally
```

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+                If yes, then merge locally and upload to remote_dir .
+                If not, download all the sub-directories index.json from remote to local , merge locally, and upload to remote_dir .
+        keep_local (bool): Keep local copy of the merged index file. Defaults to ``True``
+        download_timeout (int): The allowed time for downloading each json file. Defaults to 60s.
+    """
+    from streaming.base.storage.upload import CloudUploader
+
+    def not_merged_index(index_file_path: str, out: str):
+        """Check if index_file_path is the merged index at folder out."""
```
karan6181, line 348: Nit: Method comments with args is missing.

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+
+    if not out:
+        logger.warning('No MDS dataset folder specified, no index merged')
+        return
+
+    cu = CloudUploader.get(out, exist_ok=True, keep_local=True)
+
+    local_index_files = []
+    cl = CloudUploader.get(cu.local, exist_ok=True, keep_local=True)
+    for o in cl.list_objects():
```
karan6181, line 360: better wording than `o`?

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+        if o.endswith('.json') and not_merged_index(o, cu.local):
+            local_index_files.append(o)
+
+    if cu.remote:
+        obj = urllib.parse.urlparse(cu.remote)
+        remote_index_files = []
+        for o in cu.list_objects():
```
karan6181, line 367: Better wording than `o`?

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -17,7 +17,11 @@
 from streaming.base.converters import dataframeToMDS
 
 MY_PREFIX = 'train'
-MY_BUCKET = 'mosaicml-composer-tests'
+MY_BUCKET = {
+    'gs://': 'mosaicml-composer-tests',
```
karan6181, line 21: Please use some random names for buckets, for example, `testing-bucket`. The same applies to below two lines.

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -27,13 +31,19 @@
 def manual_integration_dir() -> Any:
     """Creates a temporary directory and then deletes it when the calling function is done."""
     if MANUAL_INTEGRATION_TEST:
-        os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = 'path/to/gooogle_api_credential.json'
+        os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = os.path.join(
+            os.environ['HOME'], '.mosaic/mosaicml-research-gcs.json')
```
karan6181, line 35: Why did this path change? CI won't have access to this path.

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -27,13 +31,19 @@
...
+        os.environ.pop('AWS_ACCESS_KEY_ID', None)
```
karan6181, line 36: Can we use `aws_credentials()` from conftest?

tests/test_list_objects.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2023 MosaicML Streaming authors
```
karan6181, line 1: Can you add the below test in the `test_upload.py` file instead of creating a new file?

streaming/base/storage/upload.py
```diff
@@ -388,6 +440,37 @@ def check_bucket_exists(self, remote: str) -> None:
...
+        if self.authentication == GCSAuthentication.HMAC:
+            prefix = os.path.join(str(obj.path).lstrip('/'), prefix)
+            paginator = self.gcs_client.get_paginator('list_objects_v2')
+            pages = paginator.paginate(Bucket=bucket_name, Prefix=prefix)
+            try:
+                return [obj['Key'] for page in pages for obj in page['Contents']]
+            except KeyError:
+                return []
+        elif self.authentication == GCSAuthentication.SERVICE_ACCOUNT:
+            #from google.cloud.storage import Blob, Bucket
```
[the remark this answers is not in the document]
XiaohanZhangCMU: removed.

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+            1. All urls are str (local). All urls are accessible locally -> no download
+            2. All urls are tuple (local, remote). All urls are accessible locally -> no download
+            3. All urls are tuple (local, remote). At least one url is not accessible locally -> download all
+            4. All urls are str (remote) -> download all
+
+        out (Union[str, Tuple[str, str]]): path to put the merged index file
+        keep_local (bool): Keep local copy of the merged index file. Defaults to ``True``
+        download_timeout (int): The allowed time for downloading each json file. Defaults to 60s.
+    """
+    from streaming.base.storage.download import download_file
```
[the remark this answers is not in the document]
XiaohanZhangCMU: yes I tried, but it did not work :(

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+                    f'Check data availability! local index {url[0]} is not accessible. remote index {url[1]} does not have a valid url format'
+                )
+            dest = os.path.join(temp_root, path.lstrip('/'))
+
+            try:
+                download_file(src, dest, download_timeout)
+            except Exception as ex:
+                raise RuntimeError(f'Failed to download index.json: {src} to {dest}') from ex
+
+            if not (os.path.exists(dest)):
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Fixed.

streaming/base/util.py
XiaohanZhangCMU: removed.

tests/base/converters/test_dataframe_to_mds.py
XiaohanZhangCMU: this is for local manual testing only.

tests/base/converters/test_dataframe_to_mds.py
XiaohanZhangCMU: This is not mocking the credentials but setting up the actual s3 access. using aws_credentials() does not work.

tests/test_list_objects.py
XiaohanZhangCMU: merged to test_upload.py.

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+            if isinstance(url, tuple):
+                src = url[0] if os.path.exists(url[0]) else url[1]
+            else:
+                src = url
+
+            obj = urllib.parse.urlparse(src)
+            scheme, bucket, path = obj.scheme, obj.netloc, obj.path
+            if scheme == '' and bucket == '' and path == '':
+                raise FileNotFoundError(
+                    f'Check data availability! local index {url[0]} is not accessible. remote index {url[1]} does not have a valid url format'
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Done.

streaming/base/util.py
```diff
@@ -209,6 +215,173 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+def merge_index(*args, **kwargs):  # pyright: ignore
+    if isinstance(args[0], list) and len(args) + len(kwargs) in [2, 3, 4]:
+        return _merge_index_from_list(*args, **kwargs)
+    elif (isinstance(args[0], str) or
+          isinstance(args[0], tuple)) and len(args) + len(kwargs) in [1, 2]:
+        return _merge_index_from_root(*args, **kwargs)
+    raise ValueError(f'Invalid arguments to merge_index: {args}, {kwargs}')
+
+
+def _merge_index_from_list(index_file_urls: List[Any],
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Make it List[Union[str, Tuple[str,str]]].

tests/base/converters/test_dataframe_to_mds.py
XiaohanZhangCMU: Done.

.github/workflows/pytest.yaml
```diff
@@ -47,10 +47,11 @@ jobs:
...
         run: |
           set -ex
-          pytest --splits 7 --group 1 --cov-fail-under=10
-          pytest --splits 7 --group 2 --cov-fail-under=10
-          pytest --splits 7 --group 3 --cov-fail-under=10
-          pytest --splits 7 --group 4 --cov-fail-under=10
-          pytest --splits 7 --group 5 --cov-fail-under=10
-          pytest --splits 7 --group 6 --cov-fail-under=10
-          pytest --splits 7 --group 7 --cov-fail-under=10
+          pytest --splits 8 --group 1 --cov-fail-under=10
```
snarayan21, line 50: ah...the classic pytest split incremental increase....hahaha

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -201,12 +160,12 @@ def write_mds(iterator: Iterable):
         else:
             raise RuntimeError('TaskContext.get() returns None')
 
-        if isinstance(mds_path, str):  # local
-            output = os.path.join(mds_path, f'{id}')
-            out_file_path = output
+        if mds_path[1] == '':  # only local
```
snarayan21, line 163: maybe i'm missing something obvious but where does the `mds_path` variable come from?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -201,12 +160,12 @@ def write_mds(iterator: Iterable):
         else:
```
snarayan21, line 160: could you add a short comment or docstring under the `write_mds` function describing what it's doing and input args? Would help clarify things

streaming/base/converters/dataframe_to_mds.py
snarayan21: it's from the `mds_path` variable in the larger `dataframeToMDS` function right?

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -119,47 +119,6 @@ def map_spark_dtype(spark_data_type: Any) -> str:
...
-    if isinstance(mds_path, str):
-        mds_index = os.path.join(mds_path, get_index_basename())
-    else:
-        mds_index = os.path.join(mds_path[0], get_index_basename())
-
-    with open(mds_index, 'w') as out:
-        json.dump(obj, out)
-
-
 def dataframeToMDS(dataframe: DataFrame,
```
snarayan21, line 122: nit: function naming convention with underscores instead of camelcase?

streaming/base/util.py
snarayan21: thank you!!

streaming/base/util.py
```diff
@@ -209,6 +215,179 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+                download_file(src, dest, download_timeout)
+            except Exception as ex:
+                raise RuntimeError(f'Failed to download index.json: {src} to {dest}') from ex
+
+            if not os.path.exists(dest):
+                raise FileNotFoundError(f'Index file {dest} does not exist or not accessible.')
+
+            partitions.append(dest)
+
+        # merge index files into shards
```
snarayan21, line 298:
```suggestion
        # merge shards from all index files
```

snarayan21: iiuc this is what this section is doing right? just wanted it to be clear for someone else going thru the function

streaming/base/util.py
```diff
@@ -209,6 +215,179 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+        if file.endswith('.json') and not_merged_index(file, cu.local):
+            local_index_files.append(file)
+
+    if cu.remote:
+        obj = urllib.parse.urlparse(cu.remote)
+        remote_index_files = []
+        for file in cu.list_objects():
+            if file.endswith(get_index_basename()) and not_merged_index(file, cu.remote):
+                remote_index_files.append(obj.scheme + '://' + os.path.join(obj.netloc, file))
+        if len(local_index_files) == len(remote_index_files):
```
snarayan21, line 376: curious, what's the case when local index files would be not equal to remote index files?

snarayan21: is this if someone hasn't specified local?

snarayan21 (commented):
some minor comments, other than that lgtm!

tests/test_upload.py
```diff
@@ -7,13 +7,32 @@
...
                                            GCSAuthentication, GCSUploader, LocalUploader,
                                            S3Uploader)
-from tests.conftest import R2_URL
+from tests.conftest import MY_BUCKET, R2_URL
+
+MY_PREFIX = 'train'
+
+
+@pytest.fixture(scope='function')
+def remote_local_dir() -> Any:
```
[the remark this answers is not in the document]
XiaohanZhangCMU: that common function does not accept cloud prefix so I created a new one.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: Yes, it is from the external function.

streaming/base/util.py
XiaohanZhangCMU: Yes, good catch! Changed.

streaming/base/util.py
XiaohanZhangCMU: this is when local files have 1 or more missing files or not accessible by the process running merge_index. Example is: when out = (/tmp/xxx, s3://xxxx) is specified in DBR notebook, at merging, the spark driver cannot access workers' directory.

streaming/base/converters/dataframe_to_mds.py
XiaohanZhangCMU: Added.

streaming/base/util.py
XiaohanZhangCMU: Good catch. Fixed.

tests/test_util.py
```diff
@@ -107,6 +123,160 @@ def test_clean_stale_shared_memory():
...
+def integrity_check(out: Union[str, Tuple[str, str]],
+                    keep_local: bool,
+                    expected_n_shard_files: int = -1):
+    """ Check if merged_index file has integrity
+        If merged_index is a cloud url, first download it to a temp local file.
+
+    Args:
+        out (Union[str, Tuple[str,str]]): folder that merged index.json resides
+        keep_local: whether to check local file
+        expected_n_shard_files (int): If -1, find the number in out with get_expected()
```
karan6181, line 135:
```suggestion
        expected_n_shard_files (int): If -1, find the number in `out` with get_expected()
```

tests/test_util.py
```diff
@@ -107,6 +123,160 @@ def test_clean_stale_shared_memory():
...
+        with tempfile.TemporaryDirectory() as a_temporary_folder:
+            index_file_urls = [(os.path.join(a_temporary_folder, os.path.basename(s)), '')
+                               for s in local_index_files]
+            with pytest.raises(FileNotFoundError, match=f'.*Check data availability!.*'):
+                merge_index(index_file_urls, out, keep_local=keep_local)
+            return
+
+    if index_file_urls_pattern == 3:
+        remote_index_files = [
+            os.path.join(scheme, MY_BUCKET[scheme], MY_PREFIX, os.path.basename(o))
```
karan6181, line 238: Is this testing remote cloud path?

tests/test_util.py
```diff
@@ -1,15 +1,31 @@
...
 from streaming.base.constant import RESUME
 from streaming.base.shared.prefix import _get_path
+from streaming.base.storage.download import download_file
+from streaming.base.storage.upload import CloudUploader
 from streaming.base.util import (bytes_to_int, clean_stale_shared_memory, get_list_arg,
-                                 number_abbrev_to_int, retry)
+                                 merge_index, number_abbrev_to_int, retry)
+
+MY_PREFIX = 'train_' + str(time.time())
+MY_BUCKET = {
```
karan6181, line 22: Can we have this as `@pytest.mark.parametrize`? I think only `s3://` is getting used.

streaming/base/util.py
```diff
@@ -209,6 +215,179 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
             f'`pip install \'mosaicml-streaming[{package_name}]\'`.'
 
 
+def merge_index(*args, **kwargs):  # pyright: ignore
```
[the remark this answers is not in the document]
karan6181: Can you also add method comments?

streaming/base/util.py
```diff
@@ -209,6 +215,179 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+    for file in cl.list_objects():
+        if file.endswith('.json') and not_merged_index(file, cu.local):
+            local_index_files.append(file)
+
+    if cu.remote:
+        obj = urllib.parse.urlparse(cu.remote)
+        remote_index_files = []
+        for file in cu.list_objects():
+            if file.endswith(get_index_basename()) and not_merged_index(file, cu.remote):
+                remote_index_files.append(obj.scheme + '://' + os.path.join(obj.netloc, file))
```
karan6181, line 375: For UC Volume and DBFS, it's `:/`.

streaming/base/storage/upload.py
```diff
@@ -478,6 +562,39 @@ def check_bucket_exists(self, remote: str):
...
+                response = self.client.list_objects(namespace_name=self.namespace,
+                                                    bucket_name=bucket_name,
+                                                    prefix=prefix,
+                                                    start=next_start_with).data
+                object_names.extend([resp_obj.name for resp_obj in response.objects])
+                next_start_with = response.next_start_with
+                if not next_start_with:
+                    response_complete = True
+            return object_names
+        except Exception as _:
```
karan6181, line 595: This is too generic. Is there a specific Exception that `list_objects` returns when the file is not found? Or something like KeyError?

streaming/base/util.py
```diff
@@ -209,6 +215,179 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+
+        out (Union[str, Tuple[str, str]]): path to put the merged index file
+        keep_local (bool): Keep local copy of the merged index file. Defaults to ``True``
+        download_timeout (int): The allowed time for downloading each json file. Defaults to 60s.
+    """
+    from streaming.base.storage.download import download_file
+    from streaming.base.storage.upload import CloudUploader
+
+    if not index_file_urls or not out:
+        logger.warning('Need to specify both index_file_urls and out. No index merged')
```
karan6181, line 251:
```suggestion
        logger.warning('Need to specify both `index_file_urls` and `out`. No index merged')
```

streaming/base/util.py
```diff
@@ -209,6 +215,179 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+                src = url[0] if os.path.exists(url[0]) else url[1]
+            else:
+                src = url
+
+            obj = urllib.parse.urlparse(src)
+            scheme, bucket, path = obj.scheme, obj.netloc, obj.path
+            if scheme == '' and bucket == '' and path == '':
+                raise FileNotFoundError(
+                    f'Check data availability! local index {url[0]} is not accessible.' +
+                    'remote index {url[1]} does not have a valid url format')
```
karan6181, line 285:
```suggestion
                    f'remote index {url[1]} does not have a valid url format')
```

streaming/base/util.py
```diff
@@ -209,6 +215,179 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+    """Merge index.json from a list of index files of MDS directories to create joined index.
+
+    Args:
+        index_file_urls (Union[str, Tuple[str,str]]): index.json from all the partitions
+            each element can take the form of a single path string or a tuple string.
+
+            The pattern of index_file_urls and corresponding reaction is one of:
+            1. All urls are str (local). All urls are accessible locally -> no download
+            2. All urls are tuple (local, remote). All urls are accessible locally -> no download
+            3. All urls are tuple (local, remote). Download url thtat is not accessible locally
```
karan6181, line 240:
```suggestion
            3. All URLs are tuple (local, remote). Download URL that is not accessible locally
```

streaming/base/storage/upload.py
XiaohanZhangCMU: Added two specific errors.

tests/test_util.py
XiaohanZhangCMU: yes, but no downloading, so still a local test.

tests/test_upload.py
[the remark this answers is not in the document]
XiaohanZhangCMU: To change the old method, the calling method will be changed from local_remote_dir to local_remote_dir(). 
The old method is being used by many files. Do we really want to make change to that many files?

streaming/base/util.py
XiaohanZhangCMU: changed. 
btw, UC Volume does not support list_objects yet. We should keep an eye on that when they make the endpoint available.

tests/test_upload.py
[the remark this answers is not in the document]
XiaohanZhangCMU: Many files are affected. Will change those in a separate PR.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -119,52 +119,26 @@ def map_spark_dtype(spark_data_type: Any) -> str:
...
                    merge_index: bool = True,
                    mds_kwargs: Optional[Dict[str, Any]] = None,
                    udf_iterable: Optional[Callable] = None,
                    udf_kwargs: Optional[Dict[str, Any]] = None) -> Tuple[Any, int]:
+    """Deprecated API Signature.
+
+    To be replaced by dataframe_to_mds
+    """
+    logger.warning(
+        'This signature is deprecated. Use dataframe_to_mds with the same arguments going forward.'
```
karan6181, line 132: nit: how about `The DataframeToMDS signature has been deprecated and will be removed in Streaming 0.8. Use dataframe_to_mds with the same arguments going forward`

tests/base/converters/test_dataframe_to_mds.py
```diff
@@ -3,55 +3,26 @@
...
 from pyspark.sql.functions import col
 from pyspark.sql.types import DecimalType, IntegerType, StringType, StructField, StructType
 
-from streaming import MDSWriter
 from streaming.base.converters import dataframeToMDS
 
 MY_PREFIX = 'train'
-MY_BUCKET = 'mosaicml-composer-tests'
-MANUAL_INTEGRATION_TEST = False
+MY_BUCKET = {
```
karan6181, line 17: It looks like `MY_BUCKET` and `MY_PREFIX` is not used in this file. Can we please remove it?

streaming/base/util.py
```diff
@@ -209,6 +215,186 @@ def get_import_exception_message(package_name: str, extra_deps: str) -> str:
...
+                           keep_local: bool = True,
+                           download_timeout: int = 60) -> None:
+    """Merge index.json from a list of index files of MDS directories to create joined index.
+
+    Args:
+        index_file_urls (Union[str, Tuple[str,str]]): index.json from all the partitions
+            each element can take the form of a single path string or a tuple string.
+
+            The pattern of index_file_urls and corresponding reaction is one of:
+            1. All URLS are str (local). All URLS are accessible locally -> no download
```
karan6181, line 239:
```suggestion
            1. All URLs are str (local). All URLs are accessible locally -> no download
```

tests/test_util.py
```diff
@@ -107,6 +123,160 @@ def test_clean_stale_shared_memory():
...
+        assert n_shard_files == expected_n_shard_files, f'expected {expected_n_shard_files} shard files but got {n_shard_files}'
+
+
+@pytest.mark.parametrize('index_file_urls_pattern', [1, 2, 3])
+@pytest.mark.parametrize('keep_local', [True, False])
+@pytest.mark.parametrize('scheme', ['gs://', 's3://', 'oci://'])
+def test_merge_index_from_list_local(local_remote_dir: Tuple[str, str], keep_local: bool,
+                                     index_file_urls_pattern: int, scheme: str):
+    """Validate the final merge index json for following patterns of index_file_urls:
+        1. All urls are str (local). All urls are accessible locally -> no download
```
karan6181, line 179:
```suggestion
        1. All URLs are str (local). All URLs are accessible locally -> no download
```

streaming/base/util.py
karan6181: `s` in lower case.

tests/base/converters/test_dataframe_to_mds.py
XiaohanZhangCMU: Done.

karan6181 (approved):

PR #456, opened 2023-10-03
Update integration test to include sample order comparison

karan6181:
## Description of changes:
- As part of testing streaming feature regularly to capture any regression, I had to add the below changes to make certain test working, such as.
  - Validating number of files locally
  - Validating the epoch iteration time
  - Validating the sample orders
- More details is in PR https://github.com/mosaicml/testing/pull/43

## Merge Checklist:

### General

### Tests

### DIFF
--- a/regression/iterate_data.py
+++ b/regression/iterate_data.py
@@ -13,7 +13,8 @@ from typing import Union
 import torch
 from torch import distributed as dist
 from torch.utils.data import DataLoader
-from utils import compare_files, get_dataloader_params, get_kwargs, get_streaming_dataset_params
+from utils import (compare_sample_order, get_dataloader_params, get_kwargs,
+                   get_streaming_dataset_params)
 
 from streaming import StreamingDataset
 from streaming.base.distributed import (all_gather, barrier, get_rank, get_world_size,
@@ -113,8 +114,8 @@ def main(args: Namespace, kwargs: dict[str, str]) -> None:
         for batch in dataloader:
             if args.sample_order_file is not None and epoch == 0:
                 key = None
-                if 'id' in batch:
-                    key = 'id'
+                if 'sample' in batch:
+                    key = 'sample'
                 elif 'number' in batch:
                     key = 'number'
                 samples = [int(sample) for sample in batch[key]]
@@ -167,6 +168,6 @@ def main(args: Namespace, kwargs: dict[str, str]) -> None:
 if __name__ == '__main__':
     args, kwargs = parse_args()
     if args.cmp_sample_order:
-        compare_files(args.cmp_sample_order)
+        compare_sample_order(args.cmp_sample_order)
     else:
         main(args, kwargs)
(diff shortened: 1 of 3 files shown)


regression/iterate_data.py
```diff
@@ -104,10 +109,9 @@ def main(args: Namespace, kwargs: dict[str, str]) -> None:
     iter_time = []
     start_time = 0
     for epoch in range(args.epochs):
-        if epoch > 0:
-            start_time = time.time()
+        start_time = time.time()
         for batch in dataloader:
-            if args.sample_order_file is not None:
+            if args.sample_order_file is not None and epoch == 0:
```
snarayan21, line 114: curious why we're not writing samples to the file past epoch 0?

regression/iterate_data.py
```diff
@@ -30,12 +34,13 @@ def parse_args() -> tuple[Namespace, dict[str, str]]:
...
                       help='Number of epochs to iterate over the data')
-    args.add_argument('--validate-files', default=False, action='store_true', help='Verify files')
-    args.add_argument('--sample-order-file', type=str, help='File location to store sample order')
-    args.add_argument('--validate-iter-time',
+    args.add_argument('--validate-num-files',
                       default=False,
                       action='store_true',
-                      help='Test iter time')
+                      help='Verify number of dataset files')
+    args.add_argument('--sample-order-file', type=str, help='File location to store sample order')
```
snarayan21, line 41: when we automatically run this test, is there a default argument to set here or will CORGI handle that for us?

regression/utils.py
```diff
@@ -204,3 +205,65 @@ def delete_oci(remote_dir: str) -> None:
 
     for filenames in objects.data.objects:
         oci_client.delete_object(namespace, obj.netloc, filenames.name)
+
+
+def compare_files(files: list[str], identical: bool = False) -> None:
```
snarayan21, line 210: can we use the library difflib for this? gets the diff between two files. that's what we need right? Example here:
[](https://www.tutorialspoint.com/How-to-find-difference-between-2-files-in-Python#:~:text=We%20invoke%20the%20open(),file%20names%20in%20the%20output.)

snarayan21 (approved):
lgtm, just had some clarifying questions

regression/utils.py
karan6181: So the method `compare_files()` is not just the diff comparison, it is actually a sample order comparison between multiple files across all GPUs. For example, each line in an output file contains a sample id of `batch_size * n_gpus_across_all_nodes`  samples. For deterministic sample ordering, given same NCN and global batch_size, each gpu gets some sample ordering, so with one forward pass across all the gpus, the set of sample ids between any two runs should be same, however, the samples can be in different order (in different gpu or in different position in a batch size) in the same forward pass.

regression/iterate_data.py
karan6181: CORGI provides the file path.

regression/iterate_data.py
karan6181: Good point. I will have to change the logic in future for testing different sampling_method

PR #463, opened 2023-10-06
Add py1e warning when Shuffle block size is smaller than shard size

snarayan21:
## Description of changes:
Previously, when shuffle block size was less than the span size aka shard part size, the run would crash at [this line](https://github.com/mosaicml/streaming/blob/6b638518294f7338f7bb54e9bb529bfad8e5fa50/streaming/base/shuffle/py1e.py#L102) with `ValueError: high - low < 0` because the new shard range would not be defined. Instead, when this is the case, we now set the shard range extension to 0, meaning the shard range stays the same, so the shard is not shuffled with other neighboring shards at all. Because this is likely not the intended behavior, we warn the user when this happens as well. 

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/shuffle/py1e.py
+++ b/streaming/base/shuffle/py1e.py
@@ -8,6 +8,8 @@ This algorithm has more balanced downloading and a lower minimum cache limit tha
 is determined by ``shuffle_block_size``.
 """
 
+import warnings
+
 import numpy as np
 from numpy.typing import NDArray
 
@@ -63,6 +65,7 @@ def get_shuffle_py1e(shard_sizes: NDArray[np.int64],
     # Populate the global sample ID mapping, shuffling within each span.
     ids = np.empty(num_samples, np.int64)
     offset = 0
+    warn_user = False
     # Iterate through each canonical node's spans.
     # We don't want samples crossing canonical node boundaries.
     for cn_begin, cn_end in super_spans:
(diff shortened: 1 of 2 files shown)


streaming/base/shuffle/py1e.py
```diff
@@ -107,6 +116,12 @@ def get_shuffle_py1e(shard_sizes: NDArray[np.int64],
             # Update sample offset for the next shard.
             cn_sample_offset += span_size
 
+        # If warn_user is true, this means the block size for shifts was smaller than a span size.
+        # This will result in no shuffling being done on that span aka shard part, so warn user.
+        if warn_user:
+            warnings.warn('Shuffle block size was smaller than shard size for some shards. This \
```
karan6181, line 122: Is there a way to validate this behavior using a unit test?

snarayan21: Oh yeah let me add that!

snarayan21: added unit test! I didn't even know pytest could detect warnings...

tests/test_streaming.py
```diff
@@ -510,6 +510,37 @@ def run_one_iter(local: str, remote: str, seed: int) -> None:
...
+                               batch_size=batch_size,
+                               num_canonical_nodes=1,
+                               shuffle_seed=seed,
+                               shuffle_algo='py1e',
+                               shuffle_block_size=shuffle_block_size)
+    dataloader = DataLoader(dataset=dataset, batch_size=batch_size)
+
+    with pytest.warns(UserWarning, match=f'Shuffle block size was smaller than shard size*'):
+        for _ in dataloader:
+            print('yo')
```
karan6181, line 541:
```suggestion
            pass
```

PR #494, opened 2023-11-02
Move examples out, merge base/ upward

knighton:
Context: breaking PR 488 into around ten sub-PRs, of which this is the first.

When streaming repo started, it was laid out like

```
streaming/
    base/
    text/
    vision/
    (future: audio/?)
    (future: multimodal/?)
```

The big picture was let’s, among other goals, exhaustively implement all the well-known datasets out there, starting with at least mirroring the repertoire of torchvision and other torchmodality datasets but doing it all in MDS.

I don’t think we have a good picture of how streaming is used in totality, but it seems like the pattern is that people tend to own the data prep/serialization and any custom StreamingDataset subclass work/deserialization entirely themselves. They want to train on their own data.

The datasets we built into Streaming as “batteries included” are not exhaustive, getting even good coverage is impossible, and suspect they are not being used directly generally speaking, but instead as examples or what you can do with streaming.

On that basis,
- We are not in the “implementing all the datasets” business, we have users now to drive what we are doing, that work belongs to them
- However, all your base continues to belong to us
- Our SD subclasses are useful as examples. Jettison every streaming subdir not named base/ using them to populate top level examples/ dir, how about one examples/ subdir per end-to-end example
- Rename streaming/base/ -> streaming/
- Some of scripts/ should move to examples/ too
- Public imports that we all use, at the streaming/ level, continue to work unchanged

Overview:

```
    examples/ -> notebooks/
    streaming/{multimodal, vision, text}/ -> examples/
    scripts/ -> mostly benchmarks/, except for examples/
    streaming/base/format/base/ -> streaming/format/
    streaming/base/ -> streaming/
```

Old code org:

```
.
├── docs
│   └── source
│       ├── examples -> ../../examples
│       ├── fundamentals
│       ├── getting_started
│       ├── how_to_guides
│       ├── _static
│       │   ├── css
│       │   ├── images
│       │   └── js
│       └── _templates
│           └── sidebar
├── examples
├── regression
├── scripts
│   ├── compression
│   ├── epoch
│   ├── hashing
│   ├── partition
│   ├── samples
│   ├── serialization
│   ├── shuffle
│   └── webvid
├── simulation
│   ├── core
│   ├── interfaces
│   └── testing
├── streaming
│   ├── base
│   │   ├── batching
│   │   ├── converters
│   │   ├── format
│   │   │   ├── base
│   │   │   ├── json
│   │   │   ├── mds
│   │   │   └── xsv
│   │   ├── partition
│   │   ├── shared
│   │   ├── shuffle
│   │   └── storage
│   ├── multimodal
│   │   └── convert
│   │       ├── laion
│   │       │   └── laion400m
│   │       └── webvid
│   ├── text
│   │   └── convert
│   │       └── enwiki
│   │           ├── mds
│   │           └── tfrecord
│   └── vision
│       └── convert
└── tests
    ├── base
    │   └── converters
    └── common
```

New code org:

```
.
├── benchmarks
│   ├── compression
│   ├── epoch
│   ├── hashing
│   ├── partition
│   ├── samples
│   ├── serialization
│   └── shuffle
├── docs
│   └── source
│       ├── examples -> ../../examples
│       ├── fundamentals
│       ├── getting_started
│       ├── how_to_guides
│       ├── _static
│       │   ├── css
│       │   ├── images
│       │   └── js
│       └── _templates
│           └── sidebar
├── examples
│   ├── multimodal
│   │   ├── laion400m
│   │   └── webvid
│   │       ├── scripts
│   │       └── write
│   ├── text
│   │   ├── c4
│   │   ├── enwiki_tok
│   │   │   ├── mds
│   │   │   └── tfrecord
│   │   ├── enwiki_txt
│   │   └── pile
│   └── vision
│       ├── ade20k
│       ├── cifar10
│       ├── coco
│       └── imagenet
├── notebooks
├── regression
├── simulation
│   ├── core
│   ├── interfaces
│   └── testing
├── streaming
│   ├── batching
│   ├── converters
│   ├── format
│   │   ├── json
│   │   ├── mds
│   │   └── xsv
│   ├── partition
│   ├── shared
│   ├── shuffle
│   └── storage
└── tests
    ├── base
    │   └── converters
    └── common
```


streaming/shared/prefix.py
```diff
@@ -128,7 +128,7 @@ def _check_and_find(streams_local: List[str], streams_remote: List[Union[str, No
                             f'Reused local directory: {streams_local} vs ' +
                             f'{their_locals}. Provide a different one. If using ' +
                             f'a unique local directory, try deleting the local directory and ' +
-                            f'call `streaming.base.util.clean_stale_shared_memory()` only once ' +
+                            f'call `streaming.util.clean_stale_shared_memory()` only once ' +
```
karan6181, line 131: This is a breaking change where if the user calls `streaming.base.util.clean_stale_shared_memory()`, it will fail since the `base` module no longer exists, and this API is being used outside. Is there a way where the user can call `streaming.base.util.clean_stale_shared_memory()` but will see a deprecation warning stating that the API path has been changed to `streaming.util.clean_stale_shared_memory()`?

pyproject.toml
```diff
@@ -17,9 +17,10 @@ include = [
 ]
 exclude = [
     "build/**",
+    "docs/source/conf.py",
+    "docs/source/examples/text/enwiki_tok/**",
```
karan6181, line 21: There is no such path `docs/source/examples/text/enwiki_tok/**`.

streaming/vision.py
```diff
@@ -0,0 +1,154 @@
+# Copyright 2023 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
```
karan6181, line 3: Feels like this file does fit well here!

karan6181 (commented):
Looks great. Few minor comments:
- Can you please update the PR description with details on this PR is about?
- The `docs/source/how_to_guides/dataset_conversion_to_mds_format.md` line 22 still reference `../../../streaming/base/converters/README.md`. Can you please change that?

pyproject.toml
knighton: Ah, `docs/source/examples` is a softlink to `examples/`.

When I remove that line and run `pre-commit run --all-files`, I get a ton of spew that we want to not mess with originating in `docs/source/examples/text/enwiki_tok/`.

streaming/shared/prefix.py
knighton: In the old code, this was not exported from either `streaming/__init__.py` or `streaming/base/__init__.py`, so I thought this is not covered by the no breaking changes rule?

In any case, I can mirror the old file tree, where each `base/` file just does ye olde `redirect(streaming/base/<path> -> streaming/<path>)`, for everything since it would be hard to justify only doing that one method?

examples/vision/ade20k/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2023 MosaicML Streaming authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""ADE20k streaming dataset example."""
```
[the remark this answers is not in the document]
knighton: nice

streaming/vision.py
snarayan21: +1, I agree. Are people using these classes? if so, can we put this back in a `streaming/vision` directory or something?

snarayan21 (commented):
just had some minor comments and agree with Karan's points, otherwise looks good

examples/text/c4/README.md
```diff
@@ -0,0 +1,7 @@
+### [C4: Colossal, Cleaned, Common Crawl dataset](https://huggingface.co/datasets/c4)
+
+1. Run the [c4.py](https://github.com/mosaicml/streaming/blob/main/streaming/text/convert/c4.py) script as shown below. The script downloads the raw format with `train` and `val` splits from HuggingFace hub and converts to StreamingDataset MDS format into their own split directories. For more advanced use cases, please see the supported arguments for [c4.py](https://github.com/mosaicml/streaming/blob/main/streaming/text/convert/c4.py) and modify as necessary.
```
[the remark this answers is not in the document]
knighton: Fixed paths

streaming/vision.py
knighton: Hmmm.

How much torchvisioning have you done? https://pytorch.org/vision/main/_modules/index.html :eyes: They're basically all `VisionDataset`s -> this is a standard PyTorch idiom and is like 1/3 as basey as `StreamingDataset`, which is basey enough for me :)

And recall the time RR had to key off some composer behavior on whether a streaming dataset was vision.

But yeah, we should only keep base classes in `streaming/`, so we need to delete `StreamingDataset` because it's not the base class, it descends from `Array` along with `LocalDataset`, to which we may add more subclasses in the future for different use cases (c.f. that user request about a lazy map-style non-streaming Streaming dataset, etc).

Also these files are pretty diverse already, I don't think a (modality).py sticks out?

I totally see moving the Streaming-specific innovation `convert_image_class_dataset()` out to a `examples/vision/base.py`, but the rest is a la [torchvision/datasets/vision.py](https://pytorch.org/vision/main/_modules/torchvision/datasets/vision.html), let's keep it in a `streaming/vision.py` I think?

```
array.py
batching/
compression.py
constant.py
converters/
dataloader.py
dataset.py
distributed.py
format/
hashing.py
__init__.py
local.py
partition/
py.typed
sampling.py
shared/
shuffle/
spanner.py
storage/
stream.py
util.py
_version.py
vision.py
world.py
```

karan6181 (changes requested):
Based on our offline discussion, adding a request change flag here to merge this PR in a separate branch rather than `main` if we have to do another patch release. Once the Delta Table changes land in, we can create a big PR to `main` and merge it. For now, we can develop all the stuff in a separate branch.

docs/source/how_to_guides/dataset_conversion_to_mds_format.md
```diff
@@ -13,34 +13,30 @@ Let's look at the steps one needs to perform to convert their raw data into an M
...
+```{include} ../../../streaming/converters/README.md
 :start-line: 2
 ```
 
 ## NLP Dataset Conversion Examples
 
-```{include} ../../../streaming/text/convert/README.md
-:start-line: 8
-```
+[examples/text/](../examples/text/)
```
karan6181, line 28: Looks like `README.md` is missing in `../examples/text/` directory! Do you know what the doc will look like after this change?

docs/source/how_to_guides/dataset_conversion_to_mds_format.md
```diff
@@ -13,34 +13,30 @@ Let's look at the steps one needs to perform to convert their raw data into an M
...
 
 ## Vision Dataset Conversion Examples
 
-```{include} ../../../streaming/vision/convert/README.md
-:start-line: 8
-```
+[examples/vision/](../examples/vision/)
```
karan6181, line 32: same here

karan6181 (commented):
Overall looks good to me. 
- Can we add a `warnings.warn('', DeprecationWarning, stacklevel=2)` in method `clean_stale_shared_memory()` stating that the import has been moved from `streaming.base.util.clean_stale_shared_memory()` to `streaming.util.clean_stale_shared_memory()` and the old import will be removed in Streaming 0.9?
- Also, going forward, you will have to run the doc locally to ensure it passes since the doc CI does not run on non-main branches. The steps is mentioned [here](https://github.com/mosaicml/streaming/blob/main/CONTRIBUTING.md#submitting-a-contribution).

examples/multimodal/webvid/scripts/bench_inside.py
```diff
@@ -8,7 +8,7 @@
 
 import numpy as np
 
-from streaming.multimodal.webvid import StreamingInsideWebVid
+from examples.multimodal.webvid.read import StreamingInsideWebVid
```
snarayan21, line 11: Can we make sure that `examples` is included when installing the `mosaicml-streaming` package, like in a new venv or something? Ran into a similar issue with the simulator and want to make sure that the reorg doesn't affect this as well.

examples/vision/cifar10/write_fake.py
```diff
@@ -7,7 +7,7 @@
 import numpy as np
 from PIL import Image
 
-from streaming.vision.convert.base import convert_image_class_dataset
+from streaming.vision import convert_image_class_dataset
```
snarayan21, line 10: as per prev review, is the function `convert_image_class_dataset` living in a separate directory or just in `vision.py`? just wanted to make sure this is in the right place

snarayan21 (commented):
just some minor comments. also, do the examples notebooks all work with the code reorg? want to make sure we got all the changes right

examples/multimodal/webvid/scripts/bench_inside.py
knighton: Oh yeah, was worrying about that. How did you handle this wrt simulator?

Tentatively, since I am repurposing `examples/`, you would think it should be safe, and `notebooks/` and `benchmarks/` might be a problem.

Following up, when I look at `setup.py`, there is this line
```
    packages=setuptools.find_packages(exclude=['tests*']),
```
which seems to suggest we're fine? What say you.

examples/vision/cifar10/write_fake.py
knighton: Thanks for pointing this out.

Argument from symmetry: since it writes sample pairs as `x` `y`, which then get accessed by `StreamingVisionDataset` as sample pairs of `x` `y`, let us have both (but it should be renamed symmetrically).

Argument from expedience: this is minor and we will still have time to fix in subsequent PRs before release, so let's move forward.

I'd like to then do a quick followup PR to rewrite that ridiculous method to this (or so):

```
def convert_vision_dataset(
    *,
    samples: Iterable[Dict[str, Any]],
    shuffle: bool = True,
    seed: int = 9176,
    out_root: str,
    split: Optional[str] = None,
    hashes: Optional[Sequence[str]] = None,
    size_limit: int = 1 << 25,
    show_progress: bool = True,
    encoding: str = 'png') -> None:
```

Finally, surely the single-letter column name thing was a bad idea too, not least because it makes it look like column name length matters to time and/or serialized size...

docs/source/how_to_guides/dataset_conversion_to_mds_format.md
knighton: Hmm. It was at the `convert/` subdir level. I went this way for, can improve in subsequent PRs.

docs/source/how_to_guides/dataset_conversion_to_mds_format.md
knighton: Hmm. It was at the `convert/` subdir level. I went this way for now, can improve in subsequent PRs.

knighton:
> Overall looks good to me.
> 
>     * Can we add a `warnings.warn('', DeprecationWarning, stacklevel=2)` in method `clean_stale_shared_memory()` stating that the import has been moved from `streaming.base.util.clean_stale_shared_memory()` to `streaming.util.clean_stale_shared_memory()` and the old import will be removed in Streaming 0.9?
> 
>     * Also, going forward, you will have to run the doc locally to ensure it passes since the doc CI does not run on non-main branches. The steps is mentioned [here](https://github.com/mosaicml/streaming/blob/main/CONTRIBUTING.md#submitting-a-contribution).

How about this?

```
-    warn(f'Please update your imports: {old_fqdn} has moved to {new_fqdn}.')
+    warn(f'Please update your imports: {old_fqdn} has moved to {new_fqdn}.',
+         DeprecationWarning,
+         stacklevel=2)
```

karan6181 (approved):
Approving the PR for now so the upcoming PR is unblocked and let's fix the doc build in the upcoming PR.

PR #530, opened 2023-12-12
Redo/generalize/tighten args shorthand

knighton:
Get more specific functionality in place to set the stage for the more interesting PRs.

Beeves I have with existing approach:
1. Miscalculates when you take decimal amounts of sufficiently large units and are unlucky in floating point. Solved by doing everything in int.
2. Sometimes people need base-1000, sometimes people need base-1024. Solved by using the standard set of suffixes where you infix an `i` when using base-1024 (`1gb` vs `1gib`). Then we can recognize both at the same time, or either.
3. Having the flexibility to accept both uppercase and lowercase or any combination thereof for units is a Tower of Babel situation. People are inevitably going to make dialects out of this flexibility. We will totes get complaints that they said `Gb` and didn't get gigabits.
4. Similarly, the flexibility to put spaces before, after, or in-between parts of the argument is TIMTOWDI and very fashionista, not pythonista, let's have a single canonical form.
5. Missed oppty to support multiple parts, then we can also do things like `1h23m45s`.

Paths:
1. `normalize_bytes` -> `normalize_dec_bytes`, `normalize_bin_bytes` -> `_normalize_nonneg_int` -> `_normalize_int` -> `_normalize_num` -> `_normalize_arg`.
2. `normalize_count` -> `_normalize_nonneg_int` -> `_normalize_int` -> `_normalize_num` -> `_normalize_arg`.
3. `normalize_duration` -> `_normalize_float` -> `_normalize_num` -> `_normalize_arg`.

Steps of the `_normalize_arg` algorithm:
1. Must be non-empty.
2. Drop commas and underscores (useful to demarcate thousands '1,337' or '1_337').
3. Must start with a digit.
4. Must alternative between numbers and units, starting with a number.
5. If just a number, that's it.
6. Pair up numbers and units: (a) special case where the implied unit is the empty string, i.e. the smallest unit; (b) if not just a number, each number must be paired with a corresponding unit.
7. Assign parts as numbers and units.
8. Each number before the last one must be integral.
9. Parse out the digits of the final number, which may be fractional.
10. Parse the digits as an integer for exact precision, no float nonsense.
11. Each unit must be known to us.
12. Each unit must be used at most once.
13. Units must be listed in descending order of size.
14. The number of any given part must not exceed the size of the next biggest part's unit. (Otherwise you would just roll its overage into the next biggest part.)
15. Collect parts, with last part being possibly scaled down to account for a decimal point.

Example of how configuration and functionality are decomposed in this PR:
```py
_count_units = _get_units(1000, ' k m b t'.split(' '))


def normalize_count(count: Union[int, str]) -> int:
    """Normalize from human-friendly count to int.

    Args:
        count (int | str): Human-friendly count.

    Returns:
        int: Integral count.
    """
    return _normalize_nonneg_int(count, _count_units)
```


tests/test_shorthand.py
```diff
@@ -0,0 +1,106 @@
...
+                                                       ('hello', ['hello']), ('', [])])
+def test_get_list_arg(text: str, expected_output: List[Optional[str]]):
+    output = get_list_arg(text)
+    assert output == expected_output
+
+
+def test_normalize_bytes():
+    input_to_expected = [
+        ('1234', 1234),
+        ('1b', 1),
```
XiaohanZhangCMU, line 21: Does an input like '1.5e9mib' work?

XiaohanZhangCMU (approved):

tests/test_shorthand.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2023 MosaicML Streaming authors
```
karan6181, line 1: Wondering if it is a good idea to move the `test_util.py` and `test_shorthand.py` under `tests/util` directory?

streaming/util/shorthand.py
```diff
@@ -1,115 +1,371 @@
...
-            try:
-                return int(float(bytes_str[0:-len(suffix)]) * units[suffix])
-            except ValueError:
-                raise ValueError(''.join([
-                    f'Unsupported value/suffix {bytes_str}. Supported suffix are ',
-                    f'{["b"] + list(units.keys())}.'
-                ]))
+    # Must be non-empty.
+    if not text:
+        raise ValueError(f'Text is empty.')
```
karan6181, line 67: Maybe a good idea to add the text in context to the facing parameter. Reading `'Text is empty` does not give me enough information on what is wrong with my param.

streaming/util/shorthand.py
```diff
@@ -1,115 +1,371 @@
...
+    part = ''.join(part)
+    parts.append(part)
+
+    # If just a number, that's it.
+    if len(parts) == 1:
+        part, = parts
+        try:
+            return to_type(part)
+        except:
+            raise ValueError(f'Simple text must be numeric, but got {part} instead (input: ' +
```
karan6181, line 110: Same here. I find it difficult to understand what `Simple text` means here from a user perspective.

Also, looks at other exception messages in the PR for improvement.

simulation/interfaces/interface_utils.py
```diff
@@ -88,7 +88,7 @@ def get_train_dataset_params(input_params: dict, old_params: Optional[dict] = No
     train_dataset_params['cache_limit'] = input_params['cache_limit']
     train_dataset_params['shuffle'] = input_params['shuffle']
     train_dataset_params['shuffle_algo'] = input_params['shuffle_algo']
-    train_dataset_params['shuffle_block_size'] = number_abbrev_to_int(
+    train_dataset_params['shuffle_block_size'] = normalize_count(
```
snarayan21, line 91: have you been able to test these changes? could you make sure these numbers are displaying and being passed as intended by running `simulator` and passing in values for these parameters?

streaming/util/shorthand.py
```diff
@@ -1,115 +1,371 @@
 # Copyright 2023 MosaicML Streaming authors
 # SPDX-License-Identifier: Apache-2.0
 
-"""Utilities for human-friendly argument shorthand."""
+"""Conversions between human-friendly string forms and int/float."""
```
snarayan21, line 4: just a suggestion -- would the [humanize library](https://pypi.org/project/humanize/) be applicable for some of these functions? Would be nice to use an external library for stuff like this, removes some burden on us as well.

snarayan21 (commented):
main things are testing simulation changes and seeing if the `humanize` package fulfills some of our needs here!

simulation/interfaces/interface_utils.py
snarayan21: If you could add a short "testing" section to the PR description that would be great as well!

simulation/core/sim_dataset.py
```diff
@@ -285,13 +280,14 @@ def __init__(self,
 
         # Check that cache limit is possible.
         if self.cache_limit:
```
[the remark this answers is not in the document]
knighton: Very nice catch

I caught/fixed the copy of this code in StreamingDataset but missed this one...

tests/test_shorthand.py
knighton: done

streaming/util/shorthand.py
knighton: How about `        raise ValueError(f'Attempting to normalize the empty strring.')`?

knighton: Er, `        raise ValueError(f'Attempted to normalize an empty string to some value.')`

streaming/util/shorthand.py
knighton: How about now...

tests/test_shorthand.py
knighton: Nope, it does not take scientific notation amounts of units, as units and e# are doing kind of the same thing...

```
>>> from streaming.util.shorthand import normalize_bytes
>>> normalize_bytes('1.5e9mib')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ape/dev/STREAMING/arg-shorthand/streaming/util/shorthand.py", line 335, in normalize_bytes
    raise ValueError(f'Invalid bytes: {size}. Reasons: {[str(e) for e in errors]}.')
ValueError: Invalid bytes: 1.5e9mib. Reasons: ['Unit is unknown: e in part 0 (input: 1.5e9mib).', 'Unit is unknown: e in part 0 (input: 1.5e9mib).'].
```

knighton: And scientific notation by itself in your yaml is going to get floated

simulation/interfaces/interface_utils.py
knighton: what is the procedure for checking i didn't break the simulator?

karan6181 (approved):

streaming/util/shorthand.py
knighton: Observations for posterity:
- the `humanfriendly` library is very pretty
- But it lacks "counts"
- So we are rolling our own functionality either way for that vertical
- Also, we are tight on time and this is enough, so I figure let's just go with this and revisit in 2024

PR #574, opened 2024-01-23
James/add varint to mds

knighton:
Add `varint`, `varuint` encodings to MDS.

### DIFF
--- a/tests/test_encodings.py
+++ b/tests/test_encodings.py
@@ -26,7 +26,6 @@ class TestMDSEncodings:
         output = byte_enc.decode(data)
         assert output == data
 
-import tempfile
     @pytest.mark.parametrize('data', ['9', 25])
     def test_byte_encode_invalid_data(self, data: Any):
         with pytest.raises(AttributeError):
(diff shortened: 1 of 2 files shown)


tests/test_encodings.py
```diff
@@ -488,6 +488,19 @@ def test_mds_scalar(self, encoding: str, decoded: Union[int, float], encoded: by
         dec = mdsEnc.mds_decode(encoding, encoded)
         assert dec == decoded
 
+    def test_varints(self):
+        from streaming.base.format.mds.encodings import mds_decode, mds_encode
+        for x in range(-700, 700, 7):
+            y = mds_encode('varint', x)
+            z = mds_decode('varint', y)
+            print(x, y, z)
```
snarayan21, line 496: do we need the print statements both here and below?

snarayan21: Also, can we test ints and their expected lengths that we would encode through tokenization? For example, the `vocab_size` of many models customers train is between 50 and 150k. Testing ints in that range would be useful

snarayan21 (commented):
some testing changes

tests/test_encodings.py
knighton: your spurious print statement life debt is paid

PR #583, opened 2024-01-29
Make yamllint consistent with Composer

b-chu:
Matches the yaml settings for Composer
https://github.com/mosaicml/composer/blob/dev/.yamllint.yaml

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,9 +5,9 @@
 
 version: 2
 updates:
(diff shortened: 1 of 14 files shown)


karan6181:
@b-chu Thanks, Brian, for creating these changes. We will hold this PR off for a week or so since there are huge changes coming into the `main` for the Delta Table reader and then we will re-base this on top of `main` after that.

snarayan21 (approved):
lgtm, pending tests passing!

PR #585, opened 2024-01-29
Canonical File Transformations

knighton:
Canonical file transformations
  1. StreamingDataset
      1. Organization of StreamingDataset init args
      2. StreamingDataset interface
  2. Stream
      1. Stream divided into three shockingly modularizable parts
      2. Stream > Shard > StreamDirConf > Stream > ...
      3. New Stream args replicated into StreamingDataset
      4. Stream init sequence is redesigned to parallelize index downloading
      5. StreamDirConf interface (inherited by Stream)
      6. StreamWeightConf interface (inherited by Stream)
      7. Stream interface (inherits from StreamDirConf and StreamWeightConf)
  3. Shard
      1. The new 3-phase shard lifecycle
      2. Terminology for prepping shards
      3. Mapping use cases to phases to cache
      4. Phaser (i.e., shard file phase cacher-deleter) API
      5. Relating shard formats, shard file phases, and use cases in practice
      6. Writers and Shards correspond, as before
      7. Shard composition
      8. Shard interface (internal)
      9. ShardFile interface (internal)
      10. ShardFilePhase interface (internal)
  4. Appendix

### 1.1. Organization of StreamingDataset init args

  * What to iterate:
      * The streams (either provide streams, or remote and/or local for an implicit stream):
          * ``epoch_size``
          * ``streams``
          * ``remote``
          * ``local``
      * Stream arguments (used as defaults if explicit streams, or as args if implicit stream):
          * Filesystem:
              * ``split``
              * ``index_size`` <- new
              * ``index_hashes`` <- new
          * Schema:
              * ``allow_schema_mismatch`` <- new
              * ``allow_unsafe_types``
              * ``allow_unchecked_resumption`` <- new
          * Downloads:
              * ``download_retry``
              * ``download_timeout``
              * ``download_max_size`` <- new
              * ``validate_hash``
              * ``keep_phases`` <- new
  * How to iterate:
      * Shard lifecycle:
          * ``predownload``
          * ``cache_limit``
      * Reproducibility:
          * ``shuffle_seed``
      * Sampling:
          * ``sampling_method``
          * ``sampling_granularity``
      * Partitioning:
          * ``partition_algo``
          * ``num_canonical_nodes``
          * ``batch_size``
      * Shuffling:
          * ``shuffle``
          * ``shuffle_algo``
          * ``shuffle_seed``
          * ``shuffle_block_size``
      * Batching:
          * ``batching_method``

### 1.2. StreamingDataset interface

```
   def __init__(
        self,
        *,
        epoch_size: Optional[Union[str, int]] = None,
        streams: Optional[Sequence[Stream]] = None,
        remote: Optional[str] = None,
        local: Optional[str] = None,
        split: Optional[str] = None,
        index_size: Optional[Union[str, int]] = None,
        index_hashes: Optional[Dict[str, str]] = None,
        allow_schema_mismatch: bool = False,
        allow_unsafe_types: bool = False,
        allow_unchecked_resumption: bool = True,
        download_retry: Union[str, int] = 2,
        download_timeout: Optional[Union[str, float]] = '2m',
        download_max_size: Optional[Union[str, int]] = '200mb',
        validate_hash: Union[None, str, Sequence[str]] = None,
        keep_phases: Union[None, str, Sequence[str], Dict[str, Optional[bool]], Phaser] = None,
        predownload: Optional[int] = None,
        cache_limit: Optional[Union[str, int]] = None,
        shuffle_seed: int = 9176,
        sampling_method: str = 'balanced',
        sampling_granularity: int = 1,
        partition_algo: str = 'relaxed',
        num_canonical_nodes: Optional[int] = None,
        batch_size: Optional[int] = None,
        shuffle: bool = False,
        shuffle_algo: str = 'py1e',
        shuffle_block_size: Optional[int] = None,
        batching_method: str = 'random',
        **kwargs: Any,
    ) -> None:

    def __del__(self) -> None:
    def size(self) -> int:
    def next_epoch(self) -> int:
    def next_epoch(self, next_epoch: int) -> None:
    def cache_usage(self) -> int:
    def cache_usage(self, cache_usage: int) -> None:
    def __len__(self) -> int:
    def _set_shuffle_block_size(self, world: World):
    def _resume(self, world: World, epoch: int) -> Tuple[int, int]:
    def _resume_incr_epoch(self, world: World) -> Tuple[int, int]:
    def state_dict(self, num_samples: int, from_beginning: bool) -> Dict[str, Any]:
    def load_state_dict(self, obj: Dict[str, Any]) -> None:
    def resample_streams(
        self,
        epoch: int,
        stream_id: Optional[int] = None, 
    ) -> Tuple[NDArray[np.int64], NDArray[np.int64]]:
    def _share_work(self, sample_ids: NDArray[np.int64]) -> Tuple[SharedMemory, SharedMemory]:
    def _attach_work(self) -> Tuple[NDArray[np.int64], SharedMemory, SharedMemory]:
    def _get_work(self, world: World, epoch: int, sample_in_epoch: int) -> NDArray[np.int64]:
    def _evict_shard(self, shard_id: int) -> None:
    def _evict_coldest_shard(self) -> None:
    def evict_shard(self, shard_id: int) -> None:
    def evict_coldest_shard(self) -> None:
    def fetch_shard(self, shard_id: int, blocking: bool = True) -> None:
    def get_item(self, sample_id: int, retry: int = 7) -> Any:
    def on_exception(self, future: Future) -> None:
    def _fetch_thread(self, it: _Iterator) -> None:
    def _ready_thread(self, it: _Iterator) -> None:
    def _each_sample_id(self, it: _Iterator) -> Iterator[int]:
    def __iter__(self) -> Iterator[Dict[str, Any]]:
```

### 2.1. Stream divided into three shockingly modularizable parts

`Stream` still has pretty much the same API imported from the same place. However, under the hood, it inherits most of its functionality from two base classes or mixins: `StreamWeightConf` (everything to do with weighting streams) and `StreamDirConf` (handling for all other stream args). What remains for `Stream` itself to do is index loading/having the shards.

### 2.2. Stream > Shard > StreamDirConf > Stream > ...

The shard API is rewritten to be a lot nicer, with methods like `download()` and `evict()` which take no args. It does this by keeping a reference back to its owning `StreamDirConf`, which is mostly about `Stream` arguments which are shared by all its `Shard`s. And recalling that `StreamDirConf` is an ancestor of `Stream`, which technically owns its `Shard`s, we now cue triumphant "it's the circle of life" Lion King music.

Alternative 1: look up the owning Stream and pass large numbers of its arguments to every call to a Shard method, which would be unpleasant and not help anything.

Alternative 2: have the functionality live in Stream, and look up the shard on every call to a shard method, which would be unworkable because there is a class hierarhcy of Shards.

Alternative 3: Do it functionally, which would present the same annoyances around Shard subclassing with the added annoyance of having literally all the things needed having from both stream and shard having to be passed in as args.

### 2.3. New Stream args replicated into StreamingDataset

Title.

### 2.4. Stream init sequence is redesigned to parallelize index downloading

Although there is a faster way within our reach here: using `pool.imap_unordered` and noting the IDs of Streams whose indexes have been downloaded to a shared memory array and initting Streams in that order. Future PR.

### 2.5. StreamDirConf interface (inherited by Stream)

```
    def __init__(
        self,
        *,
        remote: Optional[str] = None,
        local: Optional[str] = None,
        split: Union[None, str, Auto] = Auto(),
        index_size: Union[None, str, int] = None,
        index_hashes: Optional[Dict[str, str]] = None,
        allow_schema_mismatch: Union[bool, Auto] = Auto(),
        allow_unsafe_types: Union[bool, Auto] = Auto(),
        allow_unchecked_resumption: Union[bool, Auto] = Auto(),
        download_retry: Union[str, int, Auto] = Auto(),
        download_timeout: Union[None, str, float, Auto] = Auto(),
        download_max_size: Union[None, str, int, Auto] = Auto(),
        validate_hash: Union[None, str, Sequence[str], Auto] = Auto(),
        keep_phases: Union[None, str, Sequence[str], Dict[str, Optional[bool]], Phaser, Auto] = \
            Auto(),
        **kwargs: Any,
    ) -> None:

    def apply_defaults(
        self,
        *,
        split: Union[None, str, Auto] = Auto(),
        allow_schema_mismatch: Union[bool, Auto] = Auto(),
        allow_unsafe_types: Union[bool, Auto] = Auto(),
        allow_unchecked_resumption: Union[bool, Auto] = Auto(),
        download_retry: Union[str, int, Auto] = Auto(),
        download_timeout: Union[str, float, Auto] = Auto(),
        download_max_size: Union[str, int, Auto] = Auto(),
        validate_hash: Union[None, str, Sequence[str], Auto] = Auto(),
        keep_phases: Union[None, str, Sequence[str], Dict[str, Optional[bool]], Phaser, Auto] = \
            Auto(),
        **kwargs: Any,
    ) -> None:
```

### 2.6. StreamWeightConf interface (inherited by Stream)

```
    def __init__(
        self,
        *,
        proportion: Optional[float] = None,
        repeat: Optional[float] = None,
        choose: Optional[Union[str, int]] = None,
        **kwargs: Dict[str, Any],
    ) -> None:

    def validate_weights(
        cls,
        streams: Sequence[Self],
    ) -> Tuple[bool, bool]:

    def apply_weights(
        cls,
        streams: Sequence[Self],
        samples_per_stream: NDArray[np.int64],
        choose_per_epoch: Optional[int],
        seed: int,
    ) -> int:
```

### 2.7. Stream interface (inherits from StreamDirConf and StreamWeightConf)

```
    def __init__(
        self,
        **kwargs: Any,
    ) -> None:

    def download_index(self) -> None:
    def await_index(self) -> None:
    def load_index(self) -> List[Shard]:
    def inventory_local(self, du_per_shard: NDArray[np.int64]) -> None:
```

### 3.1. The new 3-phase shard lifecycle

```
     /═════╗
     ║ ·1· ║
     ║ Zip ║
     ╚═════/
          \
      - Download -
            \
        - Decompress -
              \
             /═════╗
             ║ ·2· ║
             ║ Raw ║
             ╚═════/   
                  \
             - Canonicalize -
                    \
                   /═════╗
                   ║ ·3· ║
                   ║ Can ║
                   ╚═════/
                        \
                     - Evict -
```

 (Zip != Raw != Can) = shard file.
 
### 3.2. Terminology for prepping shards

```
Fetch
 ├─ Download
 └─ Unpack
     ├─ Decompress
     └─ Canonicalize
```

You fetch, you access, then finally you evict.

Fetching consists of downloading and unpacking.

Unpacking consists of decompressing and canonicalizing.

### 3.3. Mapping use cases to phases to cache

1. For storage/download/remoting: Zip or Raw
2. Intermediates (for caching all phases): everything in-between 1 and 3
3. For (slow) checked resumption: Raw (or Can, in an edge case, someday?)
4. For access and fast (unchecked) resumption: Raw or Can

### 3.4. Phaser (i.e., shard file phase cacher-deleter) API

```
            storage: Union[bool, Auto] = Auto(),
            intermediates: Union[bool, Auto] = Auto(),
            checked_resumption: Union[bool, Auto] = Auto(),
            access: Union[bool, Auto] = Auto(),
            zip: Union[None, bool, Auto] = Auto(),
            raw: Union[None, bool, Auto] = Auto(),
            can: Union[None, bool, Auto] = Auto(),
```

### 3.5. Relating shard formats, shard file phases, and use cases in practice

| Format | Zip | Raw | Can | Storage | Medial | CheckedRes | Access | 
| :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: |
| MDS (z) | *.mds.zst | *.mds | - | Zip | - | Raw | Raw |
| MDS | - | *.mds | - | Raw | - | Raw | Raw |
| JSONL (z) | *.jsonl.zst | *.jsonl | - | Zip | - | Raw | Raw |
| JSONL | - | *.jsonl | - | Raw | - | Raw | Raw |
| XSV (z) | *.xsv.zst | *.xsv | - | Zip | - | Raw | Raw |
| XSV | - | *.xsv | - | Raw | - | Raw | Raw |
| CSV (z) | *.csv.zst | *.csv | - | Zip | - | Raw | Raw |
| CSV | - | *.csv | - | Raw | - | Raw | Raw |
| TSV (z) | *.tsv.zst | *.tsv | - | Zip | - | Raw | Raw |
| TSV | - | *.tsv | - | Raw | - | Raw | Raw |
| *Parquet* | - | *.parquet | *.mds | Raw | - | Raw | Can |
| *Delta* | - | *.parquet | *.mds | Raw | - | Raw | Can |

### 3.6. Writers and Shards correspond, as before

```
Writer
 └─ RowWriter
     ├─ DualRowWriter
     │    ├─ JSONLWriter
     │    └─ XSVWriter
     │        ├─ CSVWriter
     │        └─ TSVWriter
     └─ MonoRowWriter
         └─ MDSWriter
```

Writers write Streaming dataset directories, one shard at a time.

```
Shard
 └─ RowShard
     ├─ DualRowShard
     │   ├─ JSONLShard
     │   └─ XSVShard
     │       ├─ CSVShard
     │       └─ TSVShard
     └─ MonoRowShard
         └─ MDSShard
```

Shards read individual Streaming shards of dataset directories.

### 3.7. Shard composition

```
Shard
 └─ ShardFile
     └─ ShardFilePhase
```

A shard is realized as one or more files.

A file has from one to three forms, which are called phases.

At a high level, the phases all contain the same information, but stored differently for different use cases, e.g. minimizing download size vs minimizing random access time.

### 3.8. Shard interface (internal)

```
    def __init__(
        self,
        *,
        writer_conf: Optional[WriterConf] = None,
        stream: StreamDirConf,
        num_samples: int,
    ) -> None:
    def from_json(cls, stream: StreamDirConf, obj: Dict[str, Any]) -> Self:
    def validate(self) -> None:
    def __len__(self) -> int:
    def set_stream(self, stream: StreamDirConf) -> None:
    def inventory_local(self, listing: Set[str]) -> Optional[int]:
    def fetch(self) -> int:
    def evict(self) -> int:
    def size(self) -> int:
    def get_item(self, index: int) -> Dict[str, Any]:
```

### 3.9. ShardFile interface (internal)

```
    def __init__(
        self,
        *,
        stream: StreamDirConf,
        zip_phase: Optional[ShardFilePhase] = None,
        zip_algo: Optional[str] = None,
        raw_phase: ShardFilePhase,
        can_algo: Optional[str] = None,
        can_phase: Optional[ShardFilePhase] = None,
    ) -> None:
    def set_stream(self, stream: StreamDirConf) -> None:
    def validate(self) -> None:
    def locate(self, listing: Set[str]) -> NDArray[np.int64]:
    def inventory_local(self, listing: Set[str]) -> int:
    def _unzip(self) -> int:
    def _canonicalize(self) -> int:
    def _load_raw(self) -> int:
    def fetch(self) -> int:
    def evict(self) -> int:
    def evict_phases(self, phase_dels: NDArray[np.int64]) -> int:
```

### 3.10. ShardFilePhase interface (internal)

```
    def __init__(
        self,
        *,
        stream: StreamDirConf,
        relative_path: str,
        size: Optional[int] = None,
        hashes: Optional[Dict[str, str]] = None,
    ) -> None:
    def from_json(cls, stream: StreamDirConf, obj: Dict[str, Any]) -> Self:
    def set_stream(self, stream: StreamDirConf) -> None:
    def validate_for_download(self) -> None:
    def get_local_filename(self) -> str:
    def get_remote_filename(self) -> str:
    def probe(self, listing: Set[str]) -> bool:
    def inventory_local(self, listing: Set[str]) -> Optional[int]:
    def is_local(self) -> bool:
    def download(self) -> int:
    def evict(self) -> int:
```

### Source tree changes

```
	new file:   streaming/format/base/__init__.py
	new file:   streaming/format/base/file.py
	new file:   streaming/format/base/phase.py
	new file:   streaming/format/base/phaser.py
	new file:   streaming/format/base/shard/__init__.py
	new file:   streaming/format/base/shard/base.py
	new file:   streaming/format/base/shard/dual_row.py
	new file:   streaming/format/base/shard/mono_row.py
	new file:   streaming/format/base/shard/row.py
	new file:   streaming/format/base/writer/__init__.py
	new file:   streaming/format/base/writer/base.py
	new file:   streaming/format/base/writer/dual_row.py
	new file:   streaming/format/base/writer/mono_row.py
	new file:   streaming/format/canonical.py
	new file:   streaming/stream/__init__.py
	new file:   streaming/stream/base.py
	new file:   streaming/stream/dir_conf.py
	new file:   streaming/stream/weight_conf.py
	new file:   streaming/util/json.py
	new file:   streaming/util/waiting.py
```

```
	modified:   benchmarks/__init__.py
	modified:   benchmarks/backends/datagen.py
	modified:   benchmarks/backends/plot.py
	modified:   benchmarks/backends/read.py
	modified:   benchmarks/backends/write.py
	modified:   examples/__init__.py
	modified:   examples/multimodal/__init__.py
	modified:   examples/multimodal/laion400m/__init__.py
	modified:   examples/multimodal/webvid/__init__.py
	modified:   examples/multimodal/webvid/read.py
	modified:   examples/text/__init__.py
	modified:   examples/vision/__init__.py
	modified:   examples/vision/ade20k/__init__.py
	modified:   examples/vision/cifar10/__init__.py
	modified:   examples/vision/cifar10/read.py
	modified:   examples/vision/cifar10/write.py
	modified:   examples/vision/cifar10/write_fake.py
	modified:   examples/vision/coco/__init__.py
	modified:   examples/vision/imagenet/__init__.py
	modified:   examples/vision/imagenet/read.py
	modified:   notebooks/spark_dataframe_to_MDS.ipynb
	modified:   scripts/long_lines.py
	modified:   simulation/core/sim_dataset.py
	modified:   streaming/dataset.py
	modified:   streaming/format/__init__.py
	modified:   streaming/format/jsonl/__init__.py
	modified:   streaming/format/jsonl/writer.py
	modified:   streaming/format/mds/shard.py
	modified:   streaming/format/mds/writer.py
	modified:   streaming/format/xsv/writer.py
	modified:   streaming/phasing.py
	modified:   streaming/shared/__init__.py
	modified:   streaming/storage/__init__.py
	modified:   streaming/storage/extra.py
	modified:   streaming/util/__init__.py
	modified:   streaming/util/auto.py
	modified:   streaming/util/importing.py
	modified:   streaming/util/merging.py
	modified:   streaming/util/retrying.py
	modified:   streaming/util/shared.py
	modified:   streaming/util/shorthand.py
	modified:   streaming/util/tabulation.py
	modified:   tests/test_eviction.py
	modified:   tests/test_importing.py
	modified:   tests/test_unsafe_types.py
	modified:   tests/util/__init__.py
	modified:   tests/util/test_retrying.py
	modified:   tests/util/test_shared.py
	modified:   tests/util/test_shorthand.py
```

```
	deleted:    streaming/base/converters/README.md
	deleted:    streaming/base/format/base/__init__.py
	deleted:    streaming/base/shared/__init__.py
	deleted:    streaming/base/storage/__init__.py
	deleted:    streaming/format/score.py
	deleted:    streaming/format/shard.py
	deleted:    streaming/multimodal/__init__.py
	deleted:    streaming/stream.py
	deleted:    streaming/text/__init__.py
	deleted:    streaming/vision/__init__.py
	deleted:    streaming/vision/base.py
	deleted:    streaming/vision/convert/base.py
```

- Phase caching:
    - By intended use cases: Storage -> Intermediates -> Checked Resumption -> Access
    - By phase: Zip (MDS or JSONL compressed) -> Raw (MDS or Parquet or JSONL) -> Canonical
- Checked vs fast resumption Corner Cases
    - For performance reasons, we don't necessarily know the size of the Parquet shard canonicalized/MDS phase in advance, much less hashed it
    - This leaves us unable to validate such shard files upon resumption with them in the cache
    - If we keep the files in the cache, we do no work in prepare() but are at the mercy of whoever had access to that directroy, which is potentially dangerous depending on threat model
    - If we drop the files from the cache, we'll have to regenerate them during prepare() from the last checked phase, which is potentially expensive depending on size of shard

### DIFF
--- a/scripts/long_lines.py
+++ b/scripts/long_lines.py
@@ -199,11 +199,11 @@ def main(args: Namespace) -> int:
     max_vis_len = max(vis_lens)
     vocab = 0x2571, 0x2572
     for vis_len, out_line in pairs:
-        if args.pad:
+        if args.fancy:
             count = max_vis_len - vis_len + 1
             ords = np.random.choice(vocab, count)
-            pad = ''.join(map(chr, ords))
-            print(f'{out_line[:-1]}{chr(0x2523)}{pad}')
+            padding = ''.join(map(chr, ords))
+            print(f'{out_line[:-1]}{chr(0x2523)}{padding}')
         else:
             print(out_line[:-1])
 
--- a/simulation/core/sim_dataset.py
+++ b/simulation/core/sim_dataset.py
@@ -370,8 +370,19 @@ class SimulationDataset(StreamingDataset):
         # Also keep track of the raw and compressed sizes of each shard, indexed by shard_id.
         # TODO: replace `get_raw_size()` and `get_zip_size()` with appropriate methods from the
         # shard/file/phase API. Can assume that we are using MDS files for the simulator.
-        self.raw_shard_sizes = np.array([shard.get_raw_size() for shard in self.shards], np.int64)
-        self.zip_shard_sizes = np.array([shard.get_zip_size() or 0 for shard in self.shards],
+
+        # Get shard raw and zip sizes.
+        for shard in self.shards:
+            if len(shard.files) > 1:
+                raise ValueError(f'The Streaming Simulator currently only supports datasets ',
+                                 f'in MDS format. Please make sure your datasets are in MDS.')
+        self.raw_shard_sizes = np.array([shard.files[0].raw_phase.expected_size \
+                                         for shard in self.shards],
+                                        np.int64)
+        self.zip_shard_sizes = np.array([shard.files[0].zip_phase.expected_size or 0 \
+                                         if shard.files[0].zip_phase is not None \
+                                         else 0 \
+                                         for shard in self.shards],
                                         np.int64)
 
         logger.info(f' Total number of shards: {self.num_shards}')
--- a/simulation/core/sim_dataset.py
+++ b/simulation/core/sim_dataset.py
@@ -9,7 +9,8 @@ import shutil
 import time
 import warnings
 from math import ceil
-from typing import Optional, Sequence, Union
+from multiprocessing import Pool
+from typing import Any, Dict, Optional, Sequence, Union
 
 import numpy as np
 from core.sim_spanner import SimulationSpanner
@@ -19,6 +20,7 @@ from numpy.typing import NDArray
 from streaming import Stream, StreamingDataset
 from streaming.batching import generate_work
 from streaming.format import get_index_basename
+from streaming.format.base.phaser import Phaser
 from streaming.spanner import Spanner
 from streaming.util.shorthand import normalize_bytes, normalize_count
 
@@ -44,17 +46,26 @@ class SimulationDataset(StreamingDataset):
             StreamingDataset uses either ``streams`` or ``remote``/``local``. Defaults to ``None``.
         split (str, optional): Which dataset split to use, if any. If provided, we stream from/to
             the ``split`` subdirs of  ``remote`` and ``local``. Defaults to ``None``.
+        allow_schema_mismatch (bool): If ``True``, continue if sample columns mismatch across
+            shards, streams, or the whole dataset. If ``False``, raises if columns mismatch.
+            Defaults to ``False``.
+        allow_unsafe_types (bool): If ``True``, continue if unsafe type(s) are encountered
+            in shard(s). If ``False``, raises if unsafe type(s) encountered. Defaults to ``False``.
+        allow_unchecked_resumption (bool): If ``True``, upon resume, accept and use shard
+            file phases that we are unable to check the size/hash(es) of. If ``False``, upon
+            resume, drop such files, to regenerate on the fly when needed. Defaults to ``True``.
         download_retry (int): Number of download re-attempts before raising an error. Defaults to
             ``2``.
         download_timeout (str | float): Time in seconds to wait for a file download to complete
             before raising an error. Streaming duration shorthand (e.g., ``1m23s``) is also
             accepted. Defaults to ``1m``.
-        hash_algos (str | Sequence[str], optional): Ranked list of hashing algorithms to try.
-            Defaults to ``None``.
         validate_hash (str, optional): Deprecated. See ``hash_algos``. Defaults to ``None``.
-        keep_old_phases (str): Which old phases of shard files to cache (until shard eviction).
-            Must be one of ``nil``, ``src``, or ``all``. Defaults to ``nil``.
-        keep_zip (bool, optional): Deprecated. See ``keep_old_phases``. Defaults to ``None``.
+        keep_phases (str | Sequence[str] | Dict[str, bool] | Phaser, optional): After a phase
+            transition of a shard file, do we keep the old form of the file or garbage collect it?
+            Provided as one of: (1) ``None`` for defaults, (2) the single use case or phase to keep,
+            (3) a sequence giving the use cases or phases to keep, (4) Phaser kwargs (a mapping of
+            use case or phase to whether it must be kept, or (5) a Phaser object. All code paths
+            result in a ``Phaser``. Defaults to ``None``.
         epoch_size (Union[str, int], optional): Number of samples to draw per epoch balanced
             across all streams. If ``None``, takes its value from the total number of underlying
             samples. Provide this field if you are weighting streams relatively to target a larger
@@ -107,35 +118,39 @@ class SimulationDataset(StreamingDataset):
             if ``False``. Defaults to ``False``.
     """
 
-    def __init__(self,
-                 *,
-                 nodes: int,
-                 devices: int,
-                 workers: int,
-                 streams: Optional[Sequence[Stream]] = None,
-                 remote: Optional[str] = None,
-                 local: Optional[str] = None,
-                 split: Optional[str] = None,
-                 download_retry: int = 2,
-                 download_timeout: Union[str, float] = '1m',
-                 hash_algos: Optional[Union[str, Sequence[str]]] = None,
-                 validate_hash: Optional[str] = None,
-                 keep_old_phases: str = 'nil',
-                 keep_zip: Optional[bool] = None,
-                 epoch_size: Optional[Union[str, int]] = None,
-                 predownload: Optional[int] = None,
-                 cache_limit: Optional[Union[str, int]] = None,
-                 sampling_method: str = 'balanced',
-                 sampling_granularity: int = 1,
-                 partition_algo: str = 'relaxed',
-                 num_canonical_nodes: Optional[int] = None,
-                 batch_size: Optional[int] = None,
-                 shuffle: bool = False,
-                 shuffle_algo: str = 'py1e',
-                 shuffle_seed: int = 9176,
-                 shuffle_block_size: Optional[int] = None,
-                 batching_method: str = 'random',
-                 allow_unsafe_types: bool = False) -> None:
+    def __init__(
+        self,
+        *,
+        nodes: int,
+        devices: int,
+        workers: int,
+        streams: Optional[Sequence[Stream]] = None,
+        remote: Optional[str] = None,
+        local: Optional[str] = None,
+        split: Optional[str] = None,
+        allow_schema_mismatch: bool = False,
+        allow_unsafe_types: bool = False,
+        allow_unchecked_resumption: bool = True,
+        download_retry: int = 2,
+        download_timeout: Union[str, float] = '2m',
+        download_max_size: Optional[Union[str, int]] = '200mb',
+        validate_hash: Optional[str] = None,
+        keep_phases: Union[None, str, Sequence[str], Dict[str, bool], Phaser] = None,
+        epoch_size: Optional[Union[str, int]] = None,
+        predownload: Optional[int] = None,
+        cache_limit: Optional[Union[str, int]] = None,
+        sampling_method: str = 'balanced',
+        sampling_granularity: int = 1,
+        partition_algo: str = 'relaxed',
+        num_canonical_nodes: Optional[int] = None,
+        batch_size: Optional[int] = None,
+        shuffle: bool = False,
+        shuffle_algo: str = 'py1e',
+        shuffle_seed: int = 9176,
+        shuffle_block_size: Optional[int] = None,
+        batching_method: str = 'random',
+        **kwargs: Any,
+    ) -> None:
         # Time how long it takes for StreamingDataset instantiation
         t0 = time.time()
 
@@ -202,25 +217,22 @@ class SimulationDataset(StreamingDataset):
         epoch_size_value = normalize_count(epoch_size) if epoch_size else None
 
         # Initialize the Stream defaults and normalize to a list of Streams.
+        kwargs.update(
+            split=split,
+            allow_schema_mismatch=allow_schema_mismatch,
+            allow_unsafe_types=allow_unsafe_types,
+            allow_unchecked_resumption=allow_unchecked_resumption,
+            download_retry=download_retry,
+            download_timeout=download_timeout,
+            download_max_size=download_max_size,
+            validate_hash=validate_hash,
+            keep_phases=keep_phases,
+        )
         if streams:
             for stream in streams:
-                stream.apply_defaults(split=split,
-                                      download_retry=download_retry,
-                                      download_timeout=download_timeout,
-                                      hash_algos=hash_algos,
-                                      validate_hash=validate_hash,
-                                      keep_old_phases=keep_old_phases,
-                                      keep_zip=keep_zip)
+                stream.apply_defaults(**kwargs)
         else:
-            streams = Stream(remote=remote,
-                             local=local,
-                             split=split,
-                             download_retry=download_retry,
-                             download_timeout=download_timeout,
-                             hash_algos=hash_algos,
-                             validate_hash=validate_hash,
-                             keep_old_phases=keep_old_phases,
-                             keep_zip=keep_zip),
+            streams = Stream(remote=remote, local=local, **kwargs),
 
         # Validate the stream weighting scheme (relative or absolute) to catch errors before we go
         # to the trouble of loading them.
@@ -257,7 +269,19 @@ class SimulationDataset(StreamingDataset):
         # Initialize the SimulationWorld, which tells us about nodes/devices/workers
         self.world = SimulationWorld(self.nodes, self.devices, self.workers)
 
-        # Download each stream's index, load their shards, and map streams <-> shards.
+        # Download each Stream's index in parallel from local rank 0.
+        #
+        # Parallelism is important because there could be a very large number of Streams, and we
+        # expect equal performance as them having been concatenated into one Stream.
+        if self.world.is_local_leader:
+            pool = Pool()
+            pool.imap_unordered(Stream.download_index, self.streams)
+            pool.close()
+        else:
+            pool = None
+
+        # All ranks then walk all Streams, for which they (1) wait for its index to become
+        # downloaded, (2) load its shards, and (3) map streams <-> shards <-> samples.
         self.num_samples = 0
         self.shards = []
         stream_per_shard = []
@@ -269,13 +293,14 @@ class SimulationDataset(StreamingDataset):
         local_foldernames = []
         for stream_id, stream in enumerate(self.streams):
             logger.info(f' Processing index file for stream {stream_id + 1}')
+            stream.await_index()
             stream_shards = stream.load_index()
             num_stream_samples = sum(map(len, stream_shards))
             index_filename = os.path.join(stream.local, stream.split or '', get_index_basename())
             index_filenames.append(index_filename)
             local_foldernames.append(stream.local)
             if not num_stream_samples:
-                raise RuntimeError(f'Stream contains no samples: {index_filename}.')
+                raise RuntimeError(f'Stream contains no samples: {stream.local_index_path}.')
             stream_per_shard += [stream_id] * len(stream_shards)
             self.shard_offset_per_stream[stream_id] = len(self.shards)
             self.shards_per_stream[stream_id] = len(stream_shards)
@@ -283,9 +308,12 @@ class SimulationDataset(StreamingDataset):
             self.samples_per_stream[stream_id] = num_stream_samples
             self.shards += stream_shards
             self.num_samples += num_stream_samples
-
         self.stream_per_shard = np.array(stream_per_shard, np.int64)
         self.num_shards = len(self.shards)
+
+        # Wait for the pool workers (stream index download processes) to finish.
+        if pool is not None:
+            pool.join()
 
         # Check that cache limit is possible.
         if cache_limit:
@@ -294,7 +322,28 @@ class SimulationDataset(StreamingDataset):
             if self.cache_limit <= min_cache_usage:
                 raise ValueError(f'Minimum cache usage ({min_cache_usage} bytes) is larger than ' +
                                  f'the cache limit ({self.cache_limit} bytes). Please raise ' +
-                                 f'`cache_limit`.')
+                                 f'`cache_limit`. Recommendation is to provide a `cache_limit` ' +
+                                 f'as high as possible to avoid thrashing.')
+
+            max_phase_size = 0
+            for stream in self.streams:
+                for shard in stream.shards:
+                    for file in shard.files:
+                        for phase in file.phases:
+                            if not phase:
+                                continue
+                            if not phase.expected_size:
+                                continue
+                            if max_phase_size < phase.expected_size:
+                                max_phase_size = phase.expected_size
+
+            if self.cache_limit < 4 * max_phase_size:
+                raise ValueError(f'Cache limit ({self.cache_limit} bytes) is too low. ' +
+                                 f'Increase the `cache_limit` to at-least four times the ' +
+                                 f'largest shard size ({max_phase_size} ' +
+                                 f'bytes) which includes raw (decompressed) and zip ' +
+                                 f'(compressed) file size. Recommendation is to provide a ' +
+                                 f'`cache_limit` as high as possible to avoid thrashing.')
         else:
             self.cache_limit = None
 
@@ -319,6 +368,8 @@ class SimulationDataset(StreamingDataset):
         self.spanner = SimulationSpanner(self.samples_per_shard)
 
         # Also keep track of the raw and compressed sizes of each shard, indexed by shard_id.
+        # TODO: replace `get_raw_size()` and `get_zip_size()` with appropriate methods from the
+        # shard/file/phase API. Can assume that we are using MDS files for the simulator.
         self.raw_shard_sizes = np.array([shard.get_raw_size() for shard in self.shards], np.int64)
         self.zip_shard_sizes = np.array([shard.get_zip_size() or 0 for shard in self.shards],
                                         np.int64)
--- a/simulation/core/sim_dataset.py
+++ b/simulation/core/sim_dataset.py
@@ -367,11 +367,7 @@ class SimulationDataset(StreamingDataset):
         self.sample_offset_per_shard = self.samples_per_shard.cumsum() - self.samples_per_shard
         self.spanner = SimulationSpanner(self.samples_per_shard)
 
-        # Also keep track of the raw and compressed sizes of each shard, indexed by shard_id.
-        # TODO: replace `get_raw_size()` and `get_zip_size()` with appropriate methods from the
-        # shard/file/phase API. Can assume that we are using MDS files for the simulator.
-
-        # Get shard raw and zip sizes.
+        # Also keep track of the raw and compressed sizes of each shard.
         for shard in self.shards:
             if len(shard.files) > 1:
                 raise ValueError(f'The Streaming Simulator currently only supports datasets ',
--- a/simulation/core/sim_dataset.py
+++ b/simulation/core/sim_dataset.py
@@ -290,7 +290,7 @@ class SimulationDataset(StreamingDataset):
         # Check that cache limit is possible.
         if cache_limit:
             self.cache_limit = normalize_bytes(cache_limit)
-            min_cache_usage = sum((stream.got_index_size for stream in streams))
+            min_cache_usage = sum((stream.index_size for stream in streams))
             if self.cache_limit <= min_cache_usage:
                 raise ValueError(f'Minimum cache usage ({min_cache_usage} bytes) is larger than ' +
                                  f'the cache limit ({self.cache_limit} bytes). Please raise ' +
--- a/streaming/dataset.py
+++ b/streaming/dataset.py
@@ -1195,7 +1195,7 @@ class StreamingDataset(Array, IterableDataset):
         Args:
             shard_id (int): Shard to download.
             blocking (bool): Whether to wait or skip if the shard is currently being downloaded by
-                someone else.
+                someone else. Defaults to ``True``.
         """
         # Lock the cache. FileLocks contain threading Locks, which are not pickleable, which is
         # incompatible with spawn, so must be created lazily.
--- a/streaming/dataset.py
+++ b/streaming/dataset.py
@@ -206,6 +206,10 @@ class StreamingDataset(Array, IterableDataset):
                   * ``validate_hash``
                   * ``keep_phases``
       * How to iterate:
+          * Init:
+              * ``index_download_procs``
+              * ``index_download_procs_per_cpu``
+              * ``index_download_max_procs``
           * Shard lifecycle:
               * ``predownload``
               * ``cache_limit``
@@ -267,6 +271,15 @@ class StreamingDataset(Array, IterableDataset):
             Specified as a single use or phase to keep, a sequence of uses or phases to keep, a
             mapping of uses or phases to whether to keep or drop, or a ``Phaser`` (which performs
             the same keeping or dropping). Defaults to ``None``.
+        index_download_procs (int, optional): Size of the process pool. You may set either this
+            arg or ``index_download_procs_per_cpu``, but not both. Defaults to ``None``.
+        index_download_procs_per_cpu (float | int, optional): Size of the process pool as the
+            number of processes per CPU. You may set either this arg or ``index_download_procs``,
+            but not both. If its value is negative, it is taken to be the reciprocal of that value
+            as a positive number, e.g. ``-4`` means ``num_cpus // 4``. Defaults to ``None``.
+        index_download_max_procs (int, optional): Optional ceiling on the number of index
+            downlaod processes. Used to not overwhelm remote storage with concurrency connectinos.
+            Defaults to ``None``.
         predownload (int, optional): Target number of samples to download per worker in advance
             of current sample. Workers will attempt to download ahead by this many samples during,
             but not before, training. Recommendation is to provide a value greater than per device
@@ -333,6 +346,9 @@ class StreamingDataset(Array, IterableDataset):
         download_max_size: Optional[Union[str, int]] = '200mb',
         validate_hash: Union[None, str, Sequence[str]] = None,
         keep_phases: Union[None, str, Sequence[str], Dict[str, Optional[bool]], Phaser] = None,
+        index_download_procs: Optional[int] = None,
+        index_download_procs_per_cpu: Optional[Union[float, int]] = 4,
+        index_download_max_procs: Optional[int] = 64,
         predownload: Optional[int] = None,
         cache_limit: Optional[Union[str, int]] = None,
         shuffle_seed: int = 9176,
@@ -473,8 +489,16 @@ class StreamingDataset(Array, IterableDataset):
         # Parallelism is important because there could be a very large number of Streams, and we
         # expect equal performance as them having been concatenated into one Stream.
         if world.is_local_leader:
-            with Pool() as pool:
-                pool.imap_unordered(lambda stream: stream.download_index(), self.streams)
+            num_procs = self._get_index_download_procs(
+                index_download_procs,
+                index_download_procs_per_cpu,
+                index_download_max_procs,
+            )
+            pool = Pool(num_procs)
+            pool.imap_unordered(lambda stream: stream.download_index(), self.streams)
+            pool.close()
+        else:
+            pool = None
 
         # All ranks then walk all Streams, for which they (1) wait for its index to become
         # downloaded, (2) load its shards, and (3) map streams <-> shards <-> samples.
@@ -500,6 +524,9 @@ class StreamingDataset(Array, IterableDataset):
             self.num_samples += num_stream_samples
         self.stream_per_shard = np.array(stream_per_shard, np.int64)
         self.num_shards = len(self.shards)
+
+        if pool is not None:
+            pool.join()
 
         # Check that cache limit is possible.
         if cache_limit:
@@ -689,7 +716,64 @@ class StreamingDataset(Array, IterableDataset):
         """
         return self.length
 
-    def _set_shuffle_block_size(self, world: World):
+    def _get_index_download_procs(
+        self,
+        index_download_procs: Optional[int] = None,
+        index_download_procs_per_cpu: Optional[Union[float, int]] = 4,
+        index_download_max_procs: Optional[int] = 64,
+    ) -> int:
+        """Get the size of the process pool used by index downloading.
+
+        Args:
+            index_download_procs (int, optional): Size of the process pool. You may set either this
+                arg or ``index_download_procs_per_cpu``, but not both. Defaults to ``None``.
+            index_download_procs_per_cpu (float | int, optional): Size of the process pool as the
+                number of processes per CPU. You may set either this arg or
+                ``index_download_procs``, but not both. If its value is negative, it is taken to be
+                the reciprocal of that value as a positive number, e.g. ``-4`` means
+                ``num_cpus // 4``. Defaults to ``4``.
+            index_download_max_procs (int, optional): Optional ceiling on the number of index
+                downlaod processes. Used to not overwhelm remote storage with concurrency
+                connectinos. Defaults to ``64``.
+        """
+        if index_download_procs is not None:
+            if index_download_procs_per_cpu is not None:
+                raise ValueError(
+                    f'You may specify `index_download_procs`, `index_download_procs_per_cpu`, ' +
+                    f'or neither, but got both: {index_download_procs} and ' +
+                    f'{index_download_procs_per_cpu} respectively.')
+            else:
+                num_procs = index_download_procs
+                if num_procs <= 0:
+                    raise ValueError(f'`index_download_procs` must be a positive integer.')
+        else:
+            num_cpus = os.cpu_count() or 1
+            if index_download_procs_per_cpu is not None:
+                procs_per_cpu = index_download_procs_per_cpu
+                if procs_per_cpu < 0:
+                    num_procs = num_cpus / -procs_per_cpu
+                elif not procs_per_cpu:
+                    num_procs = num_cpus
+                else:
+                    num_procs = num_cpus * procs_per_cpu
+                num_procs = int(np.ceil(num_procs))
+                if num_procs <= 0:
+                    raise ValueError(f'`index_download_procs_per_cpu` must result in a positive ' +
+                                     f'number of index download processes after rounding up.')
+            else:
+                num_procs = num_cpus
+
+        max_procs = index_download_max_procs
+        if max_procs is not None:
+            if max_procs <= 0:
+                raise ValueError(f'`index_download_max_procs` must be a positive integer, but ' +
+                                 f'got: {max_procs}.')
+            if max_procs < num_procs:
+                num_procs = max_procs
+
+        return num_procs
+
+    def _set_shuffle_block_size(self, world: World) -> None:
         """Set the shuffle block size value."""
         if self.shuffle_block_size is None:
             if not world.worker_of_rank:
--- a/streaming/dataset.py
+++ b/streaming/dataset.py
@@ -448,37 +448,21 @@ class StreamingDataset(Array, IterableDataset):
         destroy_dist = maybe_init_dist()
 
         # Initialize the Stream defaults and normalize to a list of Streams.
-        if streams:
-            for stream in streams:
-                stream.apply_defaults(
-                    split=split,
-                    allow_schema_mismatch=allow_schema_mismatch,
-                    allow_unsafe_types=allow_unsafe_types,
-                    allow_unchecked_resumption=allow_unchecked_resumption,
-                    download_retry=download_retry,
-                    download_timeout=download_timeout,
-                    download_max_size=download_max_size,
-                    validate_hash=validate_hash,
-                    keep_phases=keep_phases,
-                    **kwargs,
-                )
-        else:
-            stream = Stream(
-                remote=remote,
-                local=local,
-                split=split,
-                allow_schema_mismatch=allow_schema_mismatch,
-                allow_unsafe_types=allow_unsafe_types,
-                allow_unchecked_resumption=allow_unchecked_resumption,
-                download_retry=download_retry,
-                download_timeout=download_timeout,
-                download_max_size=download_max_size,
-                validate_hash=validate_hash,
-                keep_phases=keep_phases,
-                **kwargs,
-            )
-            stream.apply_defaults()
-            streams = stream,
+        kwargs.update(
+            split=split,
+            allow_schema_mismatch=allow_schema_mismatch,
+            allow_unsafe_types=allow_unsafe_types,
+            allow_unchecked_resumption=allow_unchecked_resumption,
+            download_retry=download_retry,
+            download_timeout=download_timeout,
+            download_max_size=download_max_size,
+            validate_hash=validate_hash,
+            keep_phases=keep_phases,
+        )
+        if not streams:
+            streams = Stream(remote=remote, local=local, **kwargs),
+        for stream in streams:
+            stream.apply_defaults(**kwargs)
 
         # Validate the stream weighting scheme (relative or absolute) to catch errors before we go
         # to the trouble of loading them.
--- a/streaming/dataset.py
+++ b/streaming/dataset.py
@@ -1144,7 +1144,7 @@ class StreamingDataset(Array, IterableDataset):
 
         # Perform the eviction, updating cache usage to account for the removal.
         shard = self.shards[shard_id]
-        self.cache_usage -= shard.evict()
+        self.cache_usage += shard.evict()
         if self.cache_usage < 0:
             raise RuntimeError(f'Negative cache usage: {self.cache_usage}.')
 
--- a/streaming/dataset.py
+++ b/streaming/dataset.py
@@ -171,10 +171,18 @@ class _Iterator:
 def download_index(stream: Stream) -> None:
     """A Stream index download process.
 
+    If the file download fails for any reason, writes an empty file to the target location, causing
+    all ranks to raise an error when they attempt to load and parse the index file. Otherwise, they
+    would have to wait for a timeout, unless we switch to ProcessPoolExecutor.
+
     Args:
         stream (Stream): The stream.
     """
-    stream.download_index()
+    try:
+        stream.download_index()
+    except:
+        with open(stream.local_index_path, 'wb') as out:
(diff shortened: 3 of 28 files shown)


karan6181 (commented):
Can you please also add `delta` branch name under `branches` for files
- [install.yaml](https://github.com/mosaicml/streaming/blob/main/.github/workflows/install.yaml)
- [linting.yaml](https://github.com/mosaicml/streaming/blob/main/.github/workflows/linting.yaml)
- [pytest.yaml](https://github.com/mosaicml/streaming/blob/main/.github/workflows/pytest.yaml)

So that CI runs for the delta branch when you create a PR.

streaming/stream/weight_conf.py
```diff
@@ -0,0 +1,164 @@
...
+        *,
+        proportion: Optional[float] = None,
+        repeat: Optional[float] = None,
+        choose: Optional[Union[str, int]] = None,
+        **kwargs: Dict[str, Any],
+    ) -> None:
+        has_proportion = proportion is not None
+        has_repeat = repeat is not None
+        has_choose = choose is not None
+        if not (0 <= has_proportion + has_repeat + has_choose <= 1):
```
XiaohanZhangCMU, line 42: Should reflect this condition in the API arguments.

XiaohanZhangCMU (changes requested):

XiaohanZhangCMU (changes requested):

simulation/core/sim_dataset.py
```diff
@@ -290,7 +290,7 @@ def __init__(self,
         # Check that cache limit is possible.
         if cache_limit:
             self.cache_limit = normalize_bytes(cache_limit)
-            min_cache_usage = sum((stream.get_index_size() for stream in streams))
+            min_cache_usage = sum((stream.got_index_size for stream in streams))
```
karan6181, line 293: nit: why not just `stream.index_size`?

streaming/dataset.py
```diff
@@ -450,7 +449,16 @@ def __init__(self,
         # instantiate a World inside the StreamingDataset destructor.
         self._rank_world = world = World()
 
-        # Download each stream's index, load their shards, and map streams <-> shards.
+        # Download each Stream's index in parallel from local rank 0.
+        #
+        # Parallelism is important because there could be a very large number of Streams, and we
+        # expect equal performance as them having been concatenated into one Stream.
+        if world.is_local_leader:
+            with Pool() as pool:
```
karan6181, line 457: This is nice!

streaming/dataset.py
```diff
@@ -450,7 +449,16 @@ def __init__(self,
...
+                pool.imap_unordered(lambda stream: stream.download_index(), self.streams)
```
karan6181, line 458: Do we need to call the `wait()` method from `pool,` which blocks until all tasks issued asynchronously are completed?

streaming/dataset.py
```diff
@@ -459,12 +467,11 @@ def __init__(self,
         self.sample_offset_per_stream = np.zeros(self.num_streams, np.int64)
         self.samples_per_stream = np.zeros(self.num_streams, np.int64)
         for stream_id, stream in enumerate(self.streams):
-            stream_shards = stream.get_shards(world, self.allow_unsafe_types)
+            stream.await_index()
```
karan6181, line 470: I think this answers my above question, where this is a blocking call to wait for the `index.json` to cache locally.

streaming/util/json.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2022-2024 MosaicML Streaming authors
```
karan6181, line 1: Should we name this file as `typing.py`?

streaming/format/base/phaser.py
```diff
@@ -0,0 +1,103 @@
...
+    def get_phase_deletions(self, phase_locs: NDArray[np.int64]) -> NDArray[np.int64]:
+        """Get phase deletions.
+
+        Args:
+            phase_locs (NDArray[np.int64]): Phase localities.
+
+        Returns:
+            NDArray[np.int64]: Phase deletions.
+        """
+        return np.zeros(3, np.int64)  # TODO
```
XiaohanZhangCMU, line 103: what's this?

streaming/format/base/file.py
```diff
@@ -0,0 +1,302 @@
...
+        can_algo: Optional[str] = None,
+        can_phase: Optional[ShardFilePhase] = None,
+    ) -> None:
+        self.stream = stream
+        self.zip_phase = zip_phase
+        self.zip_algo = zip_algo
+        self.raw_phase = raw_phase
+        self.can_algo = can_algo
+        self.can_phase = can_phase
+        self.phases = zip_phase, raw_phase, can_phase
```
XiaohanZhangCMU, line 88: nit: cleaner if self.phases = (p for p in (zip_phase, raw_phase, can_phase) if p is not None). since ``if phase`` is performed several times in the code.

streaming/dataset.py
```diff
@@ -1082,7 +1089,7 @@ def evict_coldest_shard(self) -> None:
         with self._cache_filelock:
             self._evict_coldest_shard()
 
-    def prepare_shard(self, shard_id: int, blocking: bool = True) -> None:
+    def fetch_shard(self, shard_id: int, blocking: bool = True) -> None:
```
XiaohanZhangCMU, line 1092: doc string of arguments missing

XiaohanZhangCMU (changes requested):

streaming/format/base/file.py
```diff
@@ -0,0 +1,302 @@
...
+            can_loc = Locality.LOCAL if self.can_phase.is_local() else Locality.REMOTE
+        else:
+            can_loc = Locality.DNE
+        phase_locs = np.array([Locality.LOCAL, Locality.LOCAL, can_loc], np.int64)
+
+        # Given localities and policy, determine phase evictions.
+        phase_dels = self.stream.safe_keep_phases.get_phase_deletions(phase_locs)
+
+        # Delete phases we don't want.
+        return len(raw_data) + self.evict_phases(phase_dels)
```
snarayan21, line 207: wouldn't we want to subtract out the size of the evicted phases from the length of the raw data to correctly calculate cache usage? If i'm deleting some phases then that's a reduction in cache usage, right?

if this is correct (I may be misunderstanding) then there should be some other subtractions in this file too

snarayan21: oh just looked at `phase.py`, looks like `evict` returns a negative number

streaming/format/base/phase.py
```diff
@@ -0,0 +1,233 @@
...
+                    # Expected size is unknown. Is that a problem?
+                    if self.stream.allow_unchecked_resumption:
+                        # Not a problem.
+                        got_size = os.stat(filename).st_size
+                    else:
+                        # Is a problem.
+                        os.remove(filename)
+                        got_size = None
+
+        self.got_size = got_size
```
snarayan21, line 181: nit: can you change this to `size` instead of `got_size`

streaming/storage/extra.py
```diff
@@ -276,73 +283,155 @@ def list_dataset_files(
...
-                        hashes: Optional[Dict[str, str]] = None) -> None:
-    """Download a file from the remote path to the local path, with size/hash checks.
+def smart_download_file(
+    *,
+    remote: Optional[str],
+    local: str,
+    timeout: Union[None, str, float] = 60,
+    retry: Union[None, str, int] = 2,
+    size: Union[None, str, int] = None,
+    max_size: Union[None, str, int] = '200mb',
```
snarayan21, line 293: 200MB seems a bit low?

streaming/stream/base.py
```diff
@@ -0,0 +1,116 @@
...
+    def await_index(self) -> None:
+        """Wait for the index file to become downloaded.
+
+        Notes:
+          * This method is called by StreamingDataset init.
+          * As the index download threads run in the background, all ranks loop over all Streams,
+            calling ``await_index()`` and ``load_index()`` on each Stream.
+          * This method is on the critical path.
+        """
+        wait_for_creation(self.local_index_path, self.download_timeout, 0.07)
```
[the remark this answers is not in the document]
snarayan21: probably meant to use TICK*10 from `constants.py` or something?

snarayan21 (changes requested):

streaming/stream/base.py
```diff
@@ -0,0 +1,116 @@
...
+        kwargs (Dict[str, Any]): Arguments inherted from StreamDirConf and StreamWeightConf.
+    """
+
+    def __init__(
+        self,
+        **kwargs: Any,
+    ) -> None:
+        StreamDirConf.__init__(self, **kwargs)
+        StreamWeightConf.__init__(self, **kwargs)
+        self.got_index_size: int
```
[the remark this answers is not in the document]
knighton: yeah, `got_index_size` is not known until after calling `download_index()`.

streaming/stream/weight_conf.py
```diff
@@ -0,0 +1,164 @@
...
+        cls,
+        streams: Sequence[Self],
+    ) -> Tuple[bool, bool]:
+        """Validate stream weights, returning whether relative or absolute weighting was used.
+
+        Args:
+            streams (Sequence[Stream]): Every stream comprising the dataset.
+
+        Returns:
+            bool: Whether streams are weighted relatively (proportionally).
```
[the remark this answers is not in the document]
knighton: Fixed.

streaming/format/base/file.py
```diff
@@ -0,0 +1,302 @@
...
+        return can_size + self.evict_phases(phase_dels)
+
+    def _load_raw(self) -> int:
+        """Sub-method to prepare up to the regular phase of this file.
+
+        Returns:
+            int: Delta disk usage, in bytes.
+        """
+        if self.raw_phase.is_local():
+            ddu = 0
```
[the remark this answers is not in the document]
knighton: `ddu = Delta disk usage` (line 248 and various places)
`du = Disk usage` (various places)

Named after Unix `du`

Thoughts on what better to call it? `delta_disk_usage` is too long, `d_disk_usage` is too asymmetric, `delta_usage` is too vague, `usage_change` is too unclear, etc, might as well come up with one term and use it uniformly

streaming/format/base/phase.py
```diff
@@ -0,0 +1,233 @@
...
+        hashes: Optional[Dict[str, str]] = None,
+    ) -> None:
+        if size is not None and size < 0:
+            raise ValueError(f'Shard file size must be a non-negative integer, but got: {size}.')
+
+        self.stream = stream
+        self.relative_path = relative_path
+        self.size = size
+        self.hashes = hashes or {}
+
```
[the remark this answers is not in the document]
knighton: Added comments.

streaming/format/base/phase.py
```diff
@@ -0,0 +1,233 @@
...
+            listing (Set[str]): Recursive dataset file listing.
+
+        Returns:
+            bool: Whether this phase is present or not in the listing.
+        """
+        filename = self.get_local_filename()
+        return filename in listing
+
+    def inventory_local(self, listing: Set[str]) -> Optional[int]:
+        """Initialize the given local directory wrf this shard file phase.
```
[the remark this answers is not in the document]
knighton: Intended `wrt`, now spelled it out.

streaming/format/base/writer/mono_row.py
```diff
@@ -0,0 +1,99 @@
...
+                Default to ``2``.
+    """
+
+    def __init__(self,
+                 *,
+                 out: Union[str, Tuple[str, str]],
+                 keep_local: bool = False,
+                 compression: Optional[str] = None,
+                 hashes: Optional[List[str]] = None,
+                 size_limit: Optional[Union[int, str]] = 1 << 26,
```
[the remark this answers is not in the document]
knighton: Fixed.

streaming/format/base/writer/dual_row.py
```diff
@@ -0,0 +1,108 @@
...
+    extra_bytes_per_shard = 0
+    extra_bytes_per_sample = 0
+
+    def __init__(self,
+                 *,
+                 out: Union[str, Tuple[str, str]],
+                 keep_local: bool = False,
+                 compression: Optional[str] = None,
+                 hashes: Optional[List[str]] = None,
+                 size_limit: Optional[Union[int, str]] = 1 << 26,
```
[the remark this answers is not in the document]
knighton: Fixed.

examples/multimodal/webvid/read.py
```diff
@@ -133,7 +133,7 @@ def get_item(self, idx: int) -> Any:
 
         return obj
 
-    def _download_thread(self, it: _Iterator) -> None:
+    def _fetch_thread(self, it: _Iterator) -> None:
```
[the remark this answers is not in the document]
knighton: Well, it technically downloads and/or unzips and/or canonicalizes. Calling it "fetching" re: that note on terminology.

examples/vision/cifar10/write.py
```diff
@@ -7,8 +7,8 @@
 
 from torchvision.datasets import CIFAR10
 
+from streaming.modality.vision import convert_image_class_dataset
```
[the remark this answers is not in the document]
knighton: Yeah, I figure that makes more sense than random top-level modules named after modalities.

streaming/format/base/phaser.py
knighton: Whether to delete (0 = false) x 3 phases (Zip, Raw, Can)

“Out of time and risk, just hardcode it to cache everything for now and actually implement this later”

streaming/format/base/file.py
knighton: Position is meaningful, None is a legitimate value

It’s not a list of the phases which are in use by this shard file instance, it’s a list of what this shard file instance has for each of the possible phases, in order

streaming/format/base/file.py
```diff
@@ -0,0 +1,302 @@
...
+            stream (StreamDirConf): The Stream that owns this Shard.
+        """
+        self.stream = stream
+        for phase in self.phases:
+            if phase:
+                phase.set_stream(stream)
+
+    def validate(self) -> None:
+        """Check whether this file is acceptable to be part of some Stream."""
+        # If we have a size limit on downloads, verify that it is not exceeded by our first phase.
```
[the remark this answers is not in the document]
knighton: It’s a guard rail to prevent people with gigantic Delta table parquet files DOSing themselves and blaming us

streaming/format/base/file.py
```diff
@@ -0,0 +1,302 @@
...
+
+        Args:
+            listing (Set[str]): Recursive dataset file listing.
+
+        Returns:
+            int: Disk usage.
+        """
+        # Collect this file's current cache usage, and the locality of each of its phases.
+        phase_locs = []
+        du = 0
```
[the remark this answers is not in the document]
knighton: Ideas on better names for “cache utilization attributed to phase(s) of this file”?

streaming/format/base/file.py
knighton: Yeah, my convention was du = absolute disk usage, ddu = delta disk usage (pos or neg)

streaming/dataset.py
knighton: This is on the critical path, loading many shards takes a long time, we need to get started ASAP, hence this way

There’s a faster, slightly more complicated way to do this in the PR desc but might save for future PR

streaming/util/json.py
knighton: I figure if we add more typing annotations stuff, like for args shorthand unions, it would live together with the other args shorthand code, so similarly json typing stuff ought to live in a json-specific source file

streaming/format/base/phaser.py
knighton: Using ndarray here because I’m doing numpy operations on the returned triples out of convenience and clear code… although that code was probably mostly cut from this PR

streaming/dataset.py
```diff
@@ -567,14 +574,14 @@ def __init__(self,
...
-                self.cache_usage += stream.get_index_size()
+                self.cache_usage += stream.got_index_size
 
             # Get cache usage due to shards.
             cache_usage_per_shard = np.zeros(self.num_shards, np.int64)
             for stream_id, stream in enumerate(self.streams):
                 begin = self.shard_offset_per_stream[stream_id]
                 end = begin + self.shards_per_stream[stream_id]
-                stream.set_up_local(self.shards[begin:end], cache_usage_per_shard[begin:end])
+                stream.inventory_local(cache_usage_per_shard[begin:end])
```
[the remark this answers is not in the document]
knighton: Yup, we don’t make any assumptions about the starting state of the local dir of each stream

streaming/dataset.py
knighton: Ah yea: https://docs.python.org/3.7/library/multiprocessing.html#multiprocessing.pool.Pool.join

Fixed in this commit: https://github.com/mosaicml/streaming/pull/585/commits/25c8f152dc2cd61e21edc97b8c517441aa5d94e6

Where I also add: `procs = SD._get_index_download_procs(procs, procs / cpu, max procs)`

Note: the Parquet PR is going to add indexing to SD init -- I suspect the above control on parallelism will get merged into that -- I wouldn't take this code as the last word

streaming/dataset.py
```diff
@@ -174,98 +174,112 @@ class StreamingDataset(Array, IterableDataset):
...
+            to ``None`` for no size limit. Set to ``str`` to specify the limit using Streaming
+            bytes shorthand. Set to ``int`` to specify bytes. Defaults to ``200mb``.
+        validate_hash (str | Sequence[str], optional): Ranked list of hashing algorithms to
+            apply if expected digest is available. Defaults to ``None``.
+        keep_phases (str | Sequence[str] | Dict[str, Optional[bool]] | Phaser): Which phases
+            to keep and to drop upon conversion, given either by intended use case or literally.
+            Specified as a single use or phase to keep, a sequence of uses or phases to keep, a
+            mapping of uses or phases to whether to keep or drop, or a ``Phaser`` (which performs
+            the same keeping or dropping). Defaults to ``None``.
+        index_download_procs (int, optional): Size of the process pool. You may set either this
```
karan6181, line 274: Based on our discussion, you will be removing the `index_download_procs,` `index_download_procs_per_cpu,` and `index_download_max_procs` from here and let SD infer it under the hood when you create the future PRs, right?

streaming/dataset.py
```diff
@@ -174,98 +174,112 @@ class StreamingDataset(Array, IterableDataset):
...
+            ``remote``/``local``. Defaults to ``None``.
+        remote (str, optional): Remote path to stream the dataset from. If ``None``, dataset must
+            be complete locally. Defaults to ``None``.
+        local (str, optional): Local working directory to stream the dataset to. Uses a
+            deterministically-calculated temp directory if not set. Defaults to ``None``.
+        split (str, optional): Which dataset split sub-path to use, if any. Defaults to ``None``.
+        index_size (str | int, optional): Expected index size in bytes. Defaults to ``None``.
+        index_hashes (Dict[str, str], optional): Available index hashes. This is a mapping of hash
+            algo name to expected hex digest. Defaults to ``None``.
+        allow_schema_mismatch (bool): If ``True``, continue if schemas mismatch across
```
karan6181, line 250: Can you explain a bit in the doc what the schema actually is?

simulation/core/sim_dataset.py
knighton: https://github.com/mosaicml/streaming/pull/585/commits/eb64e40ea18ea506160adeb4b39c1f0213470b08

streaming/dataset.py
```diff
@@ -478,7 +485,7 @@ def __init__(self,
         # Check that cache limit is possible.
         if cache_limit:
             self.cache_limit = normalize_bytes(cache_limit)
-            min_cache_usage = sum((stream.get_index_size() for stream in streams))
+            min_cache_usage = sum((stream.got_index_size for stream in streams))
```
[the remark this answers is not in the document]
knighton: https://github.com/mosaicml/streaming/pull/585/commits/eb64e40ea18ea506160adeb4b39c1f0213470b08

streaming/format/base/phase.py
knighton: https://github.com/mosaicml/streaming/pull/585/commits/21f06e0e1f35e85b8e1be63e14422d4d87ca640b

streaming/dataset.py
knighton: fetch_shard has its args in its docstring?

streaming/storage/extra.py
knighton: Will revisit in the future

streaming/stream/weight_conf.py
```diff
@@ -0,0 +1,164 @@
...
+            if not choose_per_epoch:
+                choose_per_epoch = sum(samples_per_stream)
+            proportion_per_stream = np.array([stream.proportion for stream in streams], np.float64)
+            proportion_per_stream /= proportion_per_stream.sum()
+            choose_per_stream = (choose_per_epoch * proportion_per_stream).astype(np.int64)
+            shortfall = choose_per_epoch - choose_per_stream.sum()
+            rng = np.random.default_rng(seed)
+            indices = rng.choice(len(streams), shortfall, False)
+            choose_per_stream[indices] += 1
+            repeat_per_stream = choose_per_stream / samples_per_stream
```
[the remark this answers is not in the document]
knighton: Currently, an exception is raised if samples_per_stream contains a zero, as earlily (?) as possible, at Stream init time, in StreamingDataset init, although that is somewhat removed from here

streaming/stream/base.py
```diff
@@ -0,0 +1,116 @@
...
+        Notes:
+          * This method is called by StreamingDataset init.
+          * As the index download threads run in the background, all ranks loop over all Streams,
+            calling ``await_index()`` and ``load_index()`` on each Stream.
+          * This method is on the critical path.
+
+        Returns:
+            List[Shared]: List of loaded Shards.
+        """
+        # Read the index file and parse its JSON.
```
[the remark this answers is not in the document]
knighton: Should I keep the previous exception catching and throwing? To me, the parse errors `json` provides are great and we would not be adding any value. Or am I misunderstanding something

streaming/format/base/phase.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def download(self) -> int:
+        """Download this phase.
+
+        Returns:
+            int: Delta disk usage, in bytes.
+        """
+        # Verify there is a remote to download from.
+        if self.stream.remote in {None, self.stream.local}:
+            raise ValueError(f'File(s) are missing, but there is nowhere to download them from ' +
```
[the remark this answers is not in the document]
knighton: What do we need to do here, esp about being the authoritative copy of the dataset aka remote

streaming/format/mds/shard.py
```diff
@@ -133,16 +185,18 @@ def decode_sample(self, data: bytes) -> Dict[str, Any]:
         """
         sizes = []
         idx = 0
-        for key, size in zip(self.column_names, self.column_sizes):
-            if size:
-                sizes.append(size)
+        for column in self.columns:
```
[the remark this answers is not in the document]
knighton: We could, but not as elegantly as doing it two passes, because the format is like `[sizes of dynamically sized fields only] [all fields]`, and I really doubt it would make a perf difference? But worth benchmarking in future work.

streaming/dataset.py
knighton: No, I said these args will be folded into the arg handling for Parquet dataset indexing in the upcoming PR, and there are different possibilities for how to do this correctly. We will not be simply dropping this functionality

streaming/format/base/file.py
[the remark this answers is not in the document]
knighton: https://github.com/mosaicml/streaming/pull/585/commits/44fd90ef7e16e63ba117db74eb215bccae07aa8b

streaming/format/base/file.py
knighton: https://github.com/mosaicml/streaming/pull/585/commits/44fd90ef7e16e63ba117db74eb215bccae07aa8b

streaming/dataset.py
knighton: Also, unless overridden, the parallelism is already inferred in this PR, but by python multiprocessing, up to the limit on concurrent connections

streaming/stream/base.py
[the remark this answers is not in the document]
knighton: Added a constant `FILESYSTEM_POLL_INTERVAL`.

But I wonder if these should be made stochastic instead, e.g. `m + abs(normal(0, s))`.

streaming/stream/weight_conf.py
knighton: Fixed

streaming/format/base/phase.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def validate_for_download(self) -> None:
+        """Validate for download purposes.
+
+        The download phase (``zip`` or ``raw``) of a shard file must be sized, and if there is a
+        limit on download size, must fit the limit.
+        """
+        if self.size is None:
+            raise ValueError(
+                f'The first existing phase, i.e. the phase that is stored persistently and ' +
```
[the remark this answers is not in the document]
knighton: Improved docstring?

streaming/format/base/phase.py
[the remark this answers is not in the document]
knighton: How about this?

streaming/dataset.py
karan6181: Are these args user-customizable? Can we not set defaults based on the instance that the user uses? I am not convinced that we need 3 user-facing args just to tackle the index.json file.

streaming/format/base/shard/dual_row.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import Any, Optional
+
+from streaming.format.base.file import ShardFile
+from streaming.format.base.shard.row import RowShard
+from streaming.stream.dir_conf import StreamDirConf
+
+__all__ = ['DualRowShard']
+
+
+class DualRowShard(RowShard):
```
karan6181, line 15: Question: What SD data format would use `DualRowShard` and/or `MonoRawShard`?

streaming/storage/extra.py
```diff
@@ -67,9 +35,46 @@ def _normalize_path(path: str) -> Tuple[str, bool]:
         path = obj.path
     else:
         is_local = False
+
+    path = os.path.normpath(path)
+
+    idx = path.find(':/')
```
karan6181, line 41: Does this logic also apply to DBFS and UC Volume file paths where their path is `dbfs:/Volumes/<catalog-name>/<schema-name>/<volume-name>/path`?

streaming/stream/base.py
```diff
@@ -0,0 +1,118 @@
...
+from streaming.stream.dir_conf import StreamDirConf
+from streaming.stream.weight_conf import StreamWeightConf
+from streaming.util.json import JSONDict
+from streaming.util.waiting import wait_for_creation
+
+__all__ = ['Stream']
+
+
+class Stream(StreamDirConf, StreamWeightConf):
+    """A StreamingDatasdet directory, used alone or together with others.
```
karan6181, line 25:
```suggestion
    """A StreamingDataset directory, used alone or together with others.
```

tests/test_writer.py
```diff
@@ -22,7 +22,7 @@ def test_invalid_args(self, local_remote_dir: Tuple[str, str]):
         local, _ = local_remote_dir
         dataset = SequenceDataset(100)
         columns = dict(zip(dataset.column_names, dataset.column_encodings))
-        with pytest.raises(ValueError, match=f'.*Invalid Writer argument.*'):
+        with pytest.raises(ValueError):
```
karan6181, line 25: Is there any specific reason for removing `match`? Valuing against the exception message would be good to ensure we capture the correct exception.

tests/test_stream.py
```diff
@@ -76,5 +90,5 @@ def test_missing_index_json_local(local_remote_dir: Any):
     else:
         raise Exception(f"Missing {os.path.join(remote_dir, 'index.json')}")
     stream = Stream(remote=None, local=remote_dir)
-    with pytest.raises(RuntimeError, match='No `remote` provided, but local file.*'):
+    with pytest.raises(ValueError):
```
karan6181, line 93: Is there any specific reason for removing `match`? Valuing against the exception message would be good to ensure we capture the correct exception.

streaming/format/base/shard/dual_row.py
knighton: to quote PR desc part 3.6, the "what":

```
Writer
 └─ RowWriter
     ├─ DualRowWriter
     │    ├─ JSONLWriter
     │    └─ XSVWriter
     │        ├─ CSVWriter
     │        └─ TSVWriter
     └─ MonoRowWriter
         └─ MDSWriter
```

Writers write Streaming dataset directories, one shard at a time.

```
Shard
 └─ RowShard
     ├─ DualRowShard
     │   ├─ JSONLShard
     │   └─ XSVShard
     │       ├─ CSVShard
     │       └─ TSVShard
     └─ MonoRowShard
         └─ MDSShard
```

But more importantly, the "why":

I've seen github traffic patterns indicating that our previous terminology of `Joint*` and `Split*` (e.g., `SplitWriter`, `SplitReader`) was leading to people getting confused and thinking they are about dataset splits, when they are really about the file(s) that a shard's information is split over...

XiaohanZhangCMU (dismissed):
....

tests/test_writer.py
knighton: Fixed

tests/test_stream.py
knighton: Fixed

streaming/storage/extra.py
knighton: How bout now?

streaming/dataset.py
knighton: Removed x3 per discussion

streaming/dataset.py
knighton: How about this?

I figure `schema` is the more general/more proper terminology over `columns`, but in our case it's just referring to our columns...

streaming/dataset.py
```diff
@@ -174,98 +174,99 @@ class StreamingDataset(Array, IterableDataset):
...
+                  * ``allow_unchecked_resumption``
+              * Downloads:
+                  * ``download_retry``
+                  * ``download_timeout``
+                  * ``download_max_size``
+                  * ``validate_hash``
+                  * ``keep_phases``
+      * How to iterate:
+          * Init:
+              * ``index_download_procs``
```
karan6181, line 208: nit: I think you forgot to remove `index_download_procs`, `index_download_procs_per_cpu`, and `index_download_max_procs`, from here.

XiaohanZhangCMU (commented):
LGTM.

streaming/stream/dir_conf.py
```diff
@@ -0,0 +1,469 @@
...
+                encountered in shard(s). If ``False``, raises if unsafe type(s) encountered. Set to
+                ``Auto`` to not inherit from StreamingDataset. Defaults to ``Auto()``.
+            allow_unchecked_resumption (bool | Auto): If ``True``, upon resume, accept and use
+                shard file phases that we are unable to check the size/hash(es) of. If ``False``,
+                upon resume, drop such files, to regenerate on the fly when needed. Set to ``Auto``
+                to not inherit from StreamingDataset. Defaults to ``Auto()``.
+            download_retry (stgr | int | Auto): Number of download re-attempts before raising an
+                error. Set to ``Auto`` to not inherit from StreamingDataset. Defaults to
+                ``Auto()``.
+            download_timeout (None | str | float | Auto): Time to wait for a file download to
```
karan6181, line 365: Sorry if I misunderstood the logic here; if the user does not provide the `download_timeout`, what would be the default value? Our current (streaming v0.7.4) default value is `60` seconds.

tests/test_eviction.py
```diff
@@ -120,9 +121,13 @@ def cache_limit_too_low(remote: str, local: str, keep_zip: bool):
     rmtree(local, ignore_errors=False)
 
 
+# TODO: re-enable `shard_eviction` and `cache_limit_too_low` when cache_limit is returned to fully
```
karan6181, line 124: Could you let me know if you plan to enable this as part of Delta Table feature support?

tests/test_reader.py
```diff
@@ -144,15 +144,14 @@ def test_reader_download_fail(local_remote_dir: Any, missing_file: str, seed: in
...
-    with pytest.raises(FileNotFoundError) as exc_info:
+    with pytest.raises(RuntimeError):
         dataset = StreamingDataset(local=local_dir,
                                    remote=remote_dir,
                                    shuffle=False,
                                    download_timeout=1,
                                    shuffle_seed=seed)
         for _ in dataset:
             pass
-    assert exc_info.match(r'.*No such file or directory*')
```
karan6181, line 155: Is there any reason why we are not validating the error message?

knighton: Fixed

tests/test_eviction.py
knighton: Almost certainly, but I want to punt these to a future PR in order to maintain velocity.

`cache_limit_too_low` for example relies on `cache_limit` being inviolable. This PR is written with `cache_limit` being briefly violable during decompression and canonicalization, which is more or less unavoidable because we don't know the resulting size of Parquet -> MDS .mds files in advance, and if we end up letting the user intervene in column physical types, which very well will likely be requested down the road if we don't do it as part of this MVP (I cut the relevant arg already), .mds size could be dependent on that, etc

streaming/stream/dir_conf.py
knighton: The `Auto` jive lets us define the default value in exactly one place (in dataset.py line 254)

I upped `download_timeout` to `'2m'` in light of:
1. All the yamls I see specifically raising it to several to a handful of minutes
2. Assuming our users have network connections slower than us, or as fast as us, but not faster, so if anything we should be raising it for their UX and setting it lower for ourselves
3. Assuming users can tolerate waiting a bit longer in the case of download timeout errors
4. Assuming the incoming Parquet indexing PR which will have to download potentially large Parquet shards will use the same download timeout, because otherwise we would be adding args etc

knighton: In the future, I wonder if it would make more sense to create a `streaming/defaults.py` or use `streaming/constant(s).py` for such definitions...

streaming/dataset.py
knighton: Fixed thx

snarayan21 (commented):
some testing issues in addition to these comments

streaming/dataset.py
```diff
@@ -475,21 +486,36 @@ def __init__(self,
         self.stream_per_shard = np.array(stream_per_shard, np.int64)
         self.num_shards = len(self.shards)
 
+        # Wait for the pool workers (stream index download processes) to finish.
+        if pool is not None:
```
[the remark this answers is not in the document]
knighton: The join cleans up the child processes

scripts/long_lines.py
```diff
@@ -166,25 +171,43 @@ def main(args: Namespace) -> int:
         txt = ', '.join(sorted(non_text_behaviors))
         raise ValueError(f'Unknown non-text behavior (must be one of: {txt}): {args.non_text}.')
 
-    count = 0
+    pairs = []
```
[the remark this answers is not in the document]
knighton: Oh yeah, can different PR this

streaming/format/base/shard/base.py
```diff
@@ -0,0 +1,150 @@
...
+
+class Shard(Array):
+    """Streaming shard abstract base class.
+
+    Args:
+        conf (Any, optional): JSON shard config. Defaults to ``None``.
+        stream (StreamDirConf): Link back up to the Stream that owns this shard, from which we
+            get arguments which are shared across all shards like remote/local paths. Avoids an
+            import cycle by Stream subclassing StreamDirConf.
+        num_samples (int): Number of samples in this shard.
```
snarayan21, line 26: can you add `files` to this docstring?

streaming/dataset.py
```diff
@@ -311,44 +312,49 @@ class StreamingDataset(Array, IterableDataset):
         allow_unsafe_types (bool): If a shard contains Pickle, which allows arbitrary code
```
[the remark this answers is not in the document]
knighton: thanks

streaming/format/base/shard/base.py
knighton: fixed

karan6181 (approved):

PR #587, opened 2024-02-01
Change comparison in partitions to include equals

JAEarly:
## Description of changes:

A warning is raised during partitioning when the number of canonical nodes is equal to the number of samples. It is my understanding that this should be valid (one sample per node), and that issues only occur when the number of samples is less than the number of canonical nodes. Apologies if my understanding is incorrect here. 

The simple fix is to the change the comparison to be `>=` rather than `>`

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/partition/orig.py
+++ b/streaming/base/partition/orig.py
@@ -81,8 +81,12 @@ def get_partitions_orig(num_samples: int,
             padding = node_ratio - overflow
     padded_samples_per_canonical_node = samples_per_canonical_node + padding
 
+    # For samples to be properly split across canonical nodes, there must be more samples than nodes.
+    # The edge case is when the number of samples is equal to the number of canonical nodes, but this only works when
+    #  there is an equal or greater number of canonical nodes than physical nodes.
+    # If these conditions are not met, an alternative sampling approach is used that leads to many repeats.
     if num_samples > num_canonical_nodes or (num_samples == num_canonical_nodes and
-                                             num_canonical_nodes > num_physical_nodes):
+                                             num_canonical_nodes >= num_physical_nodes):
         # Create the initial sample ID matrix.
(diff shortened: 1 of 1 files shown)


JAEarly:
I didn't realise this would break the tests (I made the change in the browser). If this change makes sense I'll happily go and alter the tests as appropriate.

JAEarly:
I've now updated the code such that the tests pass but the warning is not raised for the edge case: num_canonical_nodes == num_samples and num_canonical_nodes > num_physical_nodes.

streaming/base/partition/orig.py
```diff
@@ -81,7 +81,8 @@ def get_partitions_orig(num_samples: int,
             padding = node_ratio - overflow
     padded_samples_per_canonical_node = samples_per_canonical_node + padding
 
-    if num_samples > num_canonical_nodes:
+    if num_samples > num_canonical_nodes or (num_samples == num_canonical_nodes and
+                                             num_canonical_nodes > num_physical_nodes):
```
snarayan21, line 85: I believe the condition here can actually be slightly expanded to `num_canonical nodes >= num_physical nodes`. So the full conditional would be:
```
if num_samples > num_canonical_nodes or (num_samples == num_canonical_nodes and num_canonical_nodes >= num_physical_nodes):
```
Mind making this change and adding an explanatory comment right before the if statement?

snarayan21 (commented):
Hey! Apologies for getting to this late, but just a small change and should be good to go. Great catch!

streaming/base/partition/orig.py
JAEarly: Thank you for the input; I've made that change now.

snarayan21 (approved):
lgtm!

PR #593, opened 2024-02-07
Modifications to resumption shared memory allowing `load_state_dict` multiple times.

snarayan21:
## Description of changes:
Previously, calling `load_state_dict` multiple times with state dicts of varying lengths was not possible since the shared memory was of a fixed size. This is useful, for example, in curriculum learning -- we need to update the dataset state dict again after the initial checkpoint loading in order to correctly load in a new dataset. This PR takes the approach of hardcoding the resumption state dict size to a maximum of `mmap.PAGESIZE` bytes (which is at minimum 4096 bytes, [link here](https://en.wikipedia.org/wiki/Page_(computer_memory)#Multiple_page_sizes)). We also append a null byte, `'\0'`, to the end of the loaded resumption state dict so that it can be correctly set multiple times, and loaded in correctly on resumption. Given that the current dataset state dict encoded as a `utf-8` string is 112 bytes (of characters), a state dict that uses up `mmap.PAGESIZE` bytes can really only be set by mistake.

I attempted a solution that cleaned up and re-created the SharedMemory object but this did not work even though I tried a variety of approaches. It may also be possible to use a shared list but this would also require changes in the `_resume` function. This is the simplest change to enable curriculum learning (dependent, in the PR linked [here](https://github.com/mosaicml/llm-foundry/pull/954)).

Unit tests were added to ensure that loading and saving the state dict both once and multiple times worked, as well as testing for an error to be thrown when loading in a state dict that is too large. Testing of curriculum learning is in the foundry PR linked above.

## Merge Checklist:

### General

### Tests


streaming/base/dataset.py
```diff
@@ -804,12 +804,27 @@ def load_state_dict(self, obj: Dict[str, Any]) -> None:
...
+        # and replacing it with a new one is causing great difficulties.
+
         name = _get_path(self._shm_prefix_int, RESUME)
-        data = json.dumps(obj, sort_keys=True).encode('utf-8')
+        data = json.dumps(obj, sort_keys=True)
+
+        len_needed = len(data)
+        if len_needed > RESUME_SHM_SIZE:
+            raise ValueError(f'State dict for resumption is too large to fit in shared memory. ' +
+                             f'Please reduce the state dict size to less than {RESUME_SHM_SIZE} ' +
```
knighton, line 818: More like "Please increase the state dict shared memory buffer size to be at least {len(data) + 1} bytes"

streaming/base/dataset.py
```diff
@@ -804,12 +804,27 @@ def load_state_dict(self, obj: Dict[str, Any]) -> None:
...
+                             f'bytes. Current size is {len_needed} bytes.')
         # Some platforms choose to allocate chunks of memory based upon that platform's memory page
         # size, hence the exact size of the shared memory block that was returned may be larger
         # than what was requested.
-        self._resume_shm = SharedMemory(name=name, size=len(data))
+        self._resume_shm = SharedMemory(name=name, size=RESUME_SHM_SIZE)
+        # Write a null byte at the end of the shared memory block so that we read in the state
+        # dict correctly in `_resume`.
+        data += '\0'
```
knighton, line 826: Pedantic nit: += is linear time because str is immutable, right?

```
shm[:len(data)] = data
shm[len(data)] = 0
```

streaming/base/constant.py
```diff
@@ -31,3 +31,6 @@
 
 # Time to wait, in seconds.
 TICK = 0.007
+
+# Resumption shared memory size
+RESUME_SHM_SIZE = 1024
```
knighton, line 36: I have a good feeling about 1337 bytes

streaming/base/dataset.py
snarayan21: Tried it, this throws an error: `memoryview: invalid type for format 'B'` :(
I tried encoding the null byte separately to utf-8, and adding that in as well, to no avail. So just went with this.

streaming/base/constant.py
snarayan21: lmao

streaming/base/dataset.py
snarayan21: changed this!

streaming/base/dataset.py
snarayan21: Example:
```>>> string = "hello"
>>> stringbytes = string.encode('utf-8')
>>> mem.buf[:len(string)] = stringbytes
>>> mem.buf[len(string)] = '\0'.encode('utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: memoryview: invalid type for format 'B'
```

streaming/base/dataset.py
```diff
@@ -698,6 +699,9 @@ def _resume(self, world: World, epoch: int) -> Tuple[int, int]:
         buf = buf[:index] if index != -1 else buf
         obj = json.loads(buf.decode('utf-8'))
 
+        print('READ IN STATE DICT:')
```
karan6181, line 702: nit: please remove print statements

streaming/base/dataset.py
```diff
@@ -804,12 +808,33 @@ def load_state_dict(self, obj: Dict[str, Any]) -> None:
...
         name = _get_path(self._shm_prefix_int, RESUME)
-        data = json.dumps(obj, sort_keys=True).encode('utf-8')
+        data = json.dumps(obj, sort_keys=True)
+
+        len_needed = len(data)
+        # Note: mmap.PAGESIZE has a minimum size of 4096 bytes across systems. For reference,
+        # see the link below:
+        # https://en.wikipedia.org/wiki/Page_(computer_memory)#Multiple_page_sizes
+        if len_needed > mmap.PAGESIZE:
+            raise ValueError(
```
karan6181, line 824: How would the user solve this error message? What steps user would have to take?

streaming/base/dataset.py
snarayan21: resolved

streaming/base/dataset.py
snarayan21: resolved after discussion, added clarifying statement about potential corrupted state dict

karan6181 (commented):
LGTM. Thank You!

PR #594, opened 2024-02-07
Allow writers to overwrite existing data

JAEarly:
## Description of changes:

CloudUploader has an `exist_ok` kwarg that is set to False by default.  
Writers (e.g. MDSWriter) cannot overwrite this.  

This PR makes two changes:
1. `exist_ok` can now be set by a writer.
2. If `exist_ok` is True, existing files are now deleted. This allows easy debugging when developing streaming datasets as the files do not have to be manually deleted each time the code is run.

Tests and docs updated accordingly.

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/format/base/writer.py
+++ b/streaming/base/format/base/writer.py
@@ -237,10 +237,6 @@ class Writer(ABC):
             'size_limit': self.size_limit
         }
 
-    def clear_local(self) -> None:
-        """Remove the local directory, allowing it to be written to."""
-        shutil.rmtree(self.local)
-
     @abstractmethod
     def flush_shard(self) -> None:
         """Flush cached samples to storage, creating a new shard."""
--- a/streaming/base/format/base/writer.py
+++ b/streaming/base/format/base/writer.py
@@ -100,7 +100,8 @@ class Writer(ABC):
 
         # Validate keyword arguments
         invalid_kwargs = [
-            arg for arg in kwargs.keys() if arg not in ('progress_bar', 'max_workers', 'retry')
+            arg for arg in kwargs.keys()
+            if arg not in ('progress_bar', 'max_workers', 'retry', 'exist_ok')
         ]
         if invalid_kwargs:
             raise ValueError(f'Invalid Writer argument(s): {invalid_kwargs} ')
@@ -117,7 +118,8 @@ class Writer(ABC):
         self.shards = []
 
         self.cloud_writer = CloudUploader.get(out, keep_local, kwargs.get('progress_bar', False),
-                                              kwargs.get('retry', 2))
+                                              kwargs.get('retry', 2),
+                                              kwargs.get('exist_ok', False))
         self.local = self.cloud_writer.local
         self.remote = self.cloud_writer.remote
         # `max_workers`: The maximum number of threads that can be executed in parallel.
--- a/streaming/base/format/base/writer.py
+++ b/streaming/base/format/base/writer.py
@@ -64,6 +64,9 @@ class Writer(ABC):
                 file to a remote location. Default to ``min(32, (os.cpu_count() or 1) + 4)``.
             retry (int): Number of times to retry uploading a file to a remote location.
                 Default to ``2``.
+            exist_ok (bool): If the local directory exists and is not empty, whether to overwrite
+                the content or raise an error. `False` raises an error. `True` deletes the
+                content and starts fresh. Defaults to `False`.
     """
 
     format: str = ''  # Name of the format (like "mds", "csv", "json", etc).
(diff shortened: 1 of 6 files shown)


streaming/base/storage/upload.py
```diff
@@ -171,8 +171,9 @@ def __init__(self,
                 raise FileExistsError(f'Directory is not empty: {self.local}')
```
karan6181, line 171: Can you please extend the error message saying that you set `exist_ok=True` if you would like to delete the content from the local directory and start fresh?

streaming/base/format/base/writer.py
```diff
@@ -100,7 +100,8 @@ def __init__(self,
 
         # Validate keyword arguments
         invalid_kwargs = [
-            arg for arg in kwargs.keys() if arg not in ('progress_bar', 'max_workers', 'retry')
+            arg for arg in kwargs.keys()
+            if arg not in ('progress_bar', 'max_workers', 'retry', 'exist_ok')
```
karan6181, line 104: Can you please add the args comment for `exist_ok` at the start of the method? For example, something like below or in your own words
```
exist_ok (bool): If the local directory exists and is not empty, whether to overwrite
                the content or raise an error. `False` raises an error. `True` deletes the
                content and starts fresh. Defaults to `False`.
```

Also, do the same for all `writer.py` file in directory https://github.com/mosaicml/streaming/tree/main/streaming/base/format.

karan6181 (commented):
Thank you for creating the PR. The changes looks good to me. I have posted some minor comments.

@XiaohanZhangCMU Can you please also take a look?

streaming/base/storage/upload.py
JAEarly: Thanks for the quick response! Done.

JAEarly:
Looking at the tests that are failing: I believe there's now a conflict when `exists_ok=True` and `keep_local=True`, or at least this seems to be what the GitHub test logs says. I'm having issues running the failing tests locally so I'm unable to debug. @karan6181 do you have further insight on this?

XiaohanZhangCMU:
@JAEarly Currently exists_ok=True in clouduploader.py is only used by merging index (streaming/base/utils/merge_index). Your change will make that fail. 

I suggest you add the "file-removing behavior" to mdswriter instead of upload.py.

streaming/base/storage/upload.py
```diff
@@ -168,11 +168,14 @@ def __init__(self,
...
-                raise FileExistsError(f'Directory is not empty: {self.local}')
+                raise FileExistsError(
+                    f'{self.local} not empty. Set exist_ok to True to overwrite the current contents (will delete).'
+                )
             else:
                 logger.warning(
-                    f'Directory {self.local} exists and not empty. But continue to mkdir since exist_ok is set to be True.'
+                    f'Directory {self.local} exists and is not empty; exist_ok is set to True so will remove contents.'
                 )
+                shutil.rmtree(self.local)
```
XiaohanZhangCMU, line 178: This will break all the other code that does not perform writing.

XiaohanZhangCMU (changes requested):

streaming/base/storage/upload.py
JAEarly: Thank you for the input, I'll change the implementation now.

JAEarly: Okay, I've moved the removal code into writer which should fix the issue.

XiaohanZhangCMU (approved):
Lgtm. Wait for ci checks /

JAEarly:
Should be good to merge, I've rebased just need the CI checks to be run again, thanks!

streaming/base/format/base/writer.py
```diff
@@ -116,6 +120,13 @@ def __init__(self,
 
         self.shards = []
 
+        # Remove local directory if requested prior to creating writer
+        local = out if isinstance(out, str) else out[0]
+        if os.path.exists(local) and kwargs.get('exist_ok', False):
```
XiaohanZhangCMU, line 125: shouldn't this be kwargs.get('exist_ok', True)?

XiaohanZhangCMU (changes requested):

streaming/base/format/base/writer.py
JAEarly: I would say the directory should not be removed by default - files should not be deleted unless the user has explicitly set `exist_ok` to True.

JAEarly: My understanding is:
- If `local` exists but `exist_ok` is False (default value), then `local` will not be removed and `CloudUploader` will throw an error. Avoids accidentally overwriting data.
- If `local` doesn't exist then nothing needs deleting and `CloudUploader` will work fine.
- If `local` exists and `exist_ok` is True (has to be set by the user), then `local` will be removed with a warning and `CloudUploader` will work fine.

XiaohanZhangCMU (approved):
LGTM! Thanks for the improvement!

PR #597, opened 2024-02-09
Replicating samples across devices (SP / TP enablement)

knighton:
### 1. `World`

- Streaming's model of the world/the run/the job is (num nodes, ranks per node) rank processes + (num nodes, ranks per node, workers per rank) worker processes.
- For example, 8 ranks, 8 workers/rank -> 8 + 8 * 8 = 72 StreamingDataset replicas per node.
- Currently they coordinate via filelocks and shared memory.
- A `World` is just our wrapper around `torch.dist` and `torch.utils.data.get_worker_info`.
- The `World` just tells you which one you are, out of how many (nodes, ranks/node, workers/rank).
- There's always only one `World`, semantically speaking.
- However, when in a rank, `get_worker_info()` will say we are worker 0 of 1. One has to keep that in mind.

### 2. `StreamingDataset` argument `replication: Optional[int]`

- `replication` iterates the same samples for groups of adjacent GPUs (ranks).
- It does this by scaling down (in number) and repeating ranks.
- But the underlying data doesn't change, so we are correspondingly scaling up the ranks (in length).
- It prefers to merge ranks within the same node first, then if it needs to merge further, across node boundaries.

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -308,6 +308,10 @@ class StreamingDataset(Array, IterableDataset):
         allow_unsafe_types (bool): If a shard contains Pickle, which allows arbitrary code
             execution during deserialization, whether to keep going if ``True`` or raise an error
             if ``False``. Defaults to ``False``.
+        world (World, optional): Override the world state (otherwise it is detected from standard
+            env vars). Defaults to ``None``.
+        tensor_parallelism (int, optional): Tensor parallelism to apply to the given or detected
+            world state. Defaults to ``None``.
     """
 
     def __init__(self,
@@ -333,7 +337,9 @@ class StreamingDataset(Array, IterableDataset):
                  shuffle_seed: int = 9176,
                  shuffle_block_size: Optional[int] = None,
                  batching_method: str = 'random',
-                 allow_unsafe_types: bool = False) -> None:
+                 allow_unsafe_types: bool = False,
+                 world: Optional[World] = None,
+                 tensor_parallelism: Optional[int] = None) -> None:
         # Global arguments (which do not live in Streams).
         self.predownload = predownload
         self.cache_limit = cache_limit
@@ -348,6 +354,16 @@ class StreamingDataset(Array, IterableDataset):
         self.shuffle_block_size = shuffle_block_size
         self.batching_method = batching_method
         self.allow_unsafe_types = allow_unsafe_types
+
+        # Initialize the World context.
+        #
+        # Beware: This information is for the per-rank process. DataLoader worker processes may see
+        # different values for these fields. We are saving the rank World here because we cannot
+        # instantiate a World inside the StreamingDataset destructor.
+        world = world or World.detect()
+        if tensor_parallelism is not None:
+            world = world.tensor_parallel(tensor_parallelism)
+        self._rank_world = world
 
         # Initialize initial_physical_nodes to None. If we are resuming, then we will set it to the
         # number of physical nodes of the initial run in the _resume function.
@@ -442,13 +458,6 @@ class StreamingDataset(Array, IterableDataset):
         # Set streams.
         self.streams = streams
         self.num_streams = len(streams)
-
-        # Initialize the World context.
-        #
-        # Beware: This information is for the per-rank process. DataLoader worker processes may see
-        # different values for these fields. We are saving the rank World here because we cannot
-        # instantiate a World inside the StreamingDataset destructor.
-        self._rank_world = world = World()
 
         # Download each stream's index, load their shards, and map streams <-> shards.
         self.num_samples = 0
@@ -771,7 +780,7 @@ class StreamingDataset(Array, IterableDataset):
         Returns:
             Dict[str, Any]: The state.
         """
-        world = World()
+        world = self._rank_world
         epoch = self.next_epoch - 1
         epoch, offset = self._resume(world, epoch)
         if from_beginning:
@@ -1437,7 +1446,7 @@ class StreamingDataset(Array, IterableDataset):
 
         # Discover where we left off, if there is a checkpoint, or start at the next epoch.
         # Also pre-increment the epoch counter.
-        world = World()
+        world = self._rank_world.detect_workers()
         epoch, sample_in_epoch = self._resume_incr_epoch(world)
 
         # Get this worker's partition of samples to process.
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -308,10 +308,9 @@ class StreamingDataset(Array, IterableDataset):
         allow_unsafe_types (bool): If a shard contains Pickle, which allows arbitrary code
             execution during deserialization, whether to keep going if ``True`` or raise an error
             if ``False``. Defaults to ``False``.
-        world (World, optional): Override the world state (otherwise it is detected from standard
-            env vars). Defaults to ``None``.
-        tensor_parallelism (int, optional): Tensor parallelism to apply to the given or detected
-            world state. Defaults to ``None``.
+        replication (int, optional): Determines how many consecutive devices will receive the same
+            samples. Useful for training with tensor or sequence parallelism, where multiple
+            devices need to see the same partition of the dataset. Defaults to ``None``.
     """
 
     def __init__(self,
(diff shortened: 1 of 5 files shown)


streaming/base/dataset.py
```diff
@@ -308,6 +308,10 @@ class StreamingDataset(Array, IterableDataset):
         allow_unsafe_types (bool): If a shard contains Pickle, which allows arbitrary code
             execution during deserialization, whether to keep going if ``True`` or raise an error
             if ``False``. Defaults to ``False``.
+        world (World, optional): Override the world state (otherwise it is detected from standard
+            env vars). Defaults to ``None``.
+        tensor_parallelism (int, optional): Tensor parallelism to apply to the given or detected
```
vchiley, line 313: Does this mean that the same sample is provided to multiple (consecutive) GPUs?
ie do all gpus in the `world` get the same sample?
Can this be documented so the user knows?

streaming/base/world.py
vchiley, line 1: What is the world state?
Is it analogous to a [process group](https://pytorch.org/docs/stable/distributed.html#torch.distributed.new_group)?
or is it analogous to a [device_mesh of process groups](https://pytorch.org/docs/stable/distributed.html#torch.distributed.device_mesh.init_device_mesh)?

vchiley, line 1: Can I pass in a world state of any size or are there some limitation?
(Asking because many times TP operates within a Node, I'm trying to make sure we do not have that assumption here.) 
I'd like to create worlds of size 2 or 8 or 16 or 48 or ...

streaming/base/dataset.py
vchiley: or is it that samples across worlds are repeated?
ie rankN in all worlds produces the same sample

knighton: > Does this mean that the same sample is provided to multiple (consecutive) GPUs?

Yes

>  ie do all gpus in the world get the same sample?

Nah, just all (`tensor_parallelism`)-long shingles of GPUs

> or is it that samples across worlds are repeated?

Basically `World` is just Streaming's wrapper around node/rank env vars and `torch.utils.data.get_worker_info()`, telling you which one of those you are, there's just one of them, basically.

streaming/base/world.py
knighton: Device mesh right? It doesn't do any communication or state for communication, for which Streaming has various solutions depending on the problem.

knighton: PR is now updated to do tensor parallel across both nodes and ranks, preferring ranks.

streaming/base/world.py
```diff
@@ -37,28 +40,109 @@ class World:
...
+        if 0 <= ratio:
+            raise ValueError(f'Tensor parallelism ratio must be postiive.')
+
+        if self.num_ranks % ratio:
+            raise ValueError(f'World size must be divisible by your tensor parallelism ratio.')
+
+        rank = self.rank // ratio
+        num_ranks = self.num_ranks // ratio
+        worker = rank * self.workers_per_rank + self.worker_of_rank
+        if self.ranks_per_node <= num_ranks:
```
snarayan21, line 122: @knighton mind adding some comments to the code here? I'm trying to understand what modifications are happening, and I _think_ there may be a bug or a better way of doing this, but it's pretty hard for me to reason about without completely understanding what you're trying to accomplish here.

iiuc `ratio` is the number of consecutive GPUs that should be sharing the same samples -- and should have the same `World` information. `num_ranks` here is the # of TP blocks, and `rank` is the TP block index. I'm not seeing why we need to check if `self.ranks_per_node <= num_ranks`, and it seems if we set `num_nodes` to 1, then we'll have download duplication across all the nodes which will be pretty bad. The TP ratio should be the degree of duplication of the `World` objects and I'm not sure that this does that entirely correctly...

snarayan21 (commented):
some clairfying questions

streaming/base/world.py
knighton: > iiuc ratio is the number of consecutive GPUs that should be sharing the same samples -- and should have the same World information. num_ranks here is the # of TP blocks, and rank is the TP block index.

Exactly (for iterating purposes, modulo that bug for coordinating purposes which am about to fix)

> I'm not seeing why we need to check if `self.ranks_per_node <= num_ranks`, and it seems if we set num_nodes to 1, then we'll have download duplication across all the nodes which will be pretty bad.

PR originally applied tensor parallelism intra-node only, but then iiuc VC noted TP blocks may be inter-node here: https://github.com/mosaicml/streaming/pull/597#discussion_r1484550537

knighton: Spelling out the logic for my own benefit:

```
If num TP blocks exceeds true local world size:
    Scale down perceived num nodes to TP blocks / true local world size
    Don't need to scale down its perceived local world size because we scaled nodes
else:
    Just one perceived node, and what happens in the node, stays in the node
    Map ranks to potentially fewer TP blocks by potentially scaling down its peceived local world size
```

knighton: Could rewrite it to:

```
num_nodes = (num_ranks + self.ranks_per_node - 1) // self.ranks_per_node
ranks_per_node = num_ranks // num_nodes
```

knighton: > modulo that bug for coordinating purposes which am about to fix

https://github.com/mosaicml/streaming/pull/597/commits/a17419eafbac1669e02779511c5e2470c4da1758

streaming/base/world.py
```diff
@@ -37,28 +40,118 @@ class World:
...
+    def tensor_parallel(self, ratio: int) -> Self:
+        """Get a copy of this world state with the given tensor paralellism.
+
+        Args:
+            ratio (int): Ratio of tensor parallelism.
+
+        Returns:
+            Self: A new tensor-parallel version of this World state object.
+        """
+        if 0 <= ratio:
```
andreamad8, line 126: should this be:
```
if 0 >= ratio:
```

knighton: Tyks!

andreamad8:
Now I get an error like:
```
  File "/streaming/streaming/base/dataset.py", line 1458, in __iter__
    sample_ids = self._get_work(world, epoch, sample_in_epoch)
  File "/streaming/streaming/base/dataset.py", line 1002, in _get_work
    shape_shm, data_shm = self._share_work(epoch_sample_ids)
  File "/streaming/streaming/base/dataset.py", line 949, in _share_work
    shape_shm = SharedMemory(name=name, create=True, size=size, auto_cleanup=False)
  File "/streaming/streaming/base/shared/memory.py", line 41, in __init__
    shm = BuiltinSharedMemory(name, create, size)
  File "lib/python3.9/multiprocessing/shared_memory.py", line 103, in __init__
    self._fd = _posixshmem.shm_open(
FileExistsError: [Errno 17] File exists: '/000000_epoch_shape'
```

knighton:
> Now I get an error like:
> 
> ```
>   File "/streaming/streaming/base/dataset.py", line 1458, in __iter__
>     sample_ids = self._get_work(world, epoch, sample_in_epoch)
>   File "/streaming/streaming/base/dataset.py", line 1002, in _get_work
>     shape_shm, data_shm = self._share_work(epoch_sample_ids)
>   File "/streaming/streaming/base/dataset.py", line 949, in _share_work
>     shape_shm = SharedMemory(name=name, create=True, size=size, auto_cleanup=False)
>   File "/streaming/streaming/base/shared/memory.py", line 41, in __init__
>     shm = BuiltinSharedMemory(name, create, size)
>   File "lib/python3.9/multiprocessing/shared_memory.py", line 103, in __init__
>     self._fd = _posixshmem.shm_open(
> FileExistsError: [Errno 17] File exists: '/000000_epoch_shape'
> ```

Investigating...

knighton:
@andreamad8 how about now? You may need to call `streaming.base.util.clean_stale_shared_memory()`

andreamad8:
Ok works now, If you give me 10 min I can test also with multiple nodes.

knighton:
We will begrudge you one sidereal hour

andreamad8:
was a bit longer than an hours :)

yes, it works I tested up to 16 nodes with different TensorParallel size 1, 4, 8.

snarayan21:
Cleaned up the PR to have naming not TP-specific and fixed a bug that was preventing determinism (both elastic and non-elastic). Added tests as well. Pending regression tests, should be good to go!

snarayan21 (approved):
lgtm

PR #619, opened 2024-03-05
Make streaming use the correct number of unique samples with SP/TP

snarayan21:
## Description of changes:
The TP/SP PR assumed that streaming would track/receive an increased number of samples when using SP/TP. For example, with `replication` = 2, Composer would previously track the # of elapsed samples without taking into account replication, which meant that even if 100 unique samples had been seen in training, since `replication` was 2, Composer would think that 200 samples had been seen, and pass that on to Streaming for resumption.

This issue was previously addressed by dividing the sample offset on resumption. With Composer now correctly calculating the number of unique samples elapsed during training, this fix is not needed on the Streaming side.

This change means we also need to track # unique samples using `StreamingDataLoader`.

Tested this with Shashank's help -- loss curves are overlapping on resumption.
Screenshot 2024-03-04 at 6 13 05 PM

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataloader.py
+++ b/streaming/base/dataloader.py
@@ -73,6 +73,13 @@ class StreamingDataLoader(DataLoader):
         if isinstance(self.dataset, StreamingDataset):
             world = World.detect()
             num_samples = self.num_samples_yielded * world.num_ranks
+            if self.dataset.replication is not None:
+                # Check if we are using `replication`. If we are, then we need to adjust the
+                # `num_samples_yielded` to reflect the fact that sample ids are shared across
+                # `replication` consecutive devices. For example, if `replication` is 2, then the
+                # number of samples seen is half the number of samples yielded, since every pair
+                # of devices shares sample ids. So the index into the sample partition is halved.
+                num_samples = num_samples // self.dataset.replication
             return self.dataset.state_dict(num_samples, False)
         return None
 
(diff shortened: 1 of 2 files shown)


streaming/base/dataloader.py
```diff
@@ -73,6 +73,13 @@ def state_dict(self) -> Optional[Dict[str, Any]]:
         if isinstance(self.dataset, StreamingDataset):
             world = World.detect()
             num_samples = self.num_samples_yielded * world.num_ranks
+            if self.dataset.replication is not None:
+                # Check if we are using `replication`. If we are, then we need to adjust the
+                # `num_samples_yielded` to reflect the fact that sample ids are shared across
+                # `replication` consecutive devices. For example, if `replication` is 2, then the
+                # number of samples seen is half the number of samples yielded, since every pair
+                # of devices shares sample ids. So the index into the sample partition is halved.
+                num_samples = num_samples // self.dataset.replication
```
karan6181, line 82: Why put this logic here and not in the dataset.py? Not everyone uses `StreamingDataloader`.

karan6181 (commented):
Is there a way to valid this change using a unit test?

streaming/base/dataloader.py
snarayan21: So previously this logic was in `dataset.py`, but if someone is trying to get deterministic resumption, they should be using `StreamingDataLoader` anyways. The other path for people to get deterministic resumption is through Composer, which already takes care of this issue.

snarayan21: `dataset.py` can't tell if the number of samples passed in is counting unique samples or not, so it's safer not to include it in the dataset.

snarayan21:
@karan6181 it'll only come up during resumption, so Shashank and I tested it on a real run. Image is above

karan6181:
> @karan6181 it'll only come up during resumption, so Shashank and I tested it on a real run. Image is above

Sync'd up offline. The plan is to add the test as part of integration testing.

karan6181 (approved):

PR #661, opened 2024-04-24
Adding `device_per_stream` batching

snarayan21:
## Description of changes:

Adds `device_per_stream` batching method, which ensures that each device batch is drawn from a single stream.

We have gotten asks from internal and external users to for a batching method that will make sure that each *device* batch is drawn from a single stream, which will address cases where users have multiple streams but each stream's tensor sizes are different. With the existing `per_stream` batching method, each global batch is drawn from a single stream, but this also means that the optimizer step is taken with respect to samples just from a single stream. As an example, with 2 streams and `per_stream` batching, if one stream is "easy" (low loss) and one is "hard" (high loss), then the loss can be very bimodal over the course of training:

bimodal_loss

Using `device_per_stream` batching addresses this issue since the global batch can still consist of many streams together, but each device batch will only be from a single stream.

Testing:
* Added unit test.
* Large-scale testing with vision team. Saw expected loss curves and reasonable downloading performance.

## Merge Checklist:

### General

### Tests


tests/test_streaming.py
```diff
@@ -175,6 +177,82 @@ def test_dataloader_per_stream_batching(local_remote_dir: Tuple[str, str], batch
...
+    stream2 = Stream(local=local2, remote=remote2)
+
+    # Build StreamingDataset
+    dataset = StreamingDataset(streams=[stream1, stream2],
+                               shuffle=shuffle,
+                               batch_size=batch_size,
+                               shuffle_seed=seed,
+                               num_canonical_nodes=num_canonical_nodes,
+                               batching_method='per_stream',
+                               shuffle_block_size=100)
```
XiaohanZhangCMU, line 222: Can this test be rewritten such that the new batching  is applied to StreamingDataset directly?

snarayan21: No because `StreamingDataset` will generate the sample partition only at `__iter__` time, and because these are CPU tests, the `World` object will think that we only have 1 device and 1 node. We want to check the per-device-batch behavior, so this requires us to create a `fake_world` object with multiple nodes and devices.

XiaohanZhangCMU (approved):
LGTM.

PR #685, opened 2024-05-24
Fix node calculation in `replication` for `World` object

snarayan21:
## Description of changes:

A customer was hitting an issue where the number of nodes was not being calculated correctly in the `replication` function of the `World` object. This is because we did not properly account and test for cases where the number of TP blocks was not cleanly divisible by the original number of ranks per node. For example, with 37 nodes, 8 GPUs per node, and `replication` of 8, the `World` object returned by the `replication` function would have 5 nodes and 7 ranks per node, which did not account for the last 2 TP blocks (which in this case, were the last 2 nodes).

To address this, we check if the number of TP blocks is divisible by the number of ranks per node, and if it's not, we set the number of nodes to 1. So, using the previous example, we now return a `World` object with 1 node and 37 ranks in the node, corresponding to the 37 TP blocks. Download optimality is still guaranteed with 1 node since canonical nodes will still be partitioned among these 37 ranks of the new `World` object, ensuring that no node will have to download more samples than intended.

I also expanded the test cases to include previously failing cases, since the test runs very fast anyways.

## Merge Checklist:

### General

### Tests

### DIFF
--- a/docs/source/getting_started/main_concepts.md
+++ b/docs/source/getting_started/main_concepts.md
@@ -99,14 +99,12 @@ Below, we pass in a list of {class}`streaming.Stream` objects to a {class}`strea
 stream_1 = Stream(
     remote = 's3://stream_1/directory',
     local = '/local/cache/stream_1',
-    batch_size = 4,
     proportion = 0.25,
 )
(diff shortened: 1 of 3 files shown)


j316chuck (approved):
Thanks for unblocking the customer quickly!

PR #700, opened 2024-06-13
Enable correct resumption from the end of an epoch

snarayan21:
## Description of changes:

This PR enables correct resumption from the end of an epoch. For example, suppose the dataset has 512 samples and I want to resume from sample 512. The intended behavior should be to detect that we are at the end of the epoch and start a new one. There were two small bugs preventing this from happening correctly:

1. The partition algorithms (`orig` and `relaxed`) both threw an error when `num_samples <= drop_first` instead of `num_samples < drop_first`. This means that if we wanted to create a sample partition of 512 samples but drop the first 512 samples, we would throw an error. This is not the right behavior -- we should return an empty sample partition instead.
2. Writing an empty sample partition array to shared memory was not possible since we did not consider that edge case. This has been addressed.

Manual run that had earlier failed, succeeding with these changes: `resumption-issue-testing-RnnDR4`

Unit tests have been added for these cases.

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -922,8 +922,10 @@ class StreamingDataset(Array, IterableDataset):
         sample_ids = np.concatenate(sample_ids).astype(np.int64)
         return shuffle_units, sample_ids
 
-    def _share_work(self,
-                    sample_ids: NDArray[np.int64]) -> Tuple[SharedMemory, Optional[SharedMemory]]:
+    def _share_work(
+        self,
+        sample_ids: NDArray[np.int64],
+    ) -> Tuple[SharedMemory, Optional[SharedMemory]]:
         """Put an epoch's sample ordering into shared memory.
 
         Args:
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -922,14 +922,16 @@ class StreamingDataset(Array, IterableDataset):
         sample_ids = np.concatenate(sample_ids).astype(np.int64)
         return shuffle_units, sample_ids
 
-    def _share_work(self, sample_ids: NDArray[np.int64]) -> Tuple[SharedMemory, SharedMemory]:
+    def _share_work(self,
+                    sample_ids: NDArray[np.int64]) -> Tuple[SharedMemory, Optional[SharedMemory]]:
         """Put an epoch's sample ordering into shared memory.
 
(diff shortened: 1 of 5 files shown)


streaming/base/dataset.py
```diff
@@ -922,14 +922,16 @@ def resample_streams(
         sample_ids = np.concatenate(sample_ids).astype(np.int64)
         return shuffle_units, sample_ids
 
-    def _share_work(self, sample_ids: NDArray[np.int64]) -> Tuple[SharedMemory, SharedMemory]:
+    def _share_work(self,
+                    sample_ids: NDArray[np.int64]) -> Tuple[SharedMemory, Optional[SharedMemory]]:
```
mvpatel2000, line 926:
```suggestion
                    sample_ids: NDArray[np.int64],) -> Tuple[SharedMemory, Optional[SharedMemory]]:
```
cleaner formatting with trailing ,

streaming/base/partition/relaxed.py
```diff
@@ -65,6 +65,7 @@ def get_partitions_relaxed(num_samples: int,
         return get_partitions_orig(num_samples, num_canonical_nodes, num_physical_nodes,
                                    ranks_per_node, workers_per_rank, batch_size, drop_first)
     else:
+        print('WE HERE')
```
mvpatel2000, line 68:
```suggestion
```

streaming/base/partition/relaxed.py
```diff
@@ -82,6 +83,7 @@ def get_partitions_relaxed(num_samples: int,
         partition = get_partitions_orig(num_samples, num_canonical_nodes, initial_physical_nodes,
                                         ranks_per_node, workers_per_rank, initial_batch_size,
                                         drop_first)
+        print('ORIG PARTITION SHAPE: ', partition.shape)
```
mvpatel2000, line 86:
```suggestion
```

mvpatel2000 (approved):

streaming/base/partition/relaxed.py
snarayan21: whoops lol

streaming/base/dataset.py
snarayan21: done

PR #706, opened 2024-06-18
Fix `drop_first` checking in partitioning to account for `world_size` divisibility

snarayan21:
## Description of changes:

This PR loosens restrictions on checking `epoch_size` during the sample partition creation function. If we attempt to create a partition on `epoch_size` number of samples, but `epoch_size` is not divisible by the world size, then we repeat samples to make sure the `epoch_size` is divisible by the world size. However, the current `epoch_size` checking did not take this into account. An explanation of why we have to repeat a few samples is below.

When we partition all samples over nodes/ranks/workers, we need to repeat some samples to make sure that the number of samples in the epoch is divisible by the world size. This need arises because:
1. Users don't want samples to be unnecessarily dropped during training
2. In distributed training, dataloaders across all devices should have the same behavior when a partial global batch is being processed.

For 1., after talking with stakeholders and customers, repeating just a few samples instead of dropping them entirely has been seen as the better tradeoff. So, suppose someone sets `epoch_size` to 500, with a world size of 32. Then 12 samples will be repeated in the epoch to take the epoch size to `512` to make the `epoch_size` divisible by the world size, meaning that we assign the same number of samples to every GPU.

We need 2. because imagine a training step where some GPUs have a full batch of samples, but others have only a partial batch. If `drop_last=True` for the dataloader, then some GPUs will attempt to train since they don't have a partial batch, but other GPUs will drop the last batch, causing an error during training since some GPUs try to train while others don't. Repeating samples to make sure the `epoch_size` is divisible by the world size takes care of this because each GPU will have the same number of samples in every single global batch -- either all GPUs have a full batch, or all GPUs have a partial batch. As a result, we never run into a case where some GPUs attempt to train, while others don't.

Manual test run (`resumption-issue-testing-500ep-WoeCrv`) that successfully resumed, addressing bug report that raised this issue in the first place.

Originally reported [here](https://databricks.slack.com/archives/C02EB3VCYRK/p1716927281329449).

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/partition/__init__.py
+++ b/streaming/base/partition/__init__.py
@@ -56,7 +56,7 @@ def get_partitions(algo: str,
     """
     world_size = ranks_per_node * num_physical_nodes
     num_repeated_samples = world_size - (num_samples % world_size)
-    if (num_samples + num_repeated_samples) < drop_first:
+    if num_samples + num_repeated_samples < drop_first:
         raise ValueError(f'Resuming further into the dataset ({drop_first}) than it has samples ' +
                          f'({num_samples})')
 
--- a/streaming/base/partition/__init__.py
+++ b/streaming/base/partition/__init__.py
@@ -3,6 +3,7 @@
 
 """Apportion shards/samples to nodes/ranks/workers for elastically deterministic sample order."""
 
+import logging
 from typing import Optional
 
 import numpy as np
@@ -10,6 +11,8 @@ from numpy.typing import NDArray
 
 from streaming.base.partition.orig import get_partitions_orig
 from streaming.base.partition.relaxed import get_partitions_relaxed
+
+logger = logging.getLogger(__name__)
 
 algos = {
     'orig': get_partitions_orig,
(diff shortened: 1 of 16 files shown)


dakinggg (approved):
lgtm, wouldn't mind another pair of eyes since I haven't looked at this code in a long time

streaming/base/partition/__init__.py
```diff
@@ -51,6 +54,17 @@ def get_partitions(algo: str,
         NDArray[np.int64]: Partitions of shape (physical nodes, ranks per node, workers per rank,
             batches per worker, batch size).
     """
+    world_size = ranks_per_node * num_physical_nodes
+    num_repeated_samples = world_size - (num_samples % world_size)
+    if (num_samples + num_repeated_samples) < drop_first:
```
mvpatel2000, line 59:
```suggestion
    if num_samples + num_repeated_samples < drop_first:
```

mvpatel2000 (approved):
LGTM

karan6181 (approved):
LGTM. Thank You!

PR #719, opened 2024-07-09
Fix Linting from Pillow version update

XiaohanZhangCMU:
## Description of changes:

Pillow version is bumped to 10.4.0 from 10.3.0, which results in the following Linting error

tests/test_encodings.py:206:31 - error: Argument of type "int" cannot be assigned to parameter "obj" of type "SupportsArrayInterface" in function "fromarray"

Reason is that fromarray expects np.array while np.random.randint returns Union[int, array]. Wrap with np.array fixes it.


snarayan21 (approved):
lgtm!

b-chu (approved):
Thanks!

PR #723, opened 2024-07-12
Improve error message on non-0 rank when index file download failed

bigning:
Improve error message on non-0 rank to let users look at rank 0 traceback when index file download timeout or failed

### DIFF
--- a/streaming/base/stream.py
+++ b/streaming/base/stream.py
@@ -454,8 +454,8 @@ class Stream:
                 wait_for_file_to_exist(
                     filename, TICK, self.download_timeout,
                     f'Index file {os.path.join(self.remote or "", self.split or "", basename)} ' +
-                    f'-> {filename} took too long to download or failed to download. Either increase the ' +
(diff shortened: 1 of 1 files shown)


streaming/base/stream.py
```diff
@@ -454,8 +454,8 @@ def get_shards(self, world: World, allow_unsafe_types: bool) -> List[Reader]:
                 wait_for_file_to_exist(
                     filename, TICK, self.download_timeout,
                     f'Index file {os.path.join(self.remote or "", self.split or "", basename)} ' +
-                    f'-> {filename} took too long to download. Either increase the ' +
-                    f'`download_timeout` value or check the other traceback.')
+                    f'-> {filename} took too long to download or failed to download. Either increase the ' +
+                    f'`download_timeout` value or check the local rank 0 traceback.')
```
karan6181, line 458: `local rank 0` is only true for `index.json` and not for shard file downloads.

karan6181 (approved):

PR #739, opened 2024-07-30
Make Pytest log in color in Github Action

eitanturok:
Currently, running pytest locally generates logs in color. But in Github Actions (GA), the logs are only in white. And it does not look so nice. Adding the flag
```
 --color=yes
```
causes pytest to log in color and makes things look nice again. This improves visibility and makes it easier on the eyes to see when tests fail/succeed.

I tested this flag on my fork of MegaBlocks but it should be the same here.

**Without:**
Here is a [GA](https://github.com/eitanturok/megablocks/actions/runs/10151013963/job/28069397071) without the `--color=yes` flag:

without_color

**With:**
Here is a [GA](https://github.com/eitanturok/megablocks/actions/runs/10165283474/job/28112737604) with the `--color=yes` flag:

with_color

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,7 +101,7 @@ reportUnusedCoroutine = "error"
 # Pytest
 [tool.pytest.ini_options]
 # By default, do not run remote tests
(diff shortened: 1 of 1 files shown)


snarayan21 (approved):
some fuego

karan6181 (approved):

PR #748, opened 2024-08-12
Add default compression, and warning about local paths to dataframe_to_mds

srowen:
### Set more appropriate defaults for compression in dataframe_to_mds.

In a distributed Spark env, it's probably worth saving I/O with compression, as many data types in MDS files will compress well (tokens). This enables default zstd compression if not specified.

### Warn about local-only paths

In Databricks in particular, it's possible to provide a path to remote storage as a FUSE-mounted 'local' path likes /Volumes or /dbfs. It is interpreted as a local path, not remote. This works, but, in practice seems to be suboptimal as MDSWriter will causes lots of writes to the 'local' file. Instead it's probably better and more in line with default logic in CloudUploader to specify a temp dir as the local path, and the FUSE path as remote.

Because this is not _necessarily_ a problem, it generates a warning only, not an error.

## Issue #, if available:

None.

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/converters/dataframe_to_mds.py
+++ b/streaming/base/converters/dataframe_to_mds.py
@@ -265,7 +265,7 @@ def dataframe_to_mds(dataframe: DataFrame,
         raise ValueError(f'`out` and `columns` need to be specified in `mds_kwargs`')
 
     if 'compression' not in mds_kwargs:
-        logger.info("Defaulting to zstd compression")
+        logger.info('Defaulting to zstd compression')
         mds_kwargs['compression'] = 'zstd'
 
     if udf_iterable is not None:
@@ -292,16 +292,15 @@ def dataframe_to_mds(dataframe: DataFrame,
     # Fix output format as mds_path: Tuple(local, remote)
     if cu.remote is None:
         # If dataframe_to_mds is being called, this is in a distributed Spark env.
-        # If there is no remote, it's because the given out path is local, which does not 
-        # in general make sense, unless this 'local' path is FUSE-mounted distributed 
+        # If there is no remote, it's because the given out path is local, which does not
+        # in general make sense, unless this 'local' path is FUSE-mounted distributed
         # storage such as /dbfs or /Volumes in Databricks for example.
         # It's not wrong in this case, but probably nevertheless desirable to specify a local temp
         # path explicitly, to interpret the FUSE-mounted path as remote
-        logger.warning(
-            f"Path {cu.local} is interpreted as a local path. If this is actually " +
-            "mounted distributed storage, it will work, but consider also specifying " +
-            "a local temp path. Pass a (local, remote) tuple as 'out', as in " +
-            f"('/local_disk0/my_tmp', {cu.local})")
+        logger.warning(f"Path {cu.local} is interpreted as a local path. If this is actually " +
+                       'mounted distributed storage, it will work, but consider also specifying ' +
+                       "a local temp path. Pass a (local, remote) tuple as 'out', as in " +
+                       f"('/local_disk0/my_tmp', {cu.local})")
         mds_path = (cu.local, '')
     else:
         mds_path = (cu.local, cu.remote)
--- a/streaming/base/converters/dataframe_to_mds.py
+++ b/streaming/base/converters/dataframe_to_mds.py
@@ -268,10 +268,6 @@ def dataframe_to_mds(dataframe: DataFrame,
         logger.info("Defaulting to zstd compression")
         mds_kwargs['compression'] = 'zstd'
 
-    if 'size_limit' not in mds_kwargs:
-        logger.info("Defaulting size_limit to 1 << 29")
-        mds_kwargs['size_limit'] = 1 << 29
-
     if udf_iterable is not None:
         if 'columns' not in mds_kwargs:
             raise ValueError(
@@ -301,10 +297,11 @@ def dataframe_to_mds(dataframe: DataFrame,
         # storage such as /dbfs or /Volumes in Databricks for example.
         # It's not wrong in this case, but probably nevertheless desirable to specify a local temp
         # path explicitly, to interpret the FUSE-mounted path as remote
-        logger.warning(f"Path {cu.local} is interpreted as a local path. If this is actually " +
-                       "mounted distributed storage, it will work, but consider also specifying " +
-                       "a local temp path. Pass a (local, remote) tuple in out, as in " +
-                       f"('/local_disk0/my_tmp', {cu.local})")
+        logger.warning(
+            f"Path {cu.local} is interpreted as a local path. If this is actually " +
+            "mounted distributed storage, it will work, but consider also specifying " +
+            "a local temp path. Pass a (local, remote) tuple as 'out', as in " +
+            f"('/local_disk0/my_tmp', {cu.local})")
         mds_path = (cu.local, '')
     else:
         mds_path = (cu.local, cu.remote)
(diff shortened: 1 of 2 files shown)


streaming/base/converters/dataframe_to_mds.py
```diff
@@ -287,6 +291,17 @@ def write_mds(iterator: Iterable):
 
     # Fix output format as mds_path: Tuple(local, remote)
     if cu.remote is None:
+        # If dataframe_to_mds is being called, this is in a distributed Spark env.
```
snarayan21, line 294: @XiaohanZhangCMU is this true? I can call dataframe_to_mds from outside a distributed spark env right?

I don't mind the warning, I think it's useful. Just want to check if this comment was right.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -287,6 +291,17 @@ def write_mds(iterator: Iterable):
...
+        # If there is no remote, it's because the given out path is local, which does not 
+        # in general make sense, unless this 'local' path is FUSE-mounted distributed 
+        # storage such as /dbfs or /Volumes in Databricks for example.
+        # It's not wrong in this case, but probably nevertheless desirable to specify a local temp
+        # path explicitly, to interpret the FUSE-mounted path as remote
```
snarayan21, line 299:
```suggestion
        # path explicitly to interpret the FUSE-mounted path as remote
```

streaming/base/converters/dataframe_to_mds.py
snarayan21, line 294:
```suggestion
        # If dataframe_to_mds is being called this is in a distributed Spark env.
```

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -287,6 +291,17 @@ def write_mds(iterator: Iterable):
 
     # Fix output format as mds_path: Tuple(local, remote)
     if cu.remote is None:
+        # If dataframe_to_mds is being called, this is in a distributed Spark env.
+        # If there is no remote, it's because the given out path is local, which does not
```
snarayan21:
```suggestion
        # If there is no remote it's because the given out path is local, which does not 
```

XiaohanZhangCMU (approved):
LGTM. Thanks for contributing.

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -287,6 +291,16 @@ def write_mds(iterator: Iterable):
...
+        # in general make sense, unless this 'local' path is FUSE-mounted distributed
+        # storage such as /dbfs or /Volumes in Databricks for example.
+        # It's not wrong in this case, but probably nevertheless desirable to specify a local temp
+        # path explicitly, to interpret the FUSE-mounted path as remote
+        logger.warning(f"Path {cu.local} is interpreted as a local path. If this is actually " +
+                       'mounted distributed storage, it will work, but consider also specifying ' +
```
snarayan21, line 301:
```suggestion
                       'mounted distributed storage it will work, but consider also specifying ' +
```

snarayan21 (commented):
lgtm but before approval would like @XiaohanZhangCMU to verify comment

snarayan21:
@srowen can you run `pre-commit run --all-files` and push again? seeing linting issues.

XiaohanZhangCMU:
I guess there are some test failures from test_utils/test_merge_index you need to adjust. The ground truth there are different now.

streaming/base/converters/dataframe_to_mds.py
srowen: The method is definitely specific to Spark. But Spark can be run on one machine, yeah. If one passed a local path as 'out' in this case, then that would be fine. Setting ("/tmp/...", "/my/local/path/...") instead would also work; it would just be a bit redundant - it would save to a local path, then copy to another local path.

It would not make sense to path a local path if running on a cluster of machines, as the local path is specific to each individual machine.

Even if running on a single-machine, the storage could be remote. That case is also fine, as CloudUploader will understand the single remote path argument as a remote.

The case here is, really, where you have remote storage that looks like a local path. That arises when you have FUSE mounted storage, and that is definitely how Databricks _can_ work. Then we have a little problem of a lot of little writes to a 'local' path that is actually S3 or similar. It still 'works', just, can be inefficient or hit rate limits.


I think the intended message is: if you pass what looks like a local path here, it's either:
- OK, because this is a single machine though you're using Spark anyway
- Wrong, because you're not referencing a FUSE mount and the files aren't going where you think they are
- OK but inefficient because it's FUSE mounted storage

I can try to think of a rewording here. 
If you passed a 'local' path, warn that this is only correct if running Spark on a single machine, or if the storage is a FUSE mount. In the latter case you want to specify a local temp path, ideally.

But the running Spark on one machine case seemed rare, and, setting a local temp path is OK in that case anyway. So, maybe the simplest guidance is:

Please set a 'local' temp path too, even if your intended 'remote' destination is also a local path

Keep it simple?

streaming/base/converters/dataframe_to_mds.py
srowen: I think the comma correctly adds clarity, and is not wrong in any event

streaming/base/converters/dataframe_to_mds.py
srowen: This comma seems correct, even necessary

streaming/base/converters/dataframe_to_mds.py
```diff
@@ -264,6 +264,10 @@ def write_mds(iterator: Iterable):
     if 'out' not in mds_kwargs:
         raise ValueError(f'`out` and `columns` need to be specified in `mds_kwargs`')
 
+    if 'compression' not in mds_kwargs:
+        logger.info('Defaulting to zstd compression')
+        mds_kwargs['compression'] = 'zstd'
```
karan6181, line 269: This may be very well true for the LLM dataset, but it may not be that useful for the Image-based models since JPEG/PNG is already compressed!!! and you won't see much of the benefits in terms of storage space.

srowen: I agree, that's the type of usage I was thinking of too. I was wondering how much good vs bad this default would do.

For example, in the common case of tokens or text, this probably reduces the size by 3x.
In the case of images, it won't compress and adds a bit of overhead.

It seemed like it would more often help than hurt so I had wanted to establish it as a default, but maybe not worth a behavior change. Not sure.

streaming/base/converters/dataframe_to_mds.py
snarayan21: Sure yeah. sounds good to me. just was in doubt about the wording.

streaming/base/converters/dataframe_to_mds.py
snarayan21: Cool with me, just a minor nit. Not gonna nitpick on grammar too much haha (i am trying to cut my own overuse of commas)

streaming/base/converters/dataframe_to_mds.py
snarayan21: Same as above, very minor and doesn't matter to me

streaming/base/converters/dataframe_to_mds.py
snarayan21: Performance wise this is perfectly fine even with images. I'm good with merging

snarayan21 (approved):
lgtm

srowen:
OK I think it might be ready to go.
Thanks for the attention and patience, I know it was a simple change.
I think this will make a difference to users, getting these right.

snarayan21:
@srowen of course, thanks for getting this one in! Will definitely make a difference for a lot of users.

Apologies for my comma suggestions haha. Resolved those.

Issue #749, opened 2024-08-13
Different lists of examples when shuffle == False


experiencor:
OS: [macOS]
mosaicml-streaming==0.7.6

## To reproduce

from streaming import StreamingDataset, Stream

```
streams = [
    Stream(
        local="/tmp/dataset",
        remote='s3://path/to/dataset',
        choose=10,
    ),
]

dataset = StreamingDataset(
    streams=streams,
    shuffle=False,
    batch_size=1,
    predownload=8
)

for example in dataset:
    print(example["id"])

print("=" * 10)

for example in dataset:
    print(example["id"])
```

## Expected behavior

The same list of examples for 2 iterations of dataset when shuffle = False.

## Actual behavior
Different list of examples for 2 interations of dataset.

![Screenshot 2024-08-13 at 6 23 11 PM](https://github.com/user-attachments/assets/dabfaefe-1ca9-4f0d-b3b6-5eae6f130e30)

XiaohanZhangCMU:
@experiencor with "choose' in the stream, StreamingDataset will do upsample/downsample per stream. The random seed depends on epochs. So when you compare 1st epoch with 2nd epoch, they will be different. 
The randomness comes from [here](https://github.com/mosaicml/streaming/blob/260233048936c744691dea362f115aae0055b26e/streaming/base/sampling.py#L64)

snarayan21:
@experiencor Thanks for filing this issue! Fixed in #750.

PR #750, opened 2024-08-13
Ensure deterministic sample order between epochs when `shuffle=False`

snarayan21:
## Description of changes:

Previously, the sample order would change between epochs even when `shuffle=False`, which is counterintuitive and is not the desired behavior. This PR fixes that.

Addresses #749 

Added unit testing and cleaned up a few others.

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/batching/device_per_stream.py
+++ b/streaming/base/batching/device_per_stream.py
@@ -146,7 +146,7 @@ def generate_work_device_per_stream_batching(dataset: StreamingDataset, world: W
     max_device_batches_per_node += padding_max_device_batches
 
     # Shuffle seed may change with each epoch, which impacts the order of streams in global batches.
-    epoch_seed = dataset.shuffle_seed + epoch if dataset.shuffle else dataset.shuffle_seed
+    epoch_seed = dataset.shuffle_seed + epoch if dataset.epoch_seed_change else dataset.shuffle_seed
     epoch_rng = np.random.default_rng(epoch_seed)
 
     # Shuffle the device batch origin order for each node.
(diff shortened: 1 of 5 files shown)


tests/test_streaming.py
```diff
@@ -85,6 +85,60 @@ def test_new_defaults_warning(local_remote_dir: Tuple[str, str], caplog: Callabl
     assert 'Because `num_canonical_nodes` was not specified,' in caplog.text
 
 
+@pytest.mark.parametrize('batch_size', [4, 7])
```
karan6181, line 88: There are many combinations (2 * 2 * 2 * 4), which increases the unit test time; do we need all of them? Can we reduce the number of combinations?

streaming/base/batching/device_per_stream.py
```diff
@@ -145,8 +145,9 @@ def generate_work_device_per_stream_batching(dataset: StreamingDataset, world: W
     padding_max_device_batches = num_devices - (max_device_batches_per_node % num_devices)
     max_device_batches_per_node += padding_max_device_batches
 
-    # Shuffle seed changes with every epoch, so the order of streams in our batches also changes.
-    epoch_rng = np.random.default_rng(dataset.shuffle_seed + epoch)
+    # Shuffle seed may change with each epoch, which impacts the order of streams in global batches.
```
XiaohanZhangCMU, line 148: nit: comment is a bit unclear, suggest:
#when shuffle is enabled, we use a different rng for different epochs. Otherwise, we keep the seed the same to enable deterministic sequence ordering.

similar to other places.

XiaohanZhangCMU (approved):
LGTM

tests/test_streaming.py
XiaohanZhangCMU: I think it is ok to keep the combinations, they're all relevant. One improvement here may be to deduplicate the dataset creation.

snarayan21: it's a very short test, because there's no dataloader its very quick. i'd like to keep.

but yeah there's future work to have fixtures for dataset creation since a lot of these tests use the same datasets.

streaming/base/batching/device_per_stream.py
snarayan21: addressed!

Issue #771, opened 2024-09-02
MemoryError: Unable to allocate


AugustDev:
## Environment

- OS: [Ubuntu 20.04]
- Hardware (GPU, or instance type): [H100]

## To reproduce

I have very large dataset I converted following "Spark to MDS" tutorial on the MosaicML website. I have the dataset in a disk mounted to my machine. I am able to load eval (much smaller dataset), however when loading train dataset it gives an error.

When loading the dataset I get an error

```
from streaming import StreamingDataset
dataset = StreamingDataset(
    local="/mnt/disks/dataset/train",
    remote=None,
    batch_size=64
)
```

```
Cell In[7], [line 1](vscode-notebook-cell:?execution_count=7&line=1)
----> [1](vscode-notebook-cell:?execution_count=7&line=1) dataset = StreamingDataset(
      [2](vscode-notebook-cell:?execution_count=7&line=2)     local="/mnt/disks/dataset/train"",
      [3](vscode-notebook-cell:?execution_count=7&line=3)     remote=None,
      [4](vscode-notebook-cell:?execution_count=7&line=4)     batch_size=64
      [5](vscode-notebook-cell:?execution_count=7&line=5) )

File /opt/conda/lib/python3.10/site-packages/streaming/base/dataset.py:514, in StreamingDataset.__init__(self, streams, remote, local, split, download_retry, download_timeout, validate_hash, keep_zip, epoch_size, predownload, cache_limit, sampling_method, sampling_granularity, partition_algo, num_canonical_nodes, batch_size, shuffle, shuffle_algo, shuffle_seed, shuffle_block_size, batching_method, allow_unsafe_types, replication)
    [512](https://vscode-remote+ssh-002dremote-002b34-002e123-002e178-002e5.vscode-resource.vscode-cdn.net/opt/conda/lib/python3.10/site-packages/streaming/base/dataset.py:512) self.samples_per_shard = np.array([shard.samples for shard in self.shards], np.int64)
    [513](https://vscode-remote+ssh-002dremote-002b34-002e123-002e178-002e5.vscode-resource.vscode-cdn.net/opt/conda/lib/python3.10/site-packages/streaming/base/dataset.py:513) self.sample_offset_per_shard = self.samples_per_shard.cumsum() - self.samples_per_shard
[...]

PR #773, opened 2024-09-03
Refactor spanner to avoid creating large array

XiaohanZhangCMU:
## Description of changes:

Refactor spanner init to avoid creating a large array which may lead to OOM. 

## Issue #, if available:

[OOM issue](https://github.com/mosaicml/streaming/issues/771)

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/spanner.py
+++ b/streaming/base/spanner.py
@@ -6,7 +6,6 @@
 import numpy as np
 from numpy.typing import NDArray
 
-
 class Spanner:
     """Given a list of shards, construct a mapping of global index to shard and relative index.
 
@@ -25,10 +24,34 @@ class Spanner:
         underflow = span_size - overflow if overflow else 0
         self.shard_sizes[-1] += underflow
 
-        sample_shards = np.repeat(np.arange(len(shard_sizes)), self.shard_sizes)
-        sample_shards = sample_shards.reshape(-1, span_size)
-        span_lowest_shards = sample_shards.min(1)
-        span_highest_shards = sample_shards.max(1)
+        n_shards = len(shard_sizes)
+        current_shard = 0
+        current_position_in_shard = 0
+
+        span_lowest_shards = []
+        span_highest_shards = []
+
+        while current_shard < n_shards:
+            span_min_shard = current_shard
+            span_max_shard = current_shard
+
+            remaining_span_size = span_size
+            while remaining_span_size > 0 and current_shard < n_shards:
+                available_in_current_shard = shard_sizes[current_shard] - current_position_in_shard
+
+                if remaining_span_size >= available_in_current_shard:
+                    remaining_span_size -= available_in_current_shard
+                    current_shard += 1
+                    current_position_in_shard = 0
+                else:
+                    current_position_in_shard += remaining_span_size
+                    remaining_span_size = 0
+
+                if current_shard < n_shards:
+                    span_max_shard = current_shard
+
+            span_lowest_shards.append(span_min_shard)
+            span_highest_shards.append(span_max_shard)
 
         self.spans = []
         for low, high in zip(span_lowest_shards, span_highest_shards):


mihir-db (commented):
Should this have tests

snarayan21 (commented):
ok i think this makes sense, but can you add the following:
1) tests comparing the `self.spans` object with old method vs new method
2) quick local perf benchmark for old vs. new spanner object creation -- want to make sure this doesn't increase init time excessively

AugustDev:
Hi @XiaohanZhangCMU incredible job! I just tested and can confirm that this solves a problem with large number of shards. 

However, there may be another bottleneck with `StreamingDataLoader` on datasets with large number of shards.

The following code takes ~2 minutes and CPU memory utilisation climbs to ~80GB before the first batch can be retrieved. As the dataloader is being used it consumes ~50GB of memory. If I end up using 8 GPUs there may be not enough CPU memory in H100s to be able to use it.

```python
dataset = StreamingDataset(
    remote="/mnt/disks/raw/train",
    local="./local",
    batch_size=1024,
    cache_limit="10GB",
)
dataloader = StreamingDataLoader(dataset, batch_size=64)
for batch in dataloader:
    print(batch)
    break
```

Any ideas what could be going wrong?

AugustDev:
Sharing profiler results on time to first batch

image

XiaohanZhangCMU:
@AugustDev yeah, I wouldn't be surprised with those profiling numbers. There are places that some large arrays of int32 are created, which may add up.....

AugustDev:
I see... So MosaicML currently is not suitable for datasets with large number of shards 😢

snarayan21:
Hey @AugustDev, in your case, you have a whole lot of samples and storing the sample partition array is taking up a good amount of space. However, according to your issue, you have ~27B samples, right? So the sample partition array, even in int64, will be ~216GB, which should be possible. The array is created but then saved to a shared memory file which all workers can access, so this should be a temporary memory cost. Regardless, most H100 systems have nodes with much more CPU ram than 216 GB so this should still be feasible...

Reducing the number of shards can help but will probably not make a significant difference for memory savings. It's more about the sheer number of samples you have.

xiaohanzhan-db:
@AugustDev I would still recommend give it a try using your current set up for 8 gpus (with or without this PR). And let us know if you hit any memory issue. 

To clarify, the number of shards is not the issue, having smaller shards is actually preferred when streaming from cloud. In your case, the partition array (a shared array of int) was causing the large mem allocation. This PR is to reduce the mem cost in initialization but the partition array alloc is not avoidable. But as @snarayan21 mentioned, the partition array is shared across processes so I think you should be good to scale up to 8 gpus.

XiaohanZhangCMU:
@AugustDev I would still recommend give it a try using your current set up for 8 gpus (with or without this PR). And let us know if you hit any memory issue.

To clarify, the number of shards is not the issue, having smaller shards is actually preferred when streaming from cloud. In your case, the partition array (a shared array of int) was causing the large mem allocation. This PR is to reduce the mem cost in initialization but the partition array alloc is not avoidable. But as @snarayan21 mentioned, the partition array is shared across processes so I think you should be good to scale up to 8 gpus.

AugustDev:
Thank you for the message. Will try on H100 - 8GPUs and report back the findings.

snarayan21:
Hey @AugustDev, we've been able to train on datasets that have that many (or more) samples -- this is likely an issue particular to your dataset. Are you trying to retrieve a batch locally on your laptop or from the GPUs themselves? Have you tried model training while dataloading instead of dataloading alone? For performance tuning, you can use the [streaming simulator](https://docs.mosaicml.com/projects/streaming/en/stable/distributed_training/performance_tuning.html#performance-tuning) to input your dataset characteristics and understand what performance you can expect. Happy to help further.

snarayan21:
@XiaohanZhangCMU Mind adding the tests mentioned above and we can get this one in?

PR #777, opened 2024-09-07
Allow JSON encoder to handle ndarray

srowen:
## Description of changes:

The JSON encoder encodes a variety of data types via `json.dumps`, including primitives, lists, dicts, etc.
However `json.dumps` doesn't work with `ndarray`s ; it won't serialize them.
Conceptually, there isn't a strong reason to refuse to encode input as JSON where `ndarray` is used instead of lists, when one could simply transform even multi-dimensional `ndarray`s to supported Python lists with `.tolist()`.

This change simply has the encoder call `.tolist()` on its arg if passed an `ndarray`, and the rest works as expected.

This is actually relevant because the serialization of array type data from Spark to pandas will use `ndarray`s (being based on Arrow), and the result can't be passed to `dataframe_to_mds` even if the output type is given as `'json'`. It's possible to workaround by passing a transformation of the pandas DF to this function. But, it seemed simple and natural to just support this directly.

### General

### Tests
- [X] I have added tests that prove my fix is effective or that my feature works (if appropriate).
- [X] I have added unit and/or integration tests as appropriate to ensure backward compatibility of the changes.

### DIFF
--- a/streaming/base/format/mds/encodings.py
+++ b/streaming/base/format/mds/encodings.py
@@ -510,6 +510,8 @@ class JSON(Encoding):
     """Store arbitrary data as JSON."""
 
     def encode(self, obj: Any) -> bytes:
+        if isinstance(obj, np.ndarray):
+            obj = obj.tolist()
         data = json.dumps(obj)
         self._is_valid(obj, data)
         return data.encode('utf-8')
--- a/tests/test_encodings.py
+++ b/tests/test_encodings.py
@@ -286,6 +286,22 @@ class TestMDSEncodings:
         # Validate data content
         assert dec_data == data
 
+    @pytest.mark.parametrize('data', [np.array([1]), np.array(['foo']), np.array([{'foo': 1}])])
+    def test_json_encode_decode_ndarray(self, data: Any):
+        json_enc = mdsEnc.JSON()
+        assert json_enc.size is None
+
+        # Test encode
+        enc_data = json_enc.encode(data)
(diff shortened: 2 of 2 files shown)


streaming/base/format/mds/encodings.py
```diff
@@ -510,6 +510,8 @@ class JSON(Encoding):
     """Store arbitrary data as JSON."""
 
     def encode(self, obj: Any) -> bytes:
+        if isinstance(obj, np.ndarray):
```
snarayan21, line 513: hm... i agree that this makes sense, but why would we encode ndarrays as json cols instead of directly using the ndarray encoding types that we support?

snarayan21 (commented):
why not just use the ndarray encoding type?

streaming/base/format/mds/encodings.py
srowen: I don't think that generates JSON though. The use case I have in mind (which I should have spelled out) is generating chat-style datasets.

Each input is an array of messages like `{'role': 'user', 'content': '...'}` and we want to generate a JSON array of JSON dicts. So the `'json'` type does seem most natural; this isn't an array of numbers for example and the nested elements do have to be JSON-encoded.

(This relates to my other PR, where we also need support for map types to make those dicts work)

snarayan21: I see. thanks

snarayan21 (approved):
Lgtm

PR #791, opened 2024-09-27
Search local shards directly to find shard to evict

coryMosaicML:
This PR changes the search for coldest shard to avoid looping over remote shards by considering local shards only as possible candidates for eviction.

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -1087,31 +1087,15 @@ class StreamingDataset(Array, IterableDataset):
 
         This method is called internally by ``prepare_shard`` to clear space for more downloads.
         """
-        while True:
-            # Find the shard with the oldest last access time.
-            shard_id = int(self._shard_access_times.numpy().argmin())
-
-            # Check the shard's last access time. If it is NEVER, there are no downloaded shards to
-            # evict. If any shards are currently being downloaded, wait, else raise an error.
-            if self._shard_access_times[shard_id] == NEVER:
-                if (self._shard_states.numpy() == _ShardState.PREPARING).any():
-                    sleep(TICK)
-                    continue
-                else:
-                    raise ValueError(
-                        f'Tried to evict a shard {shard_id}, but no shards are present to evict ' +
-                        f'(cache usage {self.cache_usage} of {self.cache_limit})')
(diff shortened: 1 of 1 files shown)


j316chuck:
https://github.com/databricks-mosaic/regression-testing/actions/runs/11078601583 

Adding regression runs here^

j316chuck:
You can invoke like this
Screenshot 2024-09-27 at 3 40 49 PM

j316chuck:
You can invoke like this
Screenshot 2024-09-27 at 3 40 49 PM

j316chuck:
You can invoke like this 
Screenshot 2024-09-27 at 3 41 37 PM

j316chuck:
Screenshot 2024-09-30 at 2 45 49 PM

Throughput looks [normal](https://databricks.slack.com/archives/C05T1A4UMT8/p1727483855679929 
), can we merge?

streaming/base/dataset.py
```diff
@@ -1087,31 +1087,15 @@ def _evict_coldest_shard(self) -> None:
...
-                self._shard_access_times[shard_id] = NEVER
-                continue
-
-            # Break on success.
-            break
-
+        states = self._shard_states.numpy()
+        access_times = self._shard_access_times.numpy()
+        # Filter indices to include only local shards
+        indices = np.where(states == 3)[0]
```
XiaohanZhangCMU, line 1093: suggest: == _ShardState.LOCAL

XiaohanZhangCMU (changes requested):
What if the evicted shard is currently used by other processes? Is there a mechanism to prevent that?

snarayan21 (commented):
looks nice! agreed with @XiaohanZhangCMU 's comments. I don't think we actually need a mechanism to make sure the local shard isn't used...according to regression tests, this seems fine to me. probabilistically is probably okay.

XiaohanZhangCMU:
@snarayan21 yeah, I'm ok with it. The question I had is not valid, which should have been handled by the get_item retry logic already.

coryMosaicML:
Closing because it is superseded by [795](https://github.com/mosaicml/streaming/pull/795).

PR #813, opened 2024-10-25
Fix shared memory permission issue in a shared pod environment

XiaohanZhangCMU:
## Description of changes:

We run into read permission issues when the shared memory is created in a shared computer environment. We relax the retry condition to allow the creation of a new shared memory file under permission error. When creating a shared memory file, we make sure both LOCALS and FILELOCKS do not overlap with existing ones and they have the same prefix ints in the name. 

What makes it more complicated is when a program exited non-normally, some of the shared meomory files are cleaned up but some are not. To address that, we need to make sure all files in SHM_TO_CLEAN do not have duplicates.

## Merge Checklist:

### General

### Tests

A complete list of testings done for this PR is put in this [doc](https://docs.google.com/document/d/1stz7MMN31nqFODqFf6FDAx1NuGVJ-nhv76FJhGtPIJ8/edit?tab=t.0).

### DIFF
--- a/streaming/base/dataset.py
+++ b/streaming/base/dataset.py
@@ -526,7 +526,7 @@ class StreamingDataset(Array, IterableDataset):
         ]
         self._shm_prefix_int, self._locals_shm = get_shm_prefix(streams_local, streams_remote,
                                                                 self._unique_rank_world)
-        self._filelock_root = os.path.join(gettempdir(), 'streaming')
+        self._filelock_root = gettempdir()
         os.makedirs(self._filelock_root, exist_ok=True)
 
         # Create the shared memory-backed barrier, without its lock, which is unpickleable.
--- a/streaming/base/shared/prefix.py
+++ b/streaming/base/shared/prefix.py
@@ -218,6 +218,11 @@ def get_shm_prefix(streams_local: list[str],
         _check_and_find_retrying(streams_local, streams_remote, shm_name=shm_name, retry=retry)
         for shm_name in SHM_TO_CLEAN
     ])
+    print(f'before first barrier, {world.rank=}')
+
+    if dist.is_available() and dist.is_initialized():
+        dist.barrier()
+    print(f'after first barrier, {world.rank=}')
 
     # First, the local leader registers the first available shm prefix, recording its locals.
     if world.is_local_leader:
--- a/streaming/base/shared/prefix.py
+++ b/streaming/base/shared/prefix.py
@@ -113,9 +113,6 @@ def _check_and_find(streams_local: list[str], streams_remote: list[Union[str, No
 
     for prefix_int in _each_prefix_int():
 
-        print(f'{shm_name=} - {prefix_int=}')
-        temproot = gettempdir()
-        print(f'{temproot=}')
         name = _get_path(prefix_int, shm_name)
 
         # Check if any shared memory filelocks exist for the current prefix
@@ -218,36 +215,23 @@ def get_shm_prefix(streams_local: list[str],
         _check_and_find_retrying(streams_local, streams_remote, shm_name=shm_name, retry=retry)
         for shm_name in SHM_TO_CLEAN
     ])
-    print(f'before first barrier, {world.rank=}')
 
     if dist.is_available() and dist.is_initialized():
-        print(f'before: in dist.avail 1 : {world.rank=}')
         dist.barrier()
-        print(f'after: in dist.avail 1 : {world.rank=}')
-    print(f'after first barrier, {world.rank=}')
 
     # First, the local leader registers the first available shm prefix, recording its locals.
     if world.is_local_leader:
-        print(f'in local leader')
         name = _get_path(prefix_int, LOCALS)
         data = _pack_locals(streams_local, prefix_int)
         shm = SharedMemory(name, True, len(data))
         shm.buf[:len(data)] = data
         their_locals, their_prefix_int = _unpack_locals(bytes(shm.buf))
-        print(f"In world leader: {name=}, {their_locals=}, {their_prefix_int}, {data=}")
 
-    sleep(3)
-    print(f'{world.rank=}')
     if dist.is_available() and dist.is_initialized():
-        print(f'before: in dist.avail 2: {world.rank=}')
         dist.barrier()
-        print(f'after: in dist.avail 2: {world.rank=}')
-
-    #sleep(3)
 
     # Non-local leaders go next, searching for match.
     if not world.is_local_leader:
-        print(f'in non local world leader')
         name = _get_path(prefix_int, LOCALS)
         try:
             shm = SharedMemory(name, False)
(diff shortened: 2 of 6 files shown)


streaming/base/shared/prefix.py
```diff
@@ -112,6 +112,10 @@ def _check_and_find(streams_local: list[str], streams_remote: list[Union[str, No
             shm = SharedMemory(name, False)
         except FileNotFoundError:
             break
+        except PermissionError:
+            # When trying to attach to a shm file on shared cluster
```
snarayan21, line 116:
```suggestion
            # Trying to attach to a shm file on shared cluster
```

streaming/base/shared/prefix.py
```diff
@@ -112,6 +112,10 @@ def _check_and_find(streams_local: list[str], streams_remote: list[Union[str, No
...
+            # may run into permission issue, we need to keep search.
```
snarayan21, line 117:
```suggestion
            # may run into permission issues. Continue searching.
```

snarayan21 (commented):
this seems fine to me, but are you able to reproduce this with an example where this used to fail, and then installing streaming off this branch now succeeds?

snarayan21:
closing after offline discussion with @XiaohanZhangCMU

streaming/base/dataset.py
```diff
@@ -526,7 +526,7 @@ def __init__(self,
         ]
         self._shm_prefix_int, self._locals_shm = get_shm_prefix(streams_local, streams_remote,
                                                                 self._unique_rank_world)
-        self._filelock_root = os.path.join(gettempdir(), 'streaming')
```
snarayan21, line 529: why this change?

streaming/base/shared/prefix.py
```diff
@@ -108,8 +110,22 @@ def _check_and_find(streams_local: list[str], streams_remote: list[Union[str, No
...
         name = _get_path(prefix_int, LOCALS)
+
+        # Check if any shared memory filelocks exist for the current prefix
+        try:
+            shared_memory_exists = any(
+                os.path.exists(os.path.join(gettempdir(), _get_path(prefix_int, shm_name)))
+                for shm_name in SHM_TO_CLEAN)
+            if shared_memory_exists:
+                continue
+        except PermissionError:
```
snarayan21, line 121: at which point would the permission error be thrown? when we check if the path exists?

snarayan21 (commented):
seems fine to me...i think. would like to see testing

tests/test_shared.py
```diff
@@ -157,3 +161,18 @@ def test_shared_array_size_is_integer(mock_shared_memory: MagicMock, dtype: type
     mock_shared_memory.assert_called_once()  # pyright: ignore
     size_arg = mock_shared_memory.call_args[1]['size']
     assert isinstance(size_arg, int), 'Size passed to SharedMemory is not an integer'
+
+
+def test_check_and_find_skips_filelock_conflict():
```
karan6181, line 166: Can we also add a mock test for PermissionError?

streaming/base/shared/prefix.py
```diff
@@ -127,15 +155,16 @@ def _check_and_find(streams_local: list[str], streams_remote: list[Union[str, No
                         raise ValueError(
                             f'Reused local directory: {streams_local} vs ' +
                             f'{their_locals}. Provide a different one. If using ' +
-                            f'a unique local directory, try deleting the local directory and ' +
-                            f'call `streaming.base.util.clean_stale_shared_memory()` only once ' +
-                            f'in your script to clean up the stale shared memory before ' +
+                            f'a unique local directory, try deleting the local directory and '
```
karan6181, line 158: Please revert this change, the `+` went to the next line which is not required

XiaohanZhangCMU:
@karan6181 @snarayan21 @bigning testings are done. PR is ready for review then merge. PTAL.

streaming/base/shared/prefix.py
snarayan21: @XiaohanZhangCMU is this resolved / an issue?

streaming/base/shared/prefix.py
```diff
@@ -184,29 +210,37 @@ def get_shm_prefix(streams_local: list[str],
     # Check my locals for overlap.
     _check_self(streams_local)
 
+    prefix_int = max([
```
snarayan21, line 213: is there a case where the found prefix ints would be different for the different shm names? is this the duplicate shmem stuff / not getting cleaned thing you mention in the description?

snarayan21 (commented):
couple questions, but seems fine to me overall. Deferring to @karan6181 and @bigning on approval

streaming/base/shared/prefix.py
karan6181: > is there a case where the found prefix ints would be different for the different shm names?

I strongly doubt that this will ever happen.

streaming/base/shared/prefix.py
```diff
@@ -184,29 +210,37 @@ def get_shm_prefix(streams_local: list[str],
...
+        dist.barrier()
+
     # First, the local leader registers the first available shm prefix, recording its locals.
     if world.is_local_leader:
-        prefix_int = _check_and_find_retrying(streams_local, streams_remote, retry)
         name = _get_path(prefix_int, LOCALS)
         data = _pack_locals(streams_local, prefix_int)
         shm = SharedMemory(name, True, len(data))
         shm.buf[:len(data)] = data
+        their_locals, their_prefix_int = _unpack_locals(bytes(shm.buf))
```
karan6181, line 227: We do not need this change right?

streaming/base/shared/prefix.py
```diff
@@ -184,29 +210,37 @@ def get_shm_prefix(streams_local: list[str],
...
+        name = _get_path(prefix_int, LOCALS)
+        try:
+            shm = SharedMemory(name, False)
+        except FileNotFoundError:
+            raise RuntimeError(f'Internal error: shared memory prefix was not registered by ' +
+                               f'local leader. This may be because you specified ' +
+                               f'different ``local`` parameters from different ranks.')
+
+        their_locals, their_prefix_int = _unpack_locals(bytes(shm.buf))
+        if streams_local != their_locals or prefix_int != their_prefix_int:
```
karan6181, line 243: Should we add a debug log to print all these variables before we raise an exception?

streaming/base/shared/prefix.py
```diff
@@ -184,29 +210,37 @@ def get_shm_prefix(streams_local: list[str],
     # Check my locals for overlap.
     _check_self(streams_local)
 
+    prefix_int = max([
+        _check_and_find_retrying(streams_local, streams_remote, shm_name=shm_name, retry=retry)
```
bigning, line 214: Just for my learning, why do we need retry here? is there any exception that not captured inside `_check_and_find`?

streaming/base/shared/prefix.py
bigning: or maybe print these variables in the RuntimeError message.

bigning (commented):
thank you for the fix! Will let @karan6181  stamp!

streaming/base/shared/prefix.py
XiaohanZhangCMU: Added to err msg.

streaming/base/shared/prefix.py
XiaohanZhangCMU: @karan6181 may know better why it was there. 
My understanding is that there can be some racing when multiple processes try to attach the same shm file at the same time.

karan6181: There may be some old processes attempting to clean up while the new process tries to attach to shared memory. The retries allow the old processes to finish their tasks.

karan6181 (approved):
LGTM. Thank You!

PR #816, opened 2024-10-28
Fix device_per_stream bugs

quentindrx:
## Description of changes:

Fixed a few bugs when the batching method is set to `device_per_stream` and the number of samples in a stream is small:

1. If a stream has `num_full_batches = 0`, we raise a warning and not an error. This happens typically if the numebr of samples in the stream is smaller than the device batch size.
2. In case we shuffle, we can have `samples_in_stream < num_canonical_nodes`. In this case we drop the samples in the stream instead of returning an error.
3. We can have `shuffle_block_portion = 0` if a stream proportion is too small. (Not covered in the tests)

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/batching/device_per_stream.py
+++ b/streaming/base/batching/device_per_stream.py
@@ -76,6 +76,13 @@ def generate_work_device_per_stream_batching(dataset: StreamingDataset, world: W
             # same as the ratio of the stream's samples to overall samples.
             # This ensures that the overall training shuffle block size is still approximately
             # equal to what is set by the user, and allows for reasoning about cache_limit as well.
+            if samples_in_stream < dataset.num_canonical_nodes:  # third patch
+                logger.warning(
+                    f"Because of the `device_per_stream` batching method and stream with index {stream_id} "
+                    + f"has less samples ({samples_in_stream}) than canonical nodes ({dataset.num_canonical_nodes}), stream will be dropped."
+                )
+                continue
+            
             if not isinstance(dataset.shuffle_block_size, int):
                 raise TypeError(f'Dataset `shuffle_block_size` must be an integer. ' +
                                 f'Got {type(dataset.shuffle_block_size)} instead.')
@@ -114,7 +121,7 @@ def generate_work_device_per_stream_batching(dataset: StreamingDataset, world: W
                 (stream_samples_inorder, np.full(padding_samples, -1)))
             # Reshape samples to be device batches in order of traversal.
             stream_samples_inorder = stream_samples_inorder.reshape(-1, batch_size)
-            num_full_batches = np.count_nonzero(np.min(stream_samples_inorder, axis=1) >= 0)
+            num_full_batches = max(1, np.count_nonzero(np.min(stream_samples_inorder, axis=1) >= 0))
             per_node_batches_per_stream.append(num_full_batches)
             if num_full_batches != stream_samples_inorder.shape[0]:
                 logger.warning(
@@ -123,7 +130,7 @@ def generate_work_device_per_stream_batching(dataset: StreamingDataset, world: W
             if num_full_batches > 0:
                 per_node_stream_partitions.append(stream_samples_inorder[:num_full_batches])
             else:
-                raise ValueError(
+                logger.warning(
                     f'Stream with index {stream_idx} does not have an adequate number of ' +
                     f'samples to construct even a single device batch of size {batch_size}. ' +
                     f'Training will occur without any samples from this stream!')
--- a/streaming/base/batching/device_per_stream.py
+++ b/streaming/base/batching/device_per_stream.py
@@ -76,7 +76,7 @@ def generate_work_device_per_stream_batching(dataset: StreamingDataset, world: W
             # same as the ratio of the stream's samples to overall samples.
             # This ensures that the overall training shuffle block size is still approximately
             # equal to what is set by the user, and allows for reasoning about cache_limit as well.
-            if samples_in_stream < dataset.num_canonical_nodes:  # third patch
+            if samples_in_stream < dataset.num_canonical_nodes:
                 logger.warning(
                     f'Because of the `device_per_stream` batching method and stream with index {stream_id} '
                     +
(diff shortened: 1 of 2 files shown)


snarayan21:
Hey @quentindrx, thanks for the contribution! It would be very helpful if you could add to the PR description the reasons for your changes and specifically what you changed. Then we can kick off CI and review.

tests/test_streaming.py
```diff
@@ -303,12 +303,12 @@ def test_dataloader_device_per_stream_batching(local_remote_dir: tuple[str, str]
...
     for device_batch in sample_partition:
-        if device_batch[0] < 200:
+        if device_batch[0] < 50:
             # Ensure all samples are from stream 1
-            assert (device_batch < 200).all()
+            assert (device_batch < 50).all()
         else:
             # Ensure all samples are from stream 2
-            assert (device_batch >= 200).all()
+            assert ((device_batch >= 50) + (device_batch == -1)).all()
```
quentindrx, line 311: Because of the padding value = -1

ethantang-db:
@snarayan21, want to confirm a few things

for case 1. Should we just combine the stream with another stream if available? I made a change down the line in composer to basically throw an error (unless user decides to ignore that warning) if we don't end up training a batch even on a device.

streaming/base/batching/device_per_stream.py
```diff
@@ -76,10 +76,18 @@ def generate_work_device_per_stream_batching(dataset: StreamingDataset, world: W
...
+                    +
+                    f'has less samples ({samples_in_stream}) than canonical nodes ({dataset.num_canonical_nodes}), stream will be dropped.'
+                )
+                continue
+
             if not isinstance(dataset.shuffle_block_size, int):
                 raise TypeError(f'Dataset `shuffle_block_size` must be an integer. ' +
                                 f'Got {type(dataset.shuffle_block_size)} instead.')
-            shuffle_block_portion = int(dataset.shuffle_block_size * stream.proportion)
+            shuffle_block_portion = max(1, int(dataset.shuffle_block_size * stream.proportion))
```
ethantang-db, line 90: Not sure what the implication of this change would be, or why it is necessary?

quentindrx: If you have a stream with few samples and a lot of samples across all streams, you end up with stream.proportion being very small and then `int(dataset.shuffle_block_size * stream.proportion) = 0`.
Then `get_shufffle` called next line will raise an error [this line](https://github.com/mosaicml/streaming/blob/ecca82006d6d1c97f7ee1d7c46bc5df61577efc9/streaming/base/shuffle/py1e.py#L101C61-L101C71) as it tries to sample an integer between in [0,0)

snarayan21: I see, nice catch! In this case I would actually prefer to throw an error if a stream is not being used, so if `shuffle_block_portion == 0`, we should throw an error. A stream of data that's provided but not used in training is a significant issue and imo training should not proceed.

streaming/base/batching/device_per_stream.py
```diff
@@ -76,10 +76,18 @@ def generate_work_device_per_stream_batching(dataset: StreamingDataset, world: W
             # same as the ratio of the stream's samples to overall samples.
             # This ensures that the overall training shuffle block size is still approximately
             # equal to what is set by the user, and allows for reasoning about cache_limit as well.
+            if samples_in_stream < dataset.num_canonical_nodes:  # third patch
```
snarayan21, line 79: nit: stray comment
```suggestion
            if samples_in_stream < dataset.num_canonical_nodes:
```

snarayan21 (changes requested):
Hey @quentindrx thanks for this PR. In my view, if a stream provided by the user does not have enough samples to make it into training, this should throw an error and training should not proceed. However, in your use cases that does not seem to be the case. I'd like to understand a bit more the cases in which you saw this, and why you would still want training to proceed instead of removing the small stream or reconfiguring your dataset proportions.

snarayan21:
@ethantang-db re: combining streams, that can break dataset mixing / result in unintended behaviors. For example, with multiple streams, it's not clear which ones we should combine, how to sample the new stream in all cases, etc.

quentindrx:
@snarayan21 I don't think we should throw an error when there is not enough samples in a stream to make a single batch.
1. Consistency across the codebase. For the per_device batching method [we only raise a warning ](https://github.com/mosaicml/streaming/blob/fd9b55a875e499da0a49159d4958a1bf413092a9/streaming/base/batching/per_stream.py#L101)
2. When I prepare my data I don't how many nodes I will use nor my batch size. Practically, when I launch a debug run on a few nodes, I have a quite smaller batch size and more chances to have enough samples in a given stream. Then when I scale to more nodes and a bigger batch size, I have my run crashing.
3. We always loose samples when there is not enough samples left in a stream to make a complete batch. Here it's just the extreme case.

Thanks for the great work on this library ❤️

snarayan21:
Hey @quentindrx!

I see, thanks for elaborating. And thanks for catching that inconsistency in the `per_stream` batching method.

While this may be a pain point in your workflows, we have other customers, users, and internal use cases where people expect all the streams they specified to be used in experiments and runs. I agree that we should have consistency, but in light of this I would rather be consistent in erroring if a stream is unused.

As for batch size changing when streams are used, if the device batch size stays constant when you scale your run to a larger number of GPUs, I would think that the stream sampling would remain the same (and all streams should get used). Are you using the same proportions? Additionally, StreamingDataset provides elastic determinism so whatever batch size you use at smaller scale should deterministically have the same sample order for a larger training job.

quentindrx:
Sounds fair. Feel free to close the PR then. I'll keep patching.

> As for batch size changing when streams are used, if the device batch size stays constant when you scale your run to a larger number of GPUs, I would think that the stream sampling would remain the same (and all streams should get used). 

Right ! 

> Are you using the same proportions? Additionally, StreamingDataset provides elastic determinism so whatever batch size you use at smaller scale should deterministically have the same sample order for a larger training job.

The proportion changes indeed as I am often using only a subset of my data when I debug.

snarayan21:
closing!

PR #838, opened 2024-12-03
Fix import paths in `simulation` module

srstevenson:
Prior to [93af28](https://github.com/mosaicml/streaming/commit/93af28ab877aba8a20b55eae576938b289763e49), the `simulation` directory was added to `PYTHONPATH` by appending to `sys.path` at the start of each file, before submodules of `simulation` such as `core` and `interfaces` were imported.

In [93af28](https://github.com/mosaicml/streaming/commit/93af28ab877aba8a20b55eae576938b289763e49), the calls to `sys.path.append` were moved below the import block and so relative imports of `core` and `interfaces` no longer resolve and the simulator fails to start with errors such as:

```
  File "/home/scott/projects/streaming/.venv/lib64/python3.12/site-packages/streamlit/runtime/scriptrunner/exec_code.py", line 88, in exec_func_with_error_handling
    result = func()
             ^^^^^^
  File "/home/scott/projects/streaming/.venv/lib64/python3.12/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 579, in code_to_exec
    exec(code, module.__dict__)
  File "/home/scott/projects/streaming/simulation/interfaces/sim_ui.py", line 17, in <module>
    from core.create_index import create_stream_index
ModuleNotFoundError: No module named 'core'
```

This commit prefixes the submodule imports with `simulation.` so the imports can be resolved, and without needing to manipulate `sys.path`. This works even when running the simulator outside the source repository, as `setuptools.find_packages` is used in `setup.py` to find the modules to package, and this finds the `simulation` module.

### DIFF
--- a/simulation/core/main.py
+++ b/simulation/core/main.py
@@ -8,12 +8,13 @@ import time
 from typing import Generator, Union
 
 import numpy as np
-from core.node_tracker import NodeTracker
-from core.shard_downloads import run_cache_limit, simulate_shard_downloads
-from core.sim_dataset import SimulationDataset
-from core.sim_time import Time
-from core.utils import bytes_to_time, get_batches_epochs, time_to_bytes
 from numpy.typing import NDArray
+
+from simulation.core.node_tracker import NodeTracker
+from simulation.core.shard_downloads import run_cache_limit, simulate_shard_downloads
+from simulation.core.sim_dataset import SimulationDataset
+from simulation.core.sim_time import Time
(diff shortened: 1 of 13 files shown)


ethantang-db:
This looks reasonable to me right now, kicked off the tests to see if it breaks anthing else (linting having issues is a known issue). @snarayan21 to approve though

snarayan21 (approved):
thanks for the PR, lgtm as well

PR #839, opened 2024-12-03
Change simulator UI help text to refer to directory

srstevenson:
When using the mode of the simulator in which you provide an existing `index.json` file, the help text said to provide the absolute path to the `index.json` file. However if you do so, the string `index.json` is then appended to this leading to errors such as:

```
RuntimeError: No `remote` provided, but local file /home/scott/scratch/dummy-mds-dataset/index.json/index.json does not exist either
```

This PR changes the help text to instruct the user to provide the path to the directory containing the `index.json` file, instead of to the file itself.

**Before**:

**After**:


snarayan21:
@srstevenson ah thanks, nice catch! lgtm

snarayan21 (approved):
lgtm

PR #858, opened 2025-01-02
Use registry when creating Stream in StreamingDataset

es94129:
## Description of changes:
Add registry support to allow custom implementations of components like `Stream`, (in the future) batching methods and shuffling algorithms, similar to [LLM foundry's registry](https://github.com/mosaicml/llm-foundry/tree/3269c7399add8ca30842edbeb83d0c82f7906726?tab=readme-ov-file#registry).

Changes made in this PR:
- Copy the [registry_utils.py](https://github.com/mosaicml/llm-foundry/blob/3269c7399add8ca30842edbeb83d0c82f7906726/llmfoundry/utils/registry_utils.py) file and corresponding tests from LLM foundry
- Set up a `streams_registry` that contains the `Stream` class
- Enable `StreamingDataset` to use the stream instance constructed from `streams_registry`

Existing `Stream` and `StreamingDataset` functionality remains unchanged, with added support for custom Stream implementations. See `test_registry.py` and [LLM foundry's README of registry](https://github.com/mosaicml/llm-foundry/tree/3269c7399add8ca30842edbeb83d0c82f7906726?tab=readme-ov-file#registry) for more usage examples. Minimal example:

```python
from streaming.base.registry_utils import construct_from_registry
from streaming.base.stream import streams_registry

streams_registry.register('stream', func=MyStream)

kwargs = {
    'remote': remote,
    'local': local,
    **kwargs,
}

stream = construct_from_registry(
    'stream',
    streams_registry,
    partial_function=False,
    kwargs=kwargs,
)

assert local == stream.local
```

## Issue #, if available:
N/A

## Merge Checklist:

### General
- [x] I have updated any necessary documentation, including [README](https://github.com/mosaicml/streaming/blob/main/README.md) and [API docs](https://github.com/mosaicml/streaming/tree/main/docs) (if appropriate). -- added under the section "Using Streams"

### Tests
- [x] I have added unit and/or integration tests as appropriate to ensure backward compatibility of the changes. -- covered by existing tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,8 +79,8 @@ cd docs && make clean && make doctest # run doctests
 
 <!--pytest.mark.skip-->
 ```bash
-cd docs
 pip install -e '.[docs]'
+cd docs
 make clean && make html
 make host   # open the output link in a browser.
 ```
--- a/docs/source/dataset_configuration/mixing_data_sources.md
+++ b/docs/source/dataset_configuration/mixing_data_sources.md
@@ -29,6 +29,7 @@ dataset = StreamingDataset(
     remote='s3://some/path',
     local='/local/path',
     stream_name='my_stream',
+    stream_config={'arg1': 'val1'},
 )
 ```
 
--- a/docs/source/dataset_configuration/mixing_data_sources.md
+++ b/docs/source/dataset_configuration/mixing_data_sources.md
@@ -8,6 +8,30 @@ A stream is a data source, as a collection of shard files (or set of subdirector
 
 It is possible, though not recommended, for streams to have different schemas.
 
+### Registering a custom Stream implementation
+You can also customize the implementation of a `Stream`. To modify the behavior of a `Stream` that is used in a `StreamingDataset`, you can subclass `Stream`, and register the subclass as shown in the below example without forking the library.
+
+```python
+from streaming.base.stream import streams_registry
+from streaming.base.registry_utils import construct_from_registry
+
+class MyStream(Stream):
+    # your implementation goes here
+
+# Register your custom stream class as 'stream'
+streams_registry.register('stream', func=MyStream)
+
+# StreamingDataset creates a stream instance from the streams_registry
+stream = construct_from_registry(
+    'stream',
+    streams_registry,
+    partial_function=False,
+    kwargs={'remote': remote, 'local': local}
+)
+```
+
+See more methods for registering custom Stream classes in [this README section of LLM Foundry](https://github.com/mosaicml/llm-foundry/tree/3269c7399add8ca30842edbeb83d0c82f7906726?tab=readme-ov-file#how-to-register).
+
 ## Configuring the data mix
 The `proportion`, `repeat`, or `choose` arguments to `Stream` are used to configure different dataset mixing schemes. Only one of them may be set at a time, and all streams must use the same mixing scheme (e.g., Stream A with `proportion` and Stream B with `choose` are incompatible).
 - **`proportion`**: Specifies how to sample this Stream relative to other Streams.
--- a/docs/source/dataset_configuration/mixing_data_sources.md
+++ b/docs/source/dataset_configuration/mixing_data_sources.md
@@ -11,12 +11,14 @@ It is possible, though not recommended, for streams to have different schemas.
 ### Registering a custom Stream implementation
 You can also customize the implementation of a `Stream`. To modify the behavior of a `Stream` that is used in a `StreamingDataset`, you can subclass `Stream`, and register the subclass as shown in the below example without forking the library.
 
+<!--pytest.mark.skip-->
 ```python
 from streaming.base.stream import streams_registry
 from streaming.base.registry_utils import construct_from_registry
 
 class MyStream(Stream):
     # your implementation goes here
+    pass
 
 # Register your custom stream class as 'stream'
 streams_registry.register('stream', func=MyStream)
--- a/docs/source/dataset_configuration/mixing_data_sources.md
+++ b/docs/source/dataset_configuration/mixing_data_sources.md
@@ -1,3 +1,5 @@
+from matplotlib.artist import kwdoc
+
 # Mixing Datasets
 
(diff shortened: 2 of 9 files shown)


es94129:
Hoping to request a review from @XiaohanZhangCMU too, but Github wouldn't let me do that

tests/test_stream.py
```diff
@@ -69,3 +71,26 @@ def test_missing_index_json_local(local_remote_dir: Any):
     stream = Stream(remote=None, local=remote_dir)
     with pytest.raises(RuntimeError, match='No `remote` provided, but local file.*'):
         _ = StreamingDataset(streams=[stream], batch_size=1)
+
+
+def test_construct_stream_from_registry():
+    remote = 'remote_dir'
+    local = tempfile.mkdtemp()
+
+    kwargs = {
```
snarayan21, line 80: mind adding test cases for just local only and just remote only as well?

snarayan21 (changes requested):
code and tests all lgtm, thanks for adding this! Could you please add a few sentences about registries in the streaming docs, likely in this section: https://docs.mosaicml.com/projects/streaming/en/stable/dataset_configuration/mixing_data_sources.html#using-streams

Can be very similar to what you have in the PR description. And a link to LLM Foundry's READMEs on registry usage would be useful too.

Thanks!

CONTRIBUTING.md
```diff
@@ -79,8 +79,8 @@ cd docs && make clean && make doctest # run doctests
 
 <!--pytest.mark.skip-->
 ```bash
-cd docs
 pip install -e '.[docs]'
```
es94129, line 82: Moved `cd docs` below because this pip install line needs to be run under the project root

tests/test_stream.py
es94129: Thanks for the review! Of course, added the documentation and tests now.

CONTRIBUTING.md
snarayan21: ah true. thanks!

docs/source/dataset_configuration/mixing_data_sources.md
```diff
@@ -8,6 +8,32 @@ A stream is a data source, as a collection of shard files (or set of subdirector
...
+<!--pytest.mark.skip-->
+```python
+from streaming.base.stream import streams_registry
+from streaming.base.registry_utils import construct_from_registry
+
+class MyStream(Stream):
+    # your implementation goes here
+    pass
+
+# Register your custom stream class as 'stream'
```
snarayan21, line 23: Can we have the docs example register and construct `my_stream` instead? Because the regular Stream class is already registered as `stream`.

tests/test_stream.py
```diff
@@ -69,3 +71,75 @@ def test_missing_index_json_local(local_remote_dir: Any):
...
+    )
+
+    assert isinstance(stream_instance, Stream)
+    assert remote == stream_instance.remote
+    assert local == stream_instance.local
+
+    shutil.rmtree(local, ignore_errors=True)
+
+
+def test_construct_stream_from_registry_remote_is_none():
```
snarayan21, line 126: Can you combine these three tests by parametrizing the `kwargs` variable with the three dicts, using `pytest.mark.parametrize` syntax? Would make this cleaner.

snarayan21 (commented):
thanks for adding tests and updating the docs! Just minor changes.

streaming/base/dataset.py
```diff
@@ -334,7 +335,8 @@ def __init__(self,
                  shuffle_block_size: Optional[int] = None,
                  batching_method: str = 'random',
                  allow_unsafe_types: bool = False,
-                 replication: Optional[int] = None) -> None:
+                 replication: Optional[int] = None,
+                 **kwargs: Any) -> None:
```
dakinggg, line 339: and docstring for kwargs to note that the kwargs will get passed along to the stream constructor?

streaming/base/dataset.py
```diff
@@ -438,13 +440,27 @@ def __init__(self,
...
+                'download_retry': download_retry,
+                'download_timeout': download_timeout,
+                'validate_hash': validate_hash,
+                'keep_zip': keep_zip,
+                **kwargs,
+            }
+
+            # Construct a Stream instance using registry-based construction
+            default = construct_from_registry(
+                name='stream',
```
dakinggg, line 456: maybe would be more clear to take a `stream_name` arg in the dataset constructor (with a default of `default_stream` or wtvr) and use that here? not a super strong opinion

streaming/base/dataset.py
snarayan21: hm..also instead of having kwargs, if this is only for the stream constructor, should this just be `stream_kwargs: dict[str, Any]` instead of generic **kwargs? Can we pass in stream kwargs like that instead?

docs/source/dataset_configuration/mixing_data_sources.md
es94129: Updated the example code snippet to register as `my_stream` and pass the registered name to `StreamingDataset(... stream_name='my_stream')`.

tests/test_stream.py
```diff
@@ -69,3 +71,75 @@ def test_missing_index_json_local(local_remote_dir: Any):
     stream = Stream(remote=None, local=remote_dir)
     with pytest.raises(RuntimeError, match='No `remote` provided, but local file.*'):
         _ = StreamingDataset(streams=[stream], batch_size=1)
+
+
```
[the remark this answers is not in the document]
es94129: This PR only changes StreamingDataset for the case of a single stream via local/remote. The case for multi-streams wasn't changed, so additional test coverage isn't needed for that code path.

streaming/base/dataset.py
es94129: Added `stream_name: str` and `stream_config: Optional[Dict[str, Any]]` and a corresponding unit test `test_custom_stream_name_and_kwargs` (in `test_streaming.py`).

streaming/base/dataset.py
es94129: Yeah I think it's a good idea, makes it easier to specify which Stream class to use too

Issue #958, opened 2025-10-31
Performance imporve: mmap and file handler cache


yezhengmao1:
## 🚀 Feature Request

**Optimize Streaming `mds reader` I/O Performance with `mmap` and File Handle Caching**

This proposal suggests implementing a file handle cache (using an LRU policy) and memory-mapped (`mmap`) I/O to resolve a significant performance bottleneck in the streaming `mds reader`.

## Motivation

I'm frustrated by the poor performance of our streaming data pipeline. Profiling has revealed a critical I/O bottleneck within the `mds reader` component. The current implementation is inefficient, and a large portion of execution time is spent on I/O overhead rather than useful work.

Our analysis indicates the time is spent as follows:
*   **Frequent File Handle Operations:** Approximately **$1/3$** of the execution time is consumed by repeatedly opening and closing file handles (`open()`/`close()` system calls). This is a major source of overhead.
*   **Inefficient I/O Reads:** Another **$1/3$** of the time is spent on the actual data transfer during I/O read operations.

This inefficient I/O pattern leads to high system call overhead, slow data loading, and underutilization of compute resources, significantly hindering the throughput of our training and inference workloads.

## [Optional] Implementation

To address this, I propose a two-part solution:

1.  **File Handle LRU Cache**
    *   Implement a Least Recently Used (LRU) cache to manage open file handles.
    *   When the reader needs to access a file, it will first check the cache for a valid, open file handle. If one exists, it will be reused, completely avoiding the expensive `open()` system call.
    *   If the cache is full, the least recently used file handle will be closed to free up system resources.
    *   The size of this cache should be configurable to allow users to tune it based on their system's file handle limits and specific data access patterns.

2.  **Memory-Mapped I/O (`mmap`)**
    *   Instead of using traditional `read()` system calls, we should use `mmap` to map file contents directly into the process's virtual address space.
    *   **Benefits:**
        *   **Reduces System Calls:** After the initial `mmap` call, subsequent data access does not require further system calls.
        *   **Eliminates Data Copies:** `mmap` avoids the extra memory copy from the kernel's page cache to the user-space buffer that occurs with `read()`.
        *   **Leverages OS Optimizations:** It allows the operating system's virtual memory manager to handle paging and pre-fetching in a much more efficient manner.

These two strategies work synergistically: the LRU cache provides a persistent file descriptor, and `mmap` uses that descriptor to establish a highly efficient "zero-copy" read pathway.

## Additional context

[...]

PR #965, opened 2026-01-21
Fix cache lock file name bug

### LINKED ISSUE #963: Bug: CACHE_FILELOCK attribute check always fails due to attribute name mismatch
## Description
The lazy initialization check for _cache_filelock in StreamingDataset always fails because of a mismatch between the attribute name being checked and the attribute name being set.

## Root Cause
In `streaming/base/constant.py`:
`CACHE_FILELOCK = 'cache_filelock'`  # without underscore prefix

The check uses CACHE_FILELOCK ('cache_filelock'), but the assignment is to self._cache_filelock (with underscore prefix). These names don't match, so hasattr(self, CACHE_FILELOCK) always returns `False`.

## Impact
Every call to evict_shard(), evict_coldest_shard(), or prepare_shard() creates a new FileLock or Lock object instead of reusing the previously created one. This could lead to:
* Unnecessary object creation overhead

## Suggested Fix
Change the hasattr check to use the actual attribute name:
`if not hasattr(self, '_cache_filelock'):`
Or update the constant to match:
`CACHE_FILELOCK = '_cache_filelock'`

XiaohanZhangCMU:
## Issue #, if available:

- Fixes #963 Fix.
- Related to #958 

## Merge Checklist:

### General

### Tests

### DIFF
--- a/streaming/base/constant.py
+++ b/streaming/base/constant.py
@@ -27,7 +27,7 @@ SHM_TO_CLEAN = [
 
 # filelock names
 BARRIER_FILELOCK = 'barrier_filelock'
-CACHE_FILELOCK = '_cache_filelock'
+CACHE_FILELOCK = 'cache_filelock'
 
 # Time to wait, in seconds.
 TICK = 0.007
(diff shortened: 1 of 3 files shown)


bowenyang008 (approved):
Thanks!

